PR #27, opened 2022-03-11
Select and generate

thenaomig:
Generate does work in this version, though it needs more checks for bad inputs in select. Warning that a sample operation took 10min with 64 workers.


bkg:
If you could run your code through [black](https://github.com/psf/black) and [flake8](https://flake8.pycqa.org/en/latest/), that will take care of the PEP8 style issues, unused imports, and anything undefined. I will see about automating these checks for the repo, but they are straightforward to run. After installing those packages, you can run them with:
```
black climakitae
flake8 climakitae
```
Be aware that black will make the changes in place, so if you'd rather see what it's doing prior:
```
black --diff --color climakitae
```

thenaomig:
Is the 79 characters per line thing really important? That's the main thing coming up from flake8. (Black mostly just turned all of my ' into ").

bkg:
It improves readability to limit it to something. I think black uses 88 chars where flake8 uses 79, so you can ignore it if you ran black first.

thenaomig:
This may help: https://github.com/cal-adapt/cae-notebooks/blob/timeseriestools/select_and_generate.ipynb
Demos how these should be used.

climakitae/data_loaders.py
```diff
@@ -0,0 +1,125 @@
...
+    if location.subset_by_lat_lon == True:
+        geom = _get_as_shapely(location)
+    else:
+        geom = False  # for now... later a cached polygon will be an elseif option too
+
+    all_files = xr.Dataset()
+    for one_scenario in selections.scenario:
+        files_by_scenario = _get_file_list(cat, selections, one_scenario)
+        temp = _open_and_concat(cat, files_by_scenario, selections, geom)
+        all_files.assign(one_scenario=temp)
```
bkg: Same comment about the use of `assign()` as above.

climakitae/data_loaders.py
```diff
@@ -0,0 +1,125 @@
...
+
+def _open_and_concat(cat, file_list, selections, geom):
+    """
+    Open multiple zarr files, and add them to one big xarray Dataset. Coarsens in time, and/or
+    subsets in space if selections so-indicates. Won't work unless the file_list supplied
+    contains files of only one nominal resolution (_get_file_list ensures this).
+    """
+    all_files = xr.Dataset()
+    for one_file in file_list:
+        with cat[one_file].to_dask() as data:
```
bkg, line 61: I don't believe `with` is desirable here if you want to keep a reference to the `Dataset` for later use and not call `close()` on it.

climakitae/data_loaders.py
```diff
@@ -0,0 +1,125 @@
...
+            source_id = data.attrs["source_id"]
+            if selections.variable == "precipitation (total)":
```
bkg, line 63: Since these if statements just pass and don't do anything, this could be simplified to:
``` python
if selections.variable not in ("precipitation (total)", "wind 10m magnitude"):
    var_name = _get_var_name(cat, selections.variable)
    data = data[var_name]
```

climakitae/data_loaders.py
```diff
@@ -0,0 +1,125 @@
...
+                var_name = _get_var_name(cat, selections.variable)
+                data = data[var_name]
+            # coarsen in time if 'selections' so-indicates:
+            if selections.timescale == "daily":
+                data = data.resample(time="1D").mean("time")
+            elif selections.timescale == "monthly":
+                data = data.resample(time="1MS").mean("time")
+            if geom:
+                # subset data spatially:
+                ds_region = regionmask.Regions(
```
bkg, line 77: This could be handled with `rasterio.features.geometry_mask()` which is already installed on our docker image, so that could avoid adding a new dependency to `regionmask` just for spatial subsetting which might be nice.

bkg (commented):
Very neat, I like the panel layout and widgets for data selection. I added some suggestions to consider, let me know if it's easier to talk through any of it.

climakitae/core.py
```diff
@@ -1,154 +1,39 @@
...
-    '''
-    dataOneModel = pd.read_csv('archive_workshops/workshop#1example/timeSeries_tas_global.csv', header=[0,1], index_col=0)
-    return dataOneModel
+    #=== Generate ===================================                                                 
+    def generate(self):
+        '''
+        Uses the information gathered in 'select' and store in 'selections' and 'location' 
+        to generate an xarray Dataset as specified, and return that Dataset object.
+        '''
+        print('... reading-in data ...')
```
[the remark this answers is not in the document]
thenaomig: The logging module requires it be a warning level thing before it prints to the console though, and I don't want "WARNING:data loading"... I can just remove the print statement for the moment, but this is an outstanding problem that we'll want to solve in more than one case (also with export) to let the user know that something is happening that might be taking a while.

climakitae/selectors.py
```diff
@@ -0,0 +1,222 @@
...
+
+class CatalogContents:
+    def __init__(self):
+        # get the list of data variables from one of the zarr files:
+        cat = intake.open_catalog("s3://cdcat/cae.yaml")
+        _ds = cat[list(cat)[0]].to_dask()
+        _variable_choices_hourly_wrf = [
+            _ds[i].attrs["description"] for i in _ds.data_vars
+        ]
+        _variable_choices_hourly_wrf = _variable_choices_hourly_wrf  # +['precipitation (total)', 'wind 10m magnitude'] #which we'll derive from what's there
```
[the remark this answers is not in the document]
thenaomig: There is a trello card for me to add those, so it's just a reminder to myself. But I will comment-out the entire line for now.

climakitae/data_loaders.py
thenaomig: I'm not sure either. The xarray docs make it clear that this form of opening (with 'with') is preferred, and I don't want to keep a reference to the entire thing, since I'm going to be subsetting and concatenating with other data. The reason to test it more is that it might effect speed, correct?

climakitae/data_loaders.py
thenaomig: Yes, I will be adding things there, but the most common case should still come first.

climakitae/data_loaders.py
thenaomig: so, I see your point about not adding another dependency. I was messing around with rasterio, but I couldn't get it to work, and this does work. Perhaps instead I make a card to go back and try to do that later?

climakitae/selectors.py
```diff
@@ -0,0 +1,222 @@
...
+            "Statistical": _variable_choices_daily_loca,
+        }
+
+        self._variable_choices = {
+            "hourly": _variable_choices_hourly,
+            "daily": _variable_choices_daily,
+        }
+
+        # hard-coded options:
+        self._scenario_choices = [
```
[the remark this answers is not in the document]
thenaomig: brilliant :)

climakitae/selectors.py
```diff
@@ -0,0 +1,222 @@
...
+        """Updates the 'location' object to be the point associated with the selected station."""
+        location = _stations_database[self.cached_station]
+
+
+class CatalogContents:
+    def __init__(self):
+        # get the list of data variables from one of the zarr files:
+        cat = intake.open_catalog("s3://cdcat/cae.yaml")
+        _ds = cat[list(cat)[0]].to_dask()
+        _variable_choices_hourly_wrf = [
```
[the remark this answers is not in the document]
thenaomig: this is way better 👍

climakitae/selectors.py
```diff
@@ -0,0 +1,222 @@
...
+    @param.depends("cached_station", watch=True)
+    def _update_location(self):
+        """Updates the 'location' object to be the point associated with the selected station."""
+        location = _stations_database[self.cached_station]
+
+
+class CatalogContents:
+    def __init__(self):
+        # get the list of data variables from one of the zarr files:
+        cat = intake.open_catalog("s3://cdcat/cae.yaml")
```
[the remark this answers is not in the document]
thenaomig: I get that we should try to load the catalog only once, but I'm not sure what you mean as a suggestion here "...so this could be set to an instance attribute here and referenced elsewhere which should cache the catalog and make additional calls faster."

climakitae/core.py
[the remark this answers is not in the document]
bkg: It's possible to log INFO statements, not just WARNING as there are a few different severity levels. It should also be possible to control the formatting, so we can revisit this.

climakitae/data_loaders.py
bkg: Yes, it's important when using local files to close them properly. This is different though since these are in S3 and a network request is fired every time it's re-opened which isn't optimal.

climakitae/data_loaders.py
bkg: Good idea, feel free to assign me to the card and I can work on that.

climakitae/selectors.py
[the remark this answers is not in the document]
thenaomig: Are you talking about a nested class called Catalog inside Application?

climakitae/data_loaders.py
```diff
@@ -60,13 +29,13 @@ def _open_and_concat(cat, file_list, selections, geom):
     for one_file in file_list:
         with cat[one_file].to_dask() as data:
             source_id = data.attrs["source_id"]
-            if selections.variable == "precipitation (total)":
+            if selections.variable not in ("precipitation (total)", "wind 10m magnitude"):
+                data = data[selections.variable]
+            elif selections.variable == "precipitation (total)":
                 pass
             elif selections.variable == "wind 10m magnitude":
```
bkg, line 36: Sorry, I probably wasn't clear, but these elifs aren't needed with the change to `if selections.variable not in ("precipitation (total)", "wind 10m magnitude"):`

thenaomig: I know they're not needed, but they will be needed later when there's something there other than 'pass', which is one of the cards on my to-do.

climakitae/data_loaders.py
thenaomig: Wow. I think that actually really sped things up!

PR #28, opened 2022-03-13
Threshold

nabigac:
changes/additions:

- added threshold_tools python file with list of functions
- added visualize python file with list of functions
- linked visualize commands in threshold_tools
- added more packages to setup config file
- ran black and flake8 for formatting

### DIFF
--- a/climakitae/threshold_tools.py
+++ b/climakitae/threshold_tools.py
@@ -1,0 +1,540 @@
+import numpy as np
+import pandas as pd
+import xarray as xr
+from scipy import stats
+
+import lmoments3 as lm
+from lmoments3 import distr as ldistr
+from lmoments3 import stats as lstats
+
+import matplotlib.pyplot as plt
+from matplotlib import cm
+import cartopy.crs as ccrs
+import cartopy.feature as cfeature
+import geoviews as gv
+
+import holoviews as hv
+from holoviews import opts
+import hvplot.pandas
+import hvplot.xarray
+
+from .visualize import get_frequency_plot, get_geospatial_plot
+
+#####################################################################
+
+def get_ams(da, extremes_type='max'):
+    
+    extremes_types = ['max']
+    if extremes_type not in extremes_types:
+        raise ValueError('invalid extremes type. expected one of the following: %s' % extremes_types)
+    
+    if extremes_type == 'max':
+        ams = da.resample(time='A').max(keep_attrs=True)
+        ams.attrs['extreme value extraction method'] = 'block maxima'
+        ams.attrs['extremes type'] = 'maxima'
+        ams.attrs['block size'] = '1 year'
+        ams.attrs['timeseries type'] = 'annual max series'
+    
+    # if extremes_type == 'min':
+    #     ams = da.resample(time='A').min(keep_attrs=True)
+    #     ams.attrs['extreme value extraction method'] = 'block maxima'
+    #     ams.attrs['extremes type'] = 'minima'
+    #     ams.attrs['block size'] = '1 year'
+    #     ams.attrs['timeseries type'] = 'annual min series'
+    
+    return ams
+
+#####################################################################
+
+def get_lmoments(ams, distr='gev', multiple_points=True):
+    
+    distrs = ['gev', 'gumbel', 'weibull', 'pearson3', 'genpareto']
+    if distr not in distrs:
+        raise ValueError("invalid distr type. expected one of the following: %s" % distrs)
+        
+    ams_attributes = ams.attrs
+    
+    if multiple_points == True:
+        ams = ams.stack(allpoints=['x','y']).squeeze().groupby('allpoints')
+    
+    def apply_function(ams):
+    
+        def function(ams):
+            
+            if distr == 'gev':
+                lmoments = ldistr.gev.lmom_fit(ams)
+            
+            if distr == 'gumbel':
+                lmoments = ldistr.gum.lmom_fit(ams)
+                
+            if distr == 'weibull':
+                lmoments = ldistr.wei.lmom_fit(ams)
+                
+            if distr == 'pearson3':
+                lmoments = ldistr.pe3.lmom_fit(ams)
+                
+            if distr == 'genpareto':
+                lmoments = ldistr.gpa.lmom_fit(ams)
+
+            return lmoments
+    
+        return xr.apply_ufunc(function, ams, input_core_dims=[["time"]], 
+                          exclude_dims=set(("time",)), output_core_dims=[[]])
+
+    lmoments = apply_function(ams)
+    
+    new_ds = lmoments.to_dataset()
+    new_ds = new_ds.rename({'T2':'lmoments'})
+    
+    if multiple_points == True:
+        new_ds = new_ds.unstack('allpoints')
+        
+    new_ds.attrs = ams_attributes
+    new_ds.attrs['distribution'] = '{}'.format(str(distr))
+    
+    return new_ds
+
+#####################################################################
+
+def get_aicc_stat(ams, multiple_points=True):
+    
+    ams_attributes = ams.attrs
+    
+    if multiple_points == True:
+        ams = ams.stack(allpoints=['x','y']).squeeze().groupby('allpoints')
+        
+    def apply_function(ams):
+    
+        def function(ams):
+            
+            try:
+                lmoments_gev = ldistr.gev.lmom_fit(ams)
+                aicc_gev = ['gev',lstats.AICc(ams, 'gev', lmoments_gev)]
+            except ValueError:
+                aicc_gev = ['gev', np.nan]
+            
+            try:
+                lmoments_gum = ldistr.gum.lmom_fit(ams)
+                aicc_gum = ['gumbel', lstats.AICc(ams, 'gum', lmoments_gum)]
+            except ValueError:
+                aicc_gum = ['gumbel', np.nan]
+            
+            try:
+                lmoments_wei = ldistr.wei.lmom_fit(ams)
+                aicc_wei = ['weibull',lstats.AICc(ams, 'wei', lmoments_wei)]
+            except ValueError:
+                aicc_wei = ['weibull', np.nan]
+
+            try:
+                lmoments_pe3 = ldistr.pe3.lmom_fit(ams)
+                aicc_pe3 = ['pearson3', lstats.AICc(ams, 'pe3', lmoments_pe3)]
+            except ValueError:
+                aicc_pe3 = ['pearson3', np.nan]
+            
+            try:
+                lmoments_gpa = ldistr.gpa.lmom_fit(ams)
+                aicc_gpa = ['genpareto', lstats.AICc(ams, 'gpa', lmoments_gpa)]
+            except ValueError:
+                aicc_gpa = ['genpareto', np.nan]
+            
+            all_aicc_results = str(aicc_gev) + str(aicc_gum) + str(aicc_wei) + str(aicc_pe3) + str(aicc_gpa)
+            all_aicc_results_string = str(all_aicc_results)
+            
+            lowest_aicc_value = min(aicc_gev[1], aicc_gum[1], aicc_wei[1], aicc_pe3[1], aicc_gpa[1])
+            
+            if lowest_aicc_value == aicc_gev[1]:
+                lowest_aicc_distr = aicc_gev[0]
+            elif lowest_aicc_value == aicc_gum[1]:
+                lowest_aicc_distr = aicc_gum[0]
+            elif lowest_aicc_value == aicc_wei[1]:
+                lowest_aicc_distr = aicc_wei[0]
+            elif lowest_aicc_value == aicc_pe3[1]:
+                lowest_aicc_distr = aicc_pe3[0]
+            elif lowest_aicc_value == aicc_gpa[1]:
+                lowest_aicc_distr = aicc_gpa[0]
+
+            return all_aicc_results, lowest_aicc_distr, lowest_aicc_value
+        
+        
+        return xr.apply_ufunc(function, ams, input_core_dims=[["time"]], 
+                              exclude_dims=set(("time",)), output_core_dims=[[],[],[]])
+    
+    all_aicc_results, lowest_aicc_distr, lowest_aicc_value = apply_function(ams)
+    
+    all_aicc_results = all_aicc_results.rename('all_aicc_results')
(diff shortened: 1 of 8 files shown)


climakitae/threshold_tools.py
```diff
@@ -0,0 +1,613 @@
...
+
+    lmoments = apply_function(ams)
+
+    new_ds = lmoments.to_dataset()
+    new_ds = new_ds.rename({"T2": "lmoments"})
+
+    if multiple_points == True:
+        new_ds = new_ds.unstack("allpoints")
+
+    new_ds.attrs = ams_attributes
```
thenaomig, line 103: Recommend returning this as a dataarray instead of a dataset, so that it'll be easier to parallelize across grid cells. All attributes can be attributes of the variable and that will work.

thenaomig (commented):
More soon...

climakitae/threshold_tools.py
thenaomig: That did not highlight the section of code I meant... but I'm talking about the output of the get_lmoments function.

climakitae/threshold_tools.py
```diff
@@ -0,0 +1,613 @@
...
+
+            return all_aicc_results, lowest_aicc_distr, lowest_aicc_value
+
+        return xr.apply_ufunc(
+            function,
+            ams,
+            input_core_dims=[["time"]],
+            exclude_dims=set(("time",)),
+            output_core_dims=[[], [], []],
+        )
```
thenaomig, line 184: You may want to try adding the argument dask='allowed'... which I expect should work, or else the 'parallelize' option. Doing the step of loading 'ams' in the notebook is not optimal -- takes memory and time unnecessarily.
https://xarray.pydata.org/en/stable/generated/xarray.apply_ufunc.html

thenaomig (commented):
I'd have another wrapper function so that the user never needs to see 'ams' in the notebook. For this particular type of analysis, it's a given for all that follows. Later if they have any other choice then that'll be reflected in the initial selections. So I'd consider structuring it so that there is an extreme_value object (e.g. a class), so the syntax, after selecting your parameter choices, would be output = tt.extreme_value(parameters)
and then if you wanted to see the aicc stat you'd type
aicc = output.aicc() and it'd only then calculate it. same with the ks_stat. 
Then the plotting functions could be built in as well, so you'd type
output.get_geospatial_plot('which_one')

climakitae/threshold_tools.py
```diff
@@ -0,0 +1,613 @@
...
+
+    distrs = ["gev", "gumbel", "weibull", "pearson3", "genpareto"]
+    if distr not in distrs:
+        raise ValueError(
+            "invalid distr type. expected one of the following: %s" % distrs
+        )
+
+    ams_attributes = ams.attrs
+
+    if multiple_points == True:
```
bkg, line 64: Booleans can be checked directly like `if multiple_points:`, no need to test for equality.

climakitae/threshold_tools.py
```diff
@@ -0,0 +1,613 @@
...
+        ams = ams.stack(allpoints=["x", "y"]).squeeze().groupby("allpoints")
```
bkg, line 65: It doesn't look like is needed since `unstack()` just undoes this below. It's always faster to reduce the overall number of dask tasks.

climakitae/threshold_tools.py
```diff
@@ -0,0 +1,613 @@
...
+            except ValueError:
+                aicc_pe3 = ["pearson3", np.nan]
+
+            try:
+                lmoments_gpa = ldistr.gpa.lmom_fit(ams)
+                aicc_gpa = ["genpareto", lstats.AICc(ams, "gpa", lmoments_gpa)]
+            except ValueError:
+                aicc_gpa = ["genpareto", np.nan]
+
+            all_aicc_results = (
```
bkg, line 152: Instead of a string of all the AICc values which is difficult to use afterward, you could add a coordinate to the `DataArray` with the distribution names and corresponding AICc values. That would also allow you to call the `DataArray.min()` method to find the distribution with the lowest AICc.

climakitae/threshold_tools.py
```diff
@@ -0,0 +1,613 @@
...
+
+
+def get_aicc_stat(ams, multiple_points=True):
+
+    ams_attributes = ams.attrs
+
+    if multiple_points == True:
+        ams = ams.stack(allpoints=["x", "y"]).squeeze().groupby("allpoints")
+
+    def apply_function(ams):
```
bkg, line 119: `apply_function` is an extra inner function that isn't needed here and you could use a descriptive name for `function` and move it to the module level instead of nesting it here.

climakitae/threshold_tools.py
```diff
@@ -0,0 +1,613 @@
...
+        def function(ams):
+
+            try:
+                lmoments_gev = ldistr.gev.lmom_fit(ams)
```
bkg, line 123: It would help readability here by looping over the distribution names and calling `AICc` on each one instead of multiple lines with the same call.

bkg (commented):
Thanks for the PR Nabig and all the work you put into this!

As an overall suggestion to the general approach here, it's inefficient to redo the distribution fitting within each function since that is very compute intensive at the scale of our data. Instead, the fitting can be done once to build the fitted distributions for later calls to `cdf()`, `ppf()` and the params can be used with `AICc` and `kstest`. It is tricky getting these to play nicely with `xr.apply_ufunc()`, but it will help cut down on some of the repetition here.

climakitae/threshold_tools.py
```diff
@@ -0,0 +1,613 @@
...
+        )
+
+    ams_attributes = ams.attrs
+
+    if multiple_points == True:
+        ams = ams.stack(allpoints=["x", "y"]).squeeze().groupby("allpoints")
+
+    def apply_function(ams):
+        def function(ams):
+
```
odoherty, line 69: lmoments is preferred for small samples - so makes sense for say 30 years of a single run... if we are going to be applying this to multiple models, multiple ensemble members it's worth considering the option of maximum-likelihood estimators, which have the additional benefit of being able to handle (potentially) non stationary distributions... motion to add this to backlog

climakitae/threshold_tools.py
```diff
@@ -0,0 +1,613 @@
...
+            "invalid extremes type. expected one of the following: %s" % extremes_types
+        )
+
+    if extremes_type == "max":
+        ams = da.resample(time="A").max(keep_attrs=True)
+        ams.attrs["extreme value extraction method"] = "block maxima"
+        ams.attrs["extremes type"] = "maxima"
+        ams.attrs["block size"] = "1 year"
+        ams.attrs["timeseries type"] = "annual max series"
+
```
odoherty, line 40: For training @gdicecco documentation should note limitations of annual block maxima approach, i.e. that a lot of extreme events are excluded, or said differently you could have 5 extremes in one year each of which is more extreme than the peak event in subsequent years (important for california atmospheric river events, wet years vs dry years)

odoherty: propose adding Peaks over Threshold approach to backlog

climakitae/threshold_tools.py
```diff
@@ -0,0 +1,613 @@
...
+                    d_statistic = ks[0]
+                    p_value = ks[1]
+                except ValueError:
+                    d_statistic = np.nan
+                    p_value = np.nan
+
+            if distr == "weibull":
+                try:
+                    lmoments = ldistr.wei.lmom_fit(ams)
+                    fitted_distr = stats.weibull_min(**lmoments)
```
odoherty, line 252: Just a note that the weibull_min is most common, and IMO the best choice

climakitae/threshold_tools.py
```diff
@@ -0,0 +1,613 @@
...
+
+    return new_ds
+
+
+#####################################################################
+
+
+def get_return_prob(ams, threshold, distr="gev", multiple_points=True):
+
+    distrs = ["gev", "gumbel", "weibull", "pearson3", "genpareto"]
```
odoherty, line 422: I'm wondering if we should add a note in the notebook describing when each of these distributions is most commonly used?  Like a standard best practice.

climakitae/visualize.py
```diff
@@ -0,0 +1,136 @@
...
+import geoviews as gv
+
+import holoviews as hv
+from holoviews import opts
+import hvplot.pandas
+import hvplot.xarray
+
+#####################################################################
+
+
```
odoherty, line 18: flagging here that in the initial notebook some datapoints failed to visualize

thenaomig:
FYI, I just 'rebased' to main so that if we were testing in this branch it'd have the latest changes to main included, and that required some manual merging, and is why you see commits from me in this 'conversation'.

odoherty:
threshold_tools requires package lmoments to calculate parameters for extreme distributions via L-Moments
climakitae includes lmoments 1.0.1, which references scipy functionality which has been depreciated
command below installs lmoments 1.0.4 which does not make references to missing/depreciated values

!pip install git+https://github.com/OpenHydrology/lmoments3.git

this is a conflict that needs resolving

thenaomig:
Had tried to deal with all the conflicts by rebasing to a later version of main, but now I feel like you might just want to remove core, selectors, and data_loader files in a commit. As far as I'm aware the only changes there had to do with having run black or not, and little spaces here and there, and they've changed enough since that going through and resolving conflicts would be pretty annoying, when there were no substantive changes to those files in this PR.

climakitae/threshold_tools.py
```diff
@@ -0,0 +1,613 @@
...
+                    ks = stats.kstest(
+                        ams,
+                        "genextreme",
+                        args=(lmoments["c"], lmoments["loc"], lmoments["scale"]),
+                    )
+                    d_statistic = ks[0]
+                    p_value = ks[1]
+                except ValueError:
+                    d_statistic = np.nan
+                    p_value = np.nan
```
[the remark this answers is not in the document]
nabigac: yeah, i believe if the criteria for how big or small the l-moments are (depending on the distribution), the lmoments3 package returns a ValueError, so the value isn't calculated for the grid. The issue is without that this one error for that one grid can error out the rest of the calculations for other grids. having np.nan allows these particular grids to just show up as white undefined blocks during visualization, whereas it's calculated for everything else.

i was thinking we could include details in the documentation of what it means when users might see these sort of undefined value for a grid cell.

climakitae/threshold_tools.py
nabigac: this is great to know! will update.

climakitae/threshold_tools.py
nabigac: that makes sense! for extreme value analysis, at least, it seems like the literature was pointing towards the GEV as being the default so went with that. But if you feel good about weibull_min being the one that's most commonly used by utilities, i'm happy to set this as the default. let me know!

climakitae/threshold_tools.py
nabigac: yeah! i think this might be helpful to include in the documentation, where maybe we can add more details about when certain distributions are used.

climakitae/visualize.py
nabigac: probably due to a ValueError where if some checks through the lmoments3 package didn't go through, they come back as a nan value for that particular grid cell (rather than causing none of the visualization to run). another good thing to explain in documentation, why a user might be seeing that!

climakitae/threshold_tools.py
```diff
@@ -0,0 +1,613 @@
...
+        raise ValueError(
+            "invalid distr type. expected one of the following: %s" % distrs
+        )
+
+    ams_attributes = ams.attrs
+
+    if multiple_points == True:
+        ams = ams.stack(allpoints=["x", "y"]).squeeze().groupby("allpoints")
+
+    def apply_function(ams):
```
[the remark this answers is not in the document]
nabigac: yeah, i think some of these could be turned into arrays and would streamline things a bit. would maybe add this to the backlog given there seems to be a lot of changes/adjustments that need to happen before the launch this week.

climakitae/threshold_tools.py
```diff
@@ -0,0 +1,613 @@
...
+                lmoments = ldistr.wei.lmom_fit(ams)
+
+            if distr == "pearson3":
+                lmoments = ldistr.pe3.lmom_fit(ams)
+
+            if distr == "genpareto":
+                lmoments = ldistr.gpa.lmom_fit(ams)
+
+            return lmoments
+
```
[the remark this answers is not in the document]
nabigac: time dimension is removed because multiple years (ex. 30 years of extreme value points) are inputted to produce an output that is just a single return value or probability. if time isn't removed, there's an error that occurs.

climakitae/threshold_tools.py
```diff
@@ -0,0 +1,613 @@
...
+        return xr.apply_ufunc(
+            function,
+            ams,
+            input_core_dims=[["time"]],
+            exclude_dims=set(("time",)),
+            output_core_dims=[[]],
+        )
+
+    lmoments = apply_function(ams)
```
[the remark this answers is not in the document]
nabigac: great suggestion! will update

climakitae/threshold_tools.py
nabigac: will update lmoments to return dataarray instead of dataset

climakitae/threshold_tools.py
nabigac: just to clarify, are you saying just to move the function command out as its own unique function and then call it? let me now if you were thinking of something else!

climakitae/threshold_tools.py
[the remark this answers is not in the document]
nabigac: going to apply this same strategy of removing the apply_function to the rest of the functions as well.

climakitae/threshold_tools.py
nabigac: removed apply_function as an extra inner function and moved the rest of the descriptive names of functions out to the module level instead

climakitae/threshold_tools.py
thenaomig: yes, I'm not seeing what @bkg  is getting at -- you need to stack and unstack to do this operation over all points if the data in question is an xarray object with multiple grid-cells (maybe instead of "multiple points" it'd be clearer to say if "spatial grid"?)

climakitae/threshold_tools.py
thenaomig: Made a card for this in trello.

PR #31, opened 2022-03-24
Progress bar

bethem:


climakitae/explore.py
```diff
@@ -32,3 +32,31 @@ def explore(my_data, do_smoothing=False):
...
     linePlot = hv.Overlay(temp_list)
     return linePlot
+
+
+from dask import delayed, compute
+from dask.diagnostics import ProgressBar
+pbar = ProgressBar()
+pbar.register()
+def progress_bar(func):
+    @wraps(func)
```
thenaomig, line 42: I'm getting 

> NameError: name 'wraps' is not defined

Does this require importing something else?

climakitae/data_loaders.py
```diff
@@ -3,6 +3,8 @@
 import regionmask
 import intake
 import numpy as np
+import explore # for progress bar
```
thenaomig, line 6: Instead of `import explore`, as you'd do for local work, you want this to be `from .explore import progress_bar`

climakitae/explore.py
thenaomig: add `from functools import wraps`
to explore.py

climakitae/explore.py
```diff
@@ -32,3 +32,31 @@ def explore(my_data, do_smoothing=False):
             temp_list.append(temp)
     linePlot = hv.Overlay(temp_list)
     return linePlot
+
+
+from dask import delayed, compute
+from dask.diagnostics import ProgressBar
```
thenaomig, line 38: Move the imports to the very top of the file.

climakitae/data_loaders.py
```diff
@@ -74,6 +76,8 @@ def _get_as_shapely(location):
     )
 
 
+@explore.progress_bar
```
thenaomig, line 79: And then this will just be @progress_bar (because it's imported differently)

thenaomig:
This is nice. Just make the changes I suggested, check them in locally, and push to this branch. Then I'll merge it. It seems to work!

thenaomig (approved):

PR #32, opened 2022-03-24
Metadata

bethem:


climakitae/metadata_update.py
```diff
@@ -0,0 +1,57 @@
...
+        func_kwargs = kwargs
+        
+        # get existing ds/da attributes as a dictionary
+        # which can be updated with new information.
+        orig_attrs = ds.attrs
+
+        # flag that climakitae transform has been applied
+        transform_attrs = {'post_processed' : 'true',
+                     'post_processed_by' : 'Cal-Adapt Analytics'+
+                     'Engine v 0.1'}
```
thenaomig, line 37: I think it's version 0.0.1 at the moment? Maybe there is a way to pull that information dynamically so this doesn't have to be manually updated whenever we declare a new version number.

odoherty: It is climakitae-0.0.1

print(ck.__version__)

bethem:
I fixed that! Why did it revert?? I even have a commit message about it.

climakitae/metadata_update.py
bethem: @odoherty is that an actual command, or a suggestion? Because it doesn't work for me

thenaomig: it is a suggestion -- you will, however, be able to get this somehow from github -- the numbering of releases is something that will manually be done there. You can see that there aren't currently any releases though -- we're intending whatever we have for the soft launch to be 0.0.1

thenaomig: it may not work yet because we've not done any official releases. That's something someone will manually specify on github as 0.0.1 for the soft launch -- enshrining that version of things forever as 0.0.1 -- and then we make a new release when we have an appropriate amount of updates

thenaomig:
This is cool, but sadly will not work on timeseriestools because of the weird way I did it (there is no dataarray or arguments passed -- it's all contained within a class and all the arguments are in an instance of a param-inheriting object. Someone should test this with threshold tools though.

climakitae/metadata_update.py
```diff
@@ -0,0 +1,57 @@
...
+        
+        # get existing ds/da attributes as a dictionary
+        # which can be updated with new information.
+        orig_attrs = ds.attrs
+
+        # flag that climakitae transform has been applied
+        transform_attrs = {'post_processed' : 'true',
+                     'post_processed_by' : 'Cal-Adapt Analytics'+
+                     'Engine v 0.0.1'}
+
```
thenaomig, line 38: I dare say that export should check for this, and if it's not present add a version of it, like saying the data was "retrieved and aggregated by" to capture the minimum of what 'retrieve' (nee 'generate') does.

PR #33, opened 2022-03-25
Timeseriestools

thenaomig:
Doing a minor PR here, just so I can have this for my ongoing testing.
- Changes default lat/lon box location (CA) and display (just an outline)
- adds a checkbox to 'append historical' so that the historical climate simulation and each of it's corresponding SSPs are appended as one variable with a longer time dimension
- adds a checkbox to 'area average', so that the area selected is averaged spatially to return a timeseries

### DIFF
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -56,8 +56,14 @@ def _open_and_concat(file_list, selections, ds_region):
                 .dropna("x", how="all")
                 .dropna("y", how="all")
             )
+
+        if selections.area_average:
+            weights = np.cos(np.deg2rad(data.lat))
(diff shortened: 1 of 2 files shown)


odoherty (approved):
nit: default time to monthly to be consistent with defaulting to lowest resolution

potential need for a metadata attribute to specify multiple points TRUE or FALSE, either add functionality here or remove this requirement from threshold_tools

thenaomig:
> That multiple points thing is not relevant yet in that we don't have station data, at which point you might select multiple stations. If the data is gridded with multiple gridcells, there is no need to specify that -- you can easily examine the dataset and see.

PR #34, opened 2022-03-27
Timeseriestools

thenaomig:
First pass at timeseries tools -- selection panel in place. Usage demo-ed in cae-notebooks/timeseries_example.ipynb

### DIFF
--- a/climakitae/timeseriestools.py
+++ b/climakitae/timeseriestools.py
@@ -37,7 +37,7 @@ class TimeSeriesParams(param.Parameterized):
         default="None", objects=["None", "min", "max", "percentile"]
     )
     resample_window = param.Integer(default=1, bounds=(1, 30))
-    resample_period = param.ObjectSelector(default="A", objects=_time_scales)
+    resample_period = param.ObjectSelector(default="AS-SEP", objects=_time_scales)
     percentile = param.Number(
         default=0,
         bounds=(0, 1),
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -27,6 +27,8 @@ install_requires =
     numpy
     lmoments3
```
odoherty, line 28: Just a note - I had an error pop up when I used the lmoments3 that loads with climakitae when I called threshold_tools.py functions.  I directly pip installed from a git and it worked.  May be a source of error down the road.

thenaomig: Good to know... this is not a part of this PR though -- but I'm glad you tried testing those things in concert.

climakitae/timeseriestools.py
```diff
@@ -0,0 +1,222 @@
...
+    num_timesteps = param.Integer(default=0, bounds=(0, 240))
+    separate_seasons = param.Boolean(
+        default=False, label="Disaggregate into four seasons"
+    )
+
+    extremes = param.ObjectSelector(
+        default="None", objects=["None", "min", "max", "percentile"]
+    )
+    resample_window = param.Integer(default=1, bounds=(1, 30))
+    resample_period = param.ObjectSelector(default="A", objects=_time_scales)
```
bethem, line 40: this is throwing an error:
`from climakitae import timeseriestools as tst`
...
`ValueError: A not in parameter's list of possible objects, valid options include [H, D, MS, AS-SEP]`

thenaomig: Whoops -- was in too much of a hurry with that last change. Fixed.

PR #38, opened 2022-03-29
Export to user

bethem:
Requires some debugging: Can't seem to get the metadata file to write for GeoTIFFs, even though it works for CSVs.


thenaomig:
Two big-picture comments -- 
1. @bkg  has told me that it's not ideal to use print() statements in library code. I am not, however, sure what you should use instead to display your helpful little messages below the cell, or how big a deal this is.
2. If I supply the filename 'my_testfile.nc' then it writes 'my_testfile.nc.nc' -- you could test to see if the user has already supplied a .something at the end of the file: file_var.split('.')[-1]

climakitae/data_export.py
```diff
@@ -0,0 +1,225 @@
...
+        f.write('======== Metadata for '+req_format+' file '+output_name+' ========')
+        f.write('\n')
+        f.write('\n')
+        f.write('\n')
+        f.write('===== Global file attributes =====')
+        f.write('\n')
+        f.write('\n')
+        for att_keys,att_values in list(zip(ds.attrs.keys(),ds.attrs.values())):    
+            f.write(str(att_keys)+" : "+str(att_values))
+            f.write('\n')
```
thenaomig, line 180: This part is nice. I see it in the metadata.txt that accompanies a .csv export, but I don't see something equivalent in the netCDF metadata if no transforms are performed (e.g. if you just retrieve and then export something).

climakitae/data_export.py
```diff
@@ -0,0 +1,225 @@
...
+
+def _export_to_user(user_export_format,data_to_export,
+                    file_name,**kwargs):
+    """
+    The data export method, called by core.Application.export_dataset. Saves
+    a dataset to the current working directory in the output format requested by the user 
+    (which is stored in 'user_export_format').
+    
+    user_export_format: pulled from dropdown called by app.export_as()
+    data_to_export: xarray ds or da to export
```
odoherty, line 19: Probably worth updating the developers guide to note that final transform formats should be in xarray ds or da

climakitae/data_export.py
```diff
@@ -0,0 +1,225 @@
...
+    !pip install pytest-shutil
+    """
+    
+    excel_row_limit = 1048576
+    path_and_name = './'+file_name
+    req_format = user_export_format.output_file_format
+    ds_attrs = data_to_export.attrs
+    ct = datetime.datetime.now()
+    ct_str = ct.strftime("%d-%b-%Y (%H:%M:%S)")    
+    ck_attrs = {'data_retrieved_by' : 'Cal-Adapt Analytics Engine v 0.0.1',
```
odoherty, line 34: add note to backlog to update this line either automatically, or every time we push across a new version manually....

climakitae/data_export.py
bethem: > Two big-picture comments --
> 
> 1. @bkg  has told me that it's not ideal to use print() statements in library code. I am not, however, sure what you should use instead to display your helpful little messages below the cell, or how big a deal this is.

Would love to hear the alternative. Google is giving me nothing. 

For reference, I tried the warnings package, but I disliked that it didn't just show my warning - it showed where in the function call I wrote that warning. Also, I'm not always outputting a warning - sometimes it's a success statement. 

> 2. If I supply the filename 'my_testfile.nc' then it writes 'my_testfile.nc.nc' -- you could test to see if the user has already supplied a .something at the end of the file: file_var.split('.')[-1]

Good point! Will fix.

climakitae/data_export.py
```diff
@@ -0,0 +1,225 @@
...
+   
+    data_to_export.attrs = ds_attrs
+    data_to_export = data_to_export.squeeze(drop=True)
+    file_size_threshold = 5 # in GB
+    bytes_per_gigabyte = 1024*1024*1024    
+    disk_space = shutil.disk_usage('./')[2]/bytes_per_gigabyte
+    data_size = data_to_export.nbytes/bytes_per_gigabyte
+    
+    if (disk_space < data_size):
+        raise ValueError("Not enough disk space to export data!"+
```
odoherty, line 110: Is this something we want user feedback on?  Will 10GB be enough memory?

climakitae/selectors.py
```diff
@@ -361,3 +361,49 @@ def _display_select(selections, location, location_type="area average"):
         )
     )
     return pn.Column(first_row, location_chooser)
+
```
odoherty, line 364: nit: would prefer a less generic name like file_out_selector.py to avoid potential redundancies in the future

climakitae/data_export.py
thenaomig: Based on my googling, I *think* it's:
```
import logging
import sys

logging.basicConfig(stream=sys.stdout, level=logging.INFO)
logger = logging.getLogger('LOGGER_NAME')
logger.info('shown on the console')

```
I don't know for sure if that's right, and I don't really get why it's such a big deal anyway when we're developing for the Jupyter notebook platform, and seems like something that could be changed later, so I wouldn't have it block merging.

climakitae/selectors.py
bethem: oh so a separate file for export-related selectors?

climakitae/data_export.py
bethem: oh okay! My googling failed me. Will give this a try, thank you!

climakitae/data_export.py
```diff
@@ -0,0 +1,225 @@
...
+def _export_to_user(user_export_format,data_to_export,
+                    file_name,**kwargs):
+    """
+    The data export method, called by core.Application.export_dataset. Saves
+    a dataset to the current working directory in the output format requested by the user 
+    (which is stored in 'user_export_format').
+    
+    user_export_format: pulled from dropdown called by app.export_as()
+    data_to_export: xarray ds or da to export
+    variable_name: string corresponding to variable to export
```
bethem, line 20: erase this

climakitae/data_export.py
bethem: it's disk space, not memory. And I don't know if we can change that - it depends on resources.

bethem:
I wanted to do that, but I couldn't figure out how. Is there anything like
that in your select code somewhere I can steal?

climakitae/data_export.py
bethem: oh I see - it's saving as a variable attribute rather than a global one. Huh.

climakitae/core.py
```diff
@@ -1,5 +1,8 @@
-from .selectors import DataSelector, _display_select, LocSelectorArea
+from .selectors import (DataSelector, _display_select, LocSelectorArea,
+                        UserFileChoices, _user_export_select, FileTypeSelector)
 from .data_loaders import _read_from_catalog
+from .data_export import _export_to_user
+from .metadata_update import transform_details
```
thenaomig, line 5: is transform_details used in this file? not sure it needs to be imported in core.py

climakitae/selectors.py
thenaomig: I think it's good to put all of these kinds of UI things in one file together...

PR #39, opened 2022-03-30
Generate fixes

thenaomig:
This PR addresses:
- fixes a bug when scenarios with different time dimension ranges are both selected
- fixes an issue in which the original metadata was lost in 'generate'
- curates the WRF variable list
- adds a "Precipitation (total)" option that is the sum of RAINC + RAINNC
- renames 'generate' as 'retrieve' (in core.py)

### DIFF
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -6,7 +6,6 @@ import numpy as np
 
 # support methods for core.Application.generate
 
-
 def _get_file_list(selections, scenario):
     """
     Returns a list of simulation names for all of the simulations present in the catalog
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -5,6 +5,7 @@ import intake
 import numpy as np
 
 # support methods for core.Application.generate
+xr.set_options(keep_attrs=True)
 
 def _get_file_list(selections, scenario):
     """
@@ -31,6 +32,7 @@ def _open_and_concat(file_list, selections, ds_region):
     all_files = xr.Dataset()
     for one_file in file_list:
         data = cat[one_file].to_dask()
+        attributes = data.attrs
         source_id = data.attrs["source_id"]
         if selections.variable not in ("precipitation (total)", "wind 10m magnitude"):
             data = data[selections.variable]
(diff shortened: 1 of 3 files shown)


climakitae/data_loaders.py
```diff
@@ -118,26 +133,35 @@ def _read_from_catalog(selections, location):
...
+        one_scenario = "Historical Climate"
         files_by_scenario = _get_file_list(selections, one_scenario)
         historical = _open_and_concat(files_by_scenario, selections, ds_region)
-        
-    all_files = xr.Dataset()
+
+    all_files_list = []
     for one_scenario in selections.scenario:
         if selections.append_historical:
             if "SSP" in one_scenario:
```
bethem, line 143: when I only have 'Append Historical' checked, I get
`TypeError: argument of type 'NoneType' is not iterable`

This makes sense (can't append to nothing!), but maybe amend Append Historical to "Append Historical (select SSP and/or reconstruction)". Or is there a way to only have it appear if another thing is checked in the first place? There could also be an error that explicitly states that another box needs to be checked.

climakitae/data_loaders.py
```diff
@@ -32,19 +33,22 @@ def _open_and_concat(file_list, selections, ds_region):
     all_files = xr.Dataset()
     for one_file in file_list:
         data = cat[one_file].to_dask()
+        attributes = data.attrs
         source_id = data.attrs["source_id"]
```
bethem, line 37: Many combinations of variable, scenario, and temporal resolution result in 
`KeyError: 'source_id'` 

It would take a very, very long time to make an exhaustive list. Looks like it comes down to some missing metadata?

climakitae/data_loaders.py
thenaomig: Adding another assert in data_loaders._read_from_catalog

climakitae/data_loaders.py
thenaomig: Thanks for catching that. It seems to be the second time through and every time thereafter. I need to make that line a *copy* of the contents of data.attrs (it was treating it like a reference), so if a previous iteration removes an attribute then that removal won't persist.

climakitae/data_loaders.py
thenaomig: So at least it gives an error that will make sense to the user now: "Please select at least one..."

climakitae/data_loaders.py
```diff
@@ -32,19 +34,22 @@ def _open_and_concat(file_list, selections, ds_region):
...
         data = cat[one_file].to_dask()
+        attributes = deepcopy(data.attrs)
         source_id = data.attrs["source_id"]
-        if selections.variable not in ("precipitation (total)", "wind 10m magnitude"):
+        if selections.variable not in ("Precipitation (total)", "wind 10m magnitude"):
             data = data[selections.variable]
-        elif selections.variable == "precipitation (total)":
-            pass
+        elif selections.variable == "Precipitation (total)":
+            data = data["RAINC"] + data["RAINNC"]
```
odoherty, line 42: What units attribute gets saved with this new variable?

thenaomig: mm, same as for the two that are added together to get it -- just verified

PR #41, opened 2022-04-28
Select fixes

thenaomig:
- Fixes:
    -  a bug caused by a mis-typed variable in an assert in data_loaders.py
    - scenario options no longer reset to eliminate 'Historical Climate' when you select 'Append historical'
- Adds the following to the 'select' panel:
    - Visualization of selected scenarios on a timeline 
    - Time-slicing functionality
        - If the user selects a time-slice for which there is zero data, an assert error points this out when they call 'retrieve' 
    - Shows bounds of WRF domains on area selection map
    - Makes moving between different kinds of area-selection more intuitive:
        - If you change the lat/lon sliders, changes area_subset to lat/lon
        - If you change the cached_area selection, changes area_subset to appropriate type

### DIFF
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -168,4 +168,5 @@ def _read_from_catalog(selections, location):
     all_files = all_files.to_array("scenario")
     all_files.name = selections.variable
     all_files.attrs = attributes
+    assert all_files.time.size != 0, "Dataset will be empty. Please adjust selections."
     return all_files
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -1,4 +1,6 @@
 import xarray as xr
+import cartopy.crs as ccrs
+import dask
 from shapely.geometry import box
 import regionmask
 import intake
@@ -58,18 +60,23 @@ def _open_and_concat(file_list, selections, ds_region):
             )
         )
         # subset data spatially:
+        data_crs = ccrs.LambertConformal(central_longitude=-70,central_latitude=38,
+                                 standard_parallels=(30.0,60.0))
+
         if ds_region:
+            output = data_crs.transform_points(ccrs.PlateCarree(),
+                                                   x=ds_region.coords[0][:,0],
+                                                   y=ds_region.coords[0][:,1])
+
+            data = data.sel(x=slice(np.nanmin(output[:,0]), np.nanmax(output[:,0])),
+                y=slice(np.nanmin(output[:,1]), np.nanmax(output[:,1])))
+
             mask = ds_region.mask(data.lon, data.lat, wrap_lon=False)
             assert (
                 False in mask.isnull()
             ), "Insufficient gridcells are contained within the bounds."
-
-            data = (
-                data.where(np.isnan(mask) == False)
-                .dropna("x", how="all")
-                .dropna("y", how="all")
-            )
-
+            data = data.where(np.isnan(mask) == False)
+            
         if selections.area_average:
             weights = np.cos(np.deg2rad(data.lat))
             data = data.weighted(weights).mean("x").mean("y")
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -50,8 +50,10 @@ def _open_and_concat(file_list, selections, ds_region):
         elif selections.timescale == "monthly":
             data = data.resample(time="1MS").mean("time")
             attributes["frequency"] = "1month"
+        # time-slice:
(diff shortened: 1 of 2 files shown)


climakitae/data_loaders.py
```diff
@@ -58,18 +60,23 @@ def _open_and_concat(file_list, selections, ds_region):
...
         # subset data spatially:
+        data_crs = ccrs.LambertConformal(central_longitude=-70,central_latitude=38,
+                                 standard_parallels=(30.0,60.0))
+
         if ds_region:
+            output = data_crs.transform_points(ccrs.PlateCarree(),
+                                                   x=ds_region.coords[0][:,0],
+                                                   y=ds_region.coords[0][:,1])
+
+            data = data.sel(x=slice(np.nanmin(output[:,0]), np.nanmax(output[:,0])),
```
nabigac, line 71: how does nanmin and nanmax work here? does it only select coordinates without nans?

nabigac:
did some tests and pulled some data using different time ranges, options, locations, etc., and everything ran smoothly!

nabigac (approved):

climakitae/data_loaders.py
thenaomig: It ignores the nans, whereas the default min, max, etc does not. Numpy as a nan-ignoring version of all the standard basic stats.

climakitae/selectors.py
```diff
@@ -334,38 +424,114 @@ class DataSelector(param.Parameterized):
...
+        ax.yaxis.set_major_locator(ticker.NullLocator())
+        ax.spines["top"].set_color("none")
+        ax.xaxis.set_ticks_position("bottom")
+        ax.set_xlim(1950, 2100)
+        ax.set_ylim(0, 1)
+        ax.xaxis.set_major_locator(ticker.AutoLocator())
+        ax.xaxis.set_minor_locator(ticker.AutoMinorLocator())
+        mpl_pane = pn.pane.Matplotlib(fig0, dpi=144)
+
+        def update_bars(scenario, y_offset):
```
bkg, line 498: Same comment about closures applies here, this can be inlined or if it's intended to be used elsewhere, as a method or module level function.

climakitae/selectors.py
```diff
@@ -334,38 +424,114 @@ class DataSelector(param.Parameterized):
...
-            None
-        ]  # resetting this avoids indexing errors with prior values
+        if self.append_historical and self.scenario is not None:
+            if "Historical Climate" in self.scenario:
+                _scenarios = self.scenario
+                _scenarios.remove("Historical Climate")
+                self.scenario = _scenarios
+
+    @param.depends("scenario", "append_historical", watch=True)
+    def _update_time_slice_range(self):
```
bkg, line 453: As a general comment, it would be good to think about how to read the available year ranges from the catalog instead of hard coding them here, which will be problematic if LOCA uses a different historical period.

climakitae/data_loaders.py
```diff
@@ -50,19 +52,31 @@ def _open_and_concat(file_list, selections, ds_region):
...
             attributes["frequency"] = "1month"
+        # time-slice:
+        data = data.sel(
+            time=slice(
+                str(selections.time_slice[0]) + "0101",
+                str(selections.time_slice[1]) + "1231",
+            )
+        )
+        # subset data spatially:
+        data_crs = ccrs.LambertConformal(central_longitude=-70,central_latitude=38,
```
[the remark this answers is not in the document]
thenaomig: Nice, thanks.

climakitae/selectors.py
```diff
@@ -162,13 +213,52 @@ def view(self):
...
+        _plot_wrf_domains(ax)
         ax.coastlines()
+        ax.add_feature(cfeature.BORDERS)
         ax.add_feature(cfeature.STATES, linewidth=0.5)
-
+        ax.annotate(
+            "45-km grid",
+            xy=(-154, 33.8),
+            rotation=28,
+            xycoords=ccrs.PlateCarree()._as_mpl_transform(ax),
```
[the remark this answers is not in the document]
thenaomig: Noted about the multiple instantiation... but this needs to be done on the xy-coords, since those are the coordinate dimensions the data is stored on.

climakitae/selectors.py
thenaomig: This time, though, it is called multiple times, once for each scenario.

climakitae/selectors.py
thenaomig: Agreed, except that LOCA is being trained on this same WRF data, so the years will be the same. Will make a card about going back and generalizing it though.

PR #42, opened 2022-05-02
Add missing deps and pin package versions

bkg:
This adds dependencies on intake-xarray and s3fs which are required for reading the Zarr stores from the intake catalog. It also pins the packages to the versions tested against on the container image.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,7 @@ geopandas==0.10.2
 geoviews==1.9.2
 holoviews==1.14.6
 hvplot==0.7.3
+intake-xarray==0.5.0
 lmoments3
 matplotlib==3.4.3
 numpy==1.21.2
@@ -12,5 +13,6 @@ pandas==1.3.4
 panel==0.12.4
 param==1.11.1
 regionmask
+s3fs==2021.10.1
 scipy==1.7.1
 shapely==1.7.1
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -1,2 +1,18 @@
...
+geopandas==0.10.2
+geoviews==1.9.2
+holoviews==1.14.6
+hvplot==0.7.3
+intake-xarray==0.5.0
+lmoments3
+matplotlib==3.4.3
+numpy==1.21.2
+pandas==1.3.4
+panel==0.12.4
```
thenaomig, line 13: While we're updating these, I think there are additional dependencies to get the panel stuff to work: https://panel.holoviz.org/user_guide/Server_Deployment.html

thenaomig (commented):
This is great, just wondering if we want to add a couple more of these at once?

requirements.txt
bkg: Only the minimal set of application dependencies should be specified here, essentially everything that's imported, and not the dependencies of dependencies unless there are particular versions to avoid that are known to break things. I'm not seeing any more import errors on my end due to missing packages when installing in a clean virtualenv. The panel server deployment only applies to Panel apps and dashboards, not notebook usage of it.

thenaomig: Okay, we'll make a separate issue of why the panels are not working outside of the pangeo docker image.

PR #44, opened 2022-05-19
confidence intervals for threshold_tools

nabigac:
added confidence intervals to threshold_tools functions

updates:
- added new function "bootstrap"
- added bootstrap capabilities to "return_value", "return_prob", and "return_period" capabilities
- changed ValueError exceptions to also include ZeroDivisionError (_note: this does not impact calculations but removes an error that would occur with the gumbel distribution and instead outputs an np.nan for that gridcell if either of these two errors occurs_).
- ran black for formatting

### DIFF
--- a/climakitae/threshold_tools.py
+++ b/climakitae/threshold_tools.py
@@ -258,7 +258,7 @@ def get_ks_stat(ams, distr="gev", multiple_points=True):
 
 #####################################################################
 
-def calculate_return(fitted_distr, data_variable, arg_value)
+def calculate_return(fitted_distr, data_variable, arg_value):
     """
     Returns corresponding extreme value calculation for selected data variable.
     """
--- a/climakitae/threshold_tools.py
+++ b/climakitae/threshold_tools.py
@@ -4,9 +4,6 @@
 #                                      #
 ########################################
 
-#####################################################################
-# import packages
-
 import numpy as np
 import pandas as pd
 import xarray as xr
@@ -30,12 +27,11 @@ import hvplot.xarray
 from .visualize import get_geospatial_plot
 
 #####################################################################
-# GET AMS
-# input: data array
-# export: data array of maximums
-
 
 def get_ams(da, extremes_type="max"):
+    """
+    Returns a data array of annual maximums.
+    """
 
     extremes_types = ["max"]
     if extremes_type not in extremes_types:
@@ -59,14 +55,12 @@ def get_ams(da, extremes_type="max"):
 
     return ams
 
-
-#####################################################################
-# GET L-MOMENTS DISTR
-# input: distribution name
-# export: corresponding l-moments distribution function
-
+#####################################################################a
 
 def get_lmom_distr(distr):
+    """
+    Returns corresponding l-moments distribution function from selected distribution name.
+    """
 
     distrs = ["gev", "gumbel", "weibull", "pearson3", "genpareto"]
     if distr not in distrs:
@@ -93,12 +87,11 @@ def get_lmom_distr(distr):
 
 
 #####################################################################
-# GET L-MOMENTS
-# input: annual maximum series
-# export: xarray dataset of l-moment ratios
-
 
 def get_lmoments(ams, distr="gev", multiple_points=True):
+    """
+    Returns dataset of l-moments ratios from an inputed maximum series.
+    """
 
     lmom_distr = get_lmom_distr(distr)
     ams_attributes = ams.attrs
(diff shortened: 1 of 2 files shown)


climakitae/threshold_tools.py
```diff
@@ -558,32 +937,153 @@ def return_period(ams):
...
+    ] = "{}th percentile".format(str(conf_int_upper_bound))
+
     new_ds.attrs = ams_attributes
     new_ds.attrs["distribution"] = "{}".format(str(distr))
 
     return new_ds
 
 
 #####################################################################
+# ARCHIVE
```
thenaomig, line 994: Better to delete the old version. With version-control you can always get it back. Or just store it somewhere else easier for yourself just to keep the codebase cleaner.

climakitae/visualize.py
```diff
@@ -134,3 +116,31 @@ def get_geospatial_plot(
     )
 
     return geospatial_plot
+
+#####################################################################
+# ARCHIVE
```
thenaomig, line 121: same comment again: just remove this to keep down clutter

thenaomig (commented):
For now, just some comments about repetition of code blocks (pull these out into functions), and to delete rather than commenting-out older stuff. Still need to try it out in a notebook. 

There are also some blocks of if statements that should be elif after the first one, since it's mutually exclusive, e.g. which distribution you're choosing.

climakitae/threshold_tools.py
```diff
@@ -49,6 +61,9 @@ def get_ams(da, extremes_type="max"):
 
 
 #####################################################################
+# GET L-MOMENTS DISTR
+# input: distribution name
+# export: corresponding l-moments distribution function
```
thenaomig, line 66: While this formatting looks nice, I think it's better to put blocks of 
'''
This kind of comment
'''
just after function def lines, because that standard format will get grabbed to create automatic documentation, which #comments will not.

climakitae/threshold_tools.py
nabigac: great catch! removed!

climakitae/visualize.py
nabigac: removed!

climakitae/threshold_tools.py
nabigac: updated function comments to this formatting. thanks for catching this and explaining!

climakitae/threshold_tools.py
```diff
@@ -337,79 +465,180 @@ def return_value(ams):
...
+                    arg_value=return_period,
+                )
+                bootstrap_values.append(result)
+
+            conf_int_array = np.percentile(
+                bootstrap_values, [conf_int_lower_bound, conf_int_upper_bound]
+            )
+            conf_int_lower_limit = conf_int_array[0]
+            conf_int_upper_limit = conf_int_array[1]
+
```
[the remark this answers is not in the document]
nabigac: moved all this to a new function and made the rest of the changes! i also tested and it seems like everything works but let me know once the rest looks good on your end (or if you have any additional edits) @thenaomig and then i'll merge.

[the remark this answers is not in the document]
thenaomig: Cool. I still see some other blocks of code that are repetitive though, in the if-statements for different distributions. You could put everything that follows into a function that is passed the distribution name, for example.

gdicecco:
Tried commit 68858f7 out in a notebook and I didn't run into any errors/warnings with temperature and precip and different distributions

With snowfall (area subset = none, SSP 3-7.0, monthly, 45km), I got this warning with `return_value = threshold_tools.get_return_value(ams, return_period=10, distr='gev', bootstrap_runs = 10, multiple_points=True)`:
`/srv/conda/envs/notebook/lib/python3.9/site-packages/lmoments3/__init__.py:179: RuntimeWarning: invalid value encountered in double_scalars
  l3 = coefl3 * sum_xtrans / l2`

Some things that would be nice to have:
- Add warning text in notebooks with functions producing confidence intervals that things may take a while depending on the number of bootstrap iterations
- Show upper and lower confidence intervals for the mapped value in the interactive boxes that pop up on the plots

climakitae/threshold_tools.py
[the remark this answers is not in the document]
nabigac: added two new functions to streamline code and cover more repetitive code blocks! let me know if you see any other issues! @thenaomig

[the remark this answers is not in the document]
nabigac: also updated if statements to follow if, elif pattern!

PR #49, opened 2022-06-06
Fix census county url

bkg:
This fixes the 400 bad request thrown by the US Census ArcGIS Server by adding quotes to the state query which was being parsed as an integer type instead of a string. This was likely due to an ArcGIS Server update which now treats these query parameters differently as it was working previously.


thenaomig (commented):
Looks good. I'm also going to update this in PR#45 as the url location will move. Thanks for tracking that down.

PR #50, opened 2022-07-06
Add testing and derived variables

nicolejkeeney:
**The primary purpose of this branch is to add derived variables and add tests (& a testing workflow).** Major changes include: 

- **A new workflow file** (.github/workflows/ci.yaml) that automates testing via continuous integration in the Actions tab. I've built a conda environment for testing (environment.yaml) based on the Pangeo docker image from October 2021 (used in the hub as well) containing all relevant packages. 
- **A new module derive_variables.py**. This module computes relative humidity, total precipitation, and wind magnitude. 
- **A new testing directory** containing the following: 
       - notebook to generate a small testing dataset
       - small testing dataset (netcdf file)
       - tests for the derived variables functions 

**Some notes:** 
- I've left out wind direction because we are still exploring options for the best way to compute this variable. We initially tried with the metpy wind direction function, which was incompatible with dask. 
- The notebook to create the test dataset might be better off as a script, but I like that you can observe the geographic extent of the dataset in the notebook. 
- The tests are currently set up to run in 3 operating systems: linux, windows, and osx using the matrix functionality in the workflow file. I think this is fun and cool but it's a bit slow and maybe not necessary. 
- The workflow file is set up to run tests only in this branch, since it's the only branch with tests. We may want to change this if/when we merge changes to main. 
- Victoria and I added a picture of an otter to learn how to use github pushing/pulling on our new branch, which I now deeply regret... I didn't realize at the time that this would stay in our history after removing the files, and now I don't know how to remove them from the git history... :( Sorry about that


ckingdon95 (approved):
approving because otter jpeg is only 10KB :)

ckingdon95:
> * The notebook to create the test dataset might be better off as a script, but I like that you can observe the geographic extent of the dataset in the notebook.

I agree maybe we should just have py scripts for generating test data, but that this is something we could do in a later PR to clean-up/re-organize our testing set up, no reason to not merge this as is!

nicolejkeeney:
> > * The notebook to create the test dataset might be better off as a script, but I like that you can observe the geographic extent of the dataset in the notebook.
> 
> I agree maybe we should just have py scripts for generating test data, but that this is something we could do in a later PR to clean-up/re-organize our testing set up, no reason to not merge this as is!

I agree with your comment that it would be better to have this in a script rather than a notebook. See my comment in your PR for my suggestion.

nicolejkeeney:
@ckingdon95 I just converted the notebook to a script, and deleted the notebook. Can you take a look when you get a chance? :D

ckingdon95:
Looks perfect thank you!

elehmer:
So...I'm still having trouble figuring out how to invoke these tests. Feeling dumb. How are you doing that?

When I consult the pytest documentation it says it's looking for files named 'test_*.py'  or '\*_test.py' in the current dir or subdirs. The files named here aren't to that standard.

Anyway, how do you run these tests in the Notebook environment. Thanks!

tests/create_test_dataset.py
```diff
@@ -0,0 +1,145 @@
...
+
+# Merge to form a single xr.Dataset with several data variables
+xr_ds = xr.merge(xr_da_list)
+
+
+# ----------------- SUBSET DATA TO A SMALLER GEOGRAPHIC REGION -----------------
+
+# To avoid having a huge dataset, we just want to use a geographic subset. 
+# At the time of writing this script (June 20, 2022), the subsetting function built into climakitae was broken. # So, instead we just subset the data using xarray. To do this, we construct a simple lat/lon bounding box of any region we're interested, and filter the data from the dataset using .where()
+# This section constructs a box between San Diego, CA and Joshua Tree, CA
```
thenaomig, line 106: It's nice to have it be a real place of interest, but it'd be much faster to generate a smaller testing dataset if we avoid using .where, and just .sel(x=slice(arbitrary,arbitrary2),y=slice(arbitrary3,arbitrary4))

tests/create_test_dataset.py
```diff
@@ -0,0 +1,145 @@
...
+    """Display a progress bar inside a for loop 
+    Based on Stack Overflow answer from Mark Rushakoff: https://stackoverflow.com/questions/3002085/how-to-print-out-status-bar-and-percentage
+    
+    Args: 
+        i (int): iteration number
+        tot (int): total number of iterations
+    """
+    j = (i+1) / tot
+    sys.stdout.write('\r[%-20s] %d%% complete' % ('='*int(20*j), 100*j))
+    sys.stdout.flush()
```
thenaomig, line 71: perhaps this belongs in a more generic utilities file, as we'll want to use progress bars for other things, and the previous way @bethem implemented it was not compatible with dask

thenaomig (changes requested):
Just very minor suggestions, and let me know if you need help deciding about any merge conflicts.

tests/create_test_dataset.py
ckingdon95: (Note from discussion with Naomi and Nicole): this progress bar only works for for-loops, so we will leave it here for now

tests/create_test_dataset.py
ckingdon95: (Note from discussion with Naomi and Nicole): it's okay to leave as is because this file is only run once

nicolejkeeney:
> So...I'm still having trouble figuring out how to invoke these tests. Feeling dumb. How are you doing that?
> 
> When I consult the pytest documentation it says it's looking for files named 'test_*.py' or '*_test.py' in the current dir or subdirs. The files named here aren't to that standard.
> 
> Anyway, how do you run these tests in the Notebook environment. Thanks!

In regards to the test file names: there's only one file containing tests, named **test_derived_variables.py**. The file **conftest.py** points pytest to the directory containing the test data, and the file **create_test_dataset.py** creates the testing file (not part of the pytest workflow). 

In regards to running the tests in the Notebook environment, you just need to make sure climakitae is installed and then run the tests **within the climakitae directory or within the tests directory**. Either is fine. Then you just run `pytest` in terminal. It should look something like this in terminal when it runs: 

Screen Shot 2022-07-21 at 11 56 29 AM

elehmer:
> > So...I'm still having trouble figuring out how to invoke these tests. Feeling dumb. How are you doing that?
> > When I consult the pytest documentation it says it's looking for files named 'test__.py' or '__test.py' in the current dir or subdirs. The files named here aren't to that standard.
> > Anyway, how do you run these tests in the Notebook environment. Thanks!
> 
> In regards to the test file names: there's only one file containing tests, named **test_derived_variables.py**. The file **conftest.py** points pytest to the directory containing the test data, and the file **create_test_dataset.py** creates the testing file (not part of the pytest workflow).
> 
> In regards to running the tests in the Notebook environment, you just need to make sure climakitae is installed and then run the tests **within the climakitae directory or within the tests directory**. Either is fine. Then you just run `pytest` in terminal. It should look something like this in terminal when it runs:
> 
> Screen Shot 2022-07-21 at 11 56 29 AM

Ok, now I feel silly! Just use the terminal, of course! I was trying to do this with a notebook...

PR #51, opened 2022-07-11
Fix monthly averages in Timeseries tools

ckingdon95:
This pull request makes two changes to the Timeseries tools functionality:
1. When computing a running mean with monthly data, the `_running_mean()` helper function has been updated to use the number of days in each month as weights. 
2. When the user requests de-meaned data against an average across a reference range, the `_getAnom()` helper function has been updated to take a weighted average for the reference range in the case of monthly data. 

I have also added the following files for testing:
- `"tests/test_timeseriestools_monthly.py"`: test script. As noted at the top of the file, I am only testing that various user-parameter combinations run without error in the case of monthly data, and that the data are transformed. _I am not testing for actual expected values_-- we can discuss if we want to be doing this somehow.
- `"tests/test_data/timeseries_data_T2_2014_2016_monthly_45km.nc"` this is the saved test data set, it is only 36KB

Note: I have merged the `add_var_test` branch into this branch for testing purposes, so after PR #50 is merged into main, this PR will look smaller and we can more easily review just the changes I have added (my changes are only in `"timeseriestools.py"` and the three added test files).


nicolejkeeney:
I'm not sure why you need this extra script tests/test_data/timeseries_data_T2_2014_2016_monthly_45km.nc to generate a testing dataset? The notebook create_test_dataset.ipynb should be able to do this, you'd just need to change the variables in the second cell to the following: 

`append_historical = True
variable_list =  ["T2"]
year_start=2014
year_end=2016
`

This way we avoid having two different files with essentially the same code.

nicolejkeeney:
Your tests all look great and awesome work on your documentation! Code is super easy to read.

ckingdon95:
> I'm not sure why you need this extra script tests/test_data/timeseries_data_T2_2014_2016_monthly_45km.nc to generate a testing dataset? The notebook create_test_dataset.ipynb should be able to do this, you'd just need to change the variables in the second cell (see attached screenshot). This way we avoid having two different files with essentially the same code. The settings you'd want to change in the notebook are the following:

Yeah we can definitely go that route, and agree it's bad to have code repetition across files. My only concern was with it being explicitly reproducible. If I modify the notebook file with those parameters to generate my test dataset, then it no longer explicitly shows how we create the first dataset (I like the idea of someone being able to just hit "run" and it produces exactly the same result--both datasets). But maybe the obvious compromise would be for me to just add a new section at the bottom of the notebook to produce mine? Then someone can re-run the whole notebook and it would generate both datasets?

nicolejkeeney:
> Yeah we can definitely go that route, and agree it's bad to have code repetition across files. My only concern was with it being explicitly reproducible. If I modify the notebook file with those parameters to generate my test dataset, then it no longer explicitly shows how we create the first dataset (I like the idea of someone being able to just hit "run" and it produces exactly the same result--both datasets). But maybe the obvious compromise would be for me to just add a new section at the bottom of the notebook to produce mine? Then someone can re-run the whole notebook and it would generate both datasets?

How do you feel about deleting the notebook that creates the file, and I can convert it to a script? Then we delete your script and have the options to create each but just comment out one of them at the top? Something like this: 

`year_start = 2022 # Use to generate test_dataset_2022_2022_monthly_45km.nc`
`# year_start = 2014 # Use to generate timeseries_data_T2_2014_2016_monthly_45km.nc `

Not sure if this is a better option or messier. But I feel like people wouldn't be regenerating these datasets very often (if at all) so it might be cleaner to keep everything in one script.

ckingdon95:
@nicolejkeeney yes this sounds great to me! I would say at the top of the file we can just group the parameters into chunks for each test set like:
```python

#------------- parameters for test file "name1"

append_historical = True
variable_list = ["T2"]
year_start=2014
year_end=2016
timescale="monthly"
resolution="45km"
scenario_list = ['SSP 2-4.5 -- Middle of the Road', 'Historical Climate']
filename="name1"

#------------- parameters for test file "name2"
. . .
```
Noting that we'll want to add scenarios as another argument. I also generally like autogenerating the file names, but we may also want the option to just handname sometimes.

nicolejkeeney:
What is the difference between None and running mean here? Neither appears to change the plot displayed or the output after changing the selection. Perhaps something is happening under the hood but I'm not familiar with what the expected behavior is. 

Screen Shot 2022-07-19 at 6 30 12 PM

ckingdon95:
@nicolejkeeney ah yeah, you have to change the number of timesteps to be something other than 0 with the slider below. That indicates the size of the window for the rolling average. (this also took me a while to notice the first time I rendered it; perhaps this is a confusing layout we can revisit somehow)

nicolejkeeney (approved):
This all looks good to me, assuming the unit tests run without error.

vicford:
Finally got this working on the Hub (maybe 9pm my time is the golden hour) - everything checks out for me scientifically and mathematically. Nicely done!! 

One suggestion that is more related to a non-climate user perspective, and not specifically for rolling averaging but comes up when visualizing it: When selecting time.season when disaggregating seasons, I'd recommend adding the season name with the month bounds in parentheses, e.g. Summer (JJA), Spring (MAM), and so on. Users may likely figure this out from context clues, but we're here for clarity in usability! (This can obviously be rolled into another update, and not this one specifically)

PR #53, opened 2022-07-15
Reproject watershed polygon to WGS84 before regionmask

elehmer:
This PR reprojects the Watershed shape selected from GUI from Pseudo Mercator to WGS84 to be compatible with the other boundary files used.

I have tested this on the HUB. I picked the "Carrizo Plain" watershed and works fine for 3km and 9km, but the 45km returns the "Insufficient Gridcells" assertion. I guess that is because the watershed is in 1 cell and not getting the centroid? Or some similar behavior (not majority?). Is this the behavior we want or should we work on having it return any cells it intersects instead?

Please test this in your own environments as well. Thanks.

### DIFF
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -2,6 +2,7 @@ import xarray as xr
 import cartopy.crs as ccrs
 import pyproj
 from shapely.geometry import box
+from shapely.ops import transform
 import regionmask
 import intake
 import numpy as np
@@ -137,6 +138,10 @@ def _read_from_catalog(selections, location):
         if location.area_subset == "CA watersheds":
             shape = location._geographies._ca_watersheds
             shape = shape[shape["OBJECTID"] == shape_index].iloc[0].geometry
+            wgs84 = pyproj.CRS('EPSG:4326')
+            psdo_merc = pyproj.CRS('EPSG:3857')
+            project = pyproj.Transformer.from_crs(psdo_merc, wgs84, always_xy=True).transform
+            shape = transform(project, shape)
         elif location.area_subset == "CA counties":
             shape = location._geographies._ca_counties
             shape = shape[shape.index == shape_index].iloc[0].geometry


vicford:
Hi @elehmer - I've tested this on the Hub, and it works fine for the 9km and 3km data resolution for the Carrizo Plain watershed. When testing with 3km res, I also got the "insufficient gridcells" error, so I tried with a slightly larger watershed (Big-Navarro-Garcia) which passes app.retrieve() but fails upon basic map plotting with this error: 

`WARNING:param.dynamic_operation: Callable raised "AxisError('axis 1 is out of bounds for array of dimension 1')".
Invoked as dynamic_operation(numpy.datetime64('1980-09-01T00:00:00.000000000'), 'Historical Climate', 'cesm2', height=400, scale=1.0, width=400, x_range=None, y_range=None)
WARNING:param.dynamic_mul: Callable raised "AxisError('axis 1 is out of bounds for array of dimension 1')".
Invoked as dynamic_mul(numpy.datetime64('1980-09-01T00:00:00.000000000'), 'Historical Climate', 'cesm2')
WARNING:param.dynamic_mul: Callable raised "AxisError('axis 1 is out of bounds for array of dimension 1')".
Invoked as dynamic_mul(numpy.datetime64('1980-09-01T00:00:00.000000000'), 'Historical Climate', 'cesm2')
WARNING:param.dynamic_operation: Callable raised "AxisError('axis 1 is out of bounds for array of dimension 1')".
Invoked as dynamic_operation(numpy.datetime64('1980-09-01T00:00:00.000000000'), 'Historical Climate', 'cesm2')`

with `AxisError: axis 1 is out of bounds for array of dimension 1` as well. 

Potentially related to the "one cell" issue by another name?

elehmer:
The hvplot documentation is not very detailed unfortunately. But looks like the quadmesh function does not like a 1 dimensional array - it assumes that it will be at least 2 cells wide in an axis. Which is a bad assumption since a 1 cell axis is still an axis. Arg! I guess we could put a test on the quadmesh to not run if the axis only has 1 cell.

thenaomig (approved):
Looks good. Just a note to avoid this kind of hard-coded transform in future by saving our own version of this already transformed in a future update.

thenaomig:
Made a note about handling the one-gridcell in the card about determining the type of plot to make, but seems like a side issue that should not hold up this PR.

elehmer:
Yes, I'd actually just like to fix this issue by reprojecting and then serving the data ourselves. Maybe using ArcGIS Server? I have a server running. Or convert to geojson and stash somewhere?

elehmer:
Yeah the dimensionallity thing is annoying, but as I said on slack `climakitae` shouldn't care where the spatial dims are 0,1,2 dims. But we can add some logic in the notebooks to handle the weird 1d situation.

PR #54, opened 2022-07-21
Rewired Watersheds with WGS84 Geojson and stopped reprojection

elehmer:
This PR rewires the CA Watershed boundaries file to use a WGS84 geojson reprojection served out of a GitHub repo:

https://github.com/cal-adapt/cae-data

It then removes the unnecessary reprojection from the code base.

### DIFF
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -140,10 +140,6 @@ def _read_from_catalog(selections, location, cat):
         if location.area_subset == "CA watersheds":
             shape = location._geographies._ca_watersheds
             shape = shape[shape["OBJECTID"] == shape_index].iloc[0].geometry
-            wgs84 = pyproj.CRS('EPSG:4326')
-            psdo_merc = pyproj.CRS('EPSG:3857')
-            project = pyproj.Transformer.from_crs(psdo_merc, wgs84, always_xy=True).transform
-            shape = transform(project, shape)
         elif location.area_subset == "CA counties":
             shape = location._geographies._ca_counties
             shape = shape[shape.index == shape_index].iloc[0].geometry
--- a/climakitae/selectors.py
+++ b/climakitae/selectors.py
@@ -57,7 +57,7 @@ class Boundaries:
         )
         self._ca_counties = self._ca_counties.sort_values("NAME")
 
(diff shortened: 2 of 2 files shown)


elehmer:
This works well. I assume GitHub can take the demand for serving these data? If so we can convert the other boundary data to geojson and store in the `cae-data` repo as well.

vicford:
Unsure if this is related or not, but in testing I am getting specifically these errors:
- On using 45km spatial res for Carrizo Plain and Goose Lake watersheds on app.retrieve()
 ```
/srv/conda/envs/notebook/lib/python3.9/site-packages/climakitae/data_loaders.py:76: UserWarning: No gridpoint belongs to any region. Returning an all-NaN mask.
  mask = ds_region.mask(data.lon, data.lat, wrap_lon=False)
```
^ these are "small" watersheds


- On using 45km spatial res for Big-Navarro-Garcia (big watershed) on plotting (our favorite error)
`AxisError: axis 1 is out of bounds for array of dimension 1`

elehmer:
> Unsure if this is related or not, but in testing I am getting specifically these errors:
> 
> * On using 45km spatial res for Carrizo Plain and Goose Lake watersheds on app.retrieve()
> 
> ```
> /srv/conda/envs/notebook/lib/python3.9/site-packages/climakitae/data_loaders.py:76: UserWarning: No gridpoint belongs to any region. Returning an all-NaN mask.
>  mask = ds_region.mask(data.lon, data.lat, wrap_lon=False)
> ```
> 
> ^ these are "small" watersheds
> 
> * On using 45km spatial res for Big-Navarro-Garcia (big watershed) on plotting (our favorite error)
>   `AxisError: axis 1 is out of bounds for array of dimension 1`

This is the unrelated issue that when I boundary shape is smaller than or doesn't intersect any centroid (not sure which method is being used) of the xarray input we are getting a NaN back. This maybe be acceptable condition or we can and change it so that any intersection brings back at least 1 cell?

vicford:
> 

Scientifically, I think I am torn between a single cell return for data usefulness reasons and the logic behind "smaller resolutions ~roughly~ averaged to coarser resolution would result in a single cell at some level". 

Maybe we can add some sort of text flag when watersheds are selected that coarser resolutions may result in a single (or few) cell/data point(s), and that selecting a finer resolution may be more beneficial to the user's needs, but go ahead with the intersection of 1 grid cell working for data return?

PR #45, opened 2022-05-20
Dynamically populate param classes

thenaomig:
The primary purpose of this PR is to speed up the import of climakitae. It achieves this by:
- delaying the reading of the catalog to the call to ck.Application()
- delaying the reading of area-selection shapefiles to ck.Application()
As a result, ck.Application() is slower to execute, but climakitae imports immediately, as one might expect it to.

This was accomplished by learning to pass parameter values to classes that inherit from param.Parameterized, and adding an __init__ on top of the inherited one from the metaclass for both DataSelector and LocSelectorArea classes in selectors.py. 

Then the resulting changes in core.py and data_loaders.py were brought into agreement with the new data organization. The catalog is read in core.py now, and a get_catalog_contents function was added there to get a dictionary of choices to pass to select(); this is in place of the data class that held the same content previously (plus the catalog) in selectors.py

PR #56, opened 2022-07-26
Fix testing bug

nicolejkeeney:
**Description:**
This PR fixes the file create_test_dataset.py to be compatible with code change from PR #45 to the structure of core.py and data_loaders.py. 

**How to review:**
Just make sure the script create_test_dataset.py runs without error. To do this, ensure you have climakitae installed, migrate to the tests/ directory, and run `python create_test_dataset.py` in terminal.

**Notes for future reference:** 
**EDIT**: Cora fixed this but I'm leaving the text here for reference. 
I tried to write a test for this, but it throws a super weird error. I ran into this same error a lot when troubleshooting issues with create_test_dataset.py. I was able to write the script in such a way as to not cause that error, but I'm honestly not sure why/how it works, considering the test I wrote tests the function from that script. This might be good to diagnose at a later date when I have more free time to look into this. I believe it has to do with connecting with and storing catalog information. I pushed the test to the repository (it's called test_create_test_dataset.py) but deleted it (because it doesn't run without error). 

The error message from that test: 
error_message


ckingdon95 (approved):

vicford (approved):
Spent some time looking this over -- seems straightforward to me!

PR #58, opened 2022-07-27
Unit Conversion Options

vicford:
Hi all, 
Different unit options are now available for a range of variables as the following: 
- Temperature variables can now be selected in app.selected() as the native unit K, as well as °C, °F, and °R
- Wind variables can be selected as native unit m/s, and knots
- Pressure can be selected as native Pa, and now hPa, mb, inHg
- Precipitation variables can be selected as native mm, and now also inches
- Moisture ratio can be selected as native kg/kg, and now g/kg
Please test these out, and if different units need to be added or removed from the list. 

@nicolejkeeney is writing one last additional test to ensure different spellings of alternative units checks out ("kts" vs. "knots")

How this modifies app.select(): 
Screen Shot 2022-07-27 at 12 13 51 PM


nicolejkeeney:
If anyone has any ideas for additional unit conversions they think would be useful, feel free to suggest them! It's pretty easy to add more now that the code base is all set up

climakitae/selectors.py
```diff
@@ -350,8 +355,21 @@ def __init__(self, **params):
...
         ]
-        self.variable = self.default_variable
-
+        self.units = self.descrip_dict[self.variable]["native_unit"]
+        self.native_units = self.descrip_dict[self.variable]["native_unit"]
+
+    @param.depends("variable","native_units","units","descrip_dict", watch=True)
+    def _update_unit_options(self): 
+        """ Update unit options and native units for selected variable. """
+        _default_unit = self.descrip_dict[self.variable]["native_unit"]
```
thenaomig, line 364: I don't like that the native_unit is read from the .csv instead of the catalog metadata. It's also not something that the user will ever actually select, so it doesn't make sense to have it as a param in DataSelector. Can you explain this choice or change it?

climakitae/unit_conversions.py
```diff
@@ -0,0 +1,64 @@
...
+
+    # Temperature units
+    elif native_units == "K":
+        if selected_units == "degC":
+            da = da - 273.15
+            da.attrs["units"] = selected_units
+        elif selected_units == "degF":
+            da = (1.8 * (da - 273.15)) + 32
+            da.attrs["units"] = selected_units
+        elif selected_units == "degR":
```
thenaomig, line 60: I am skeptical that anyone would actually use the degR option, and so it might cause more confusion that its worth.

thenaomig (changes requested):
Overall great, and love the tests. A couple of comments. Mostly concerned about reading the native units from the .csv instead of the catalog metadata -- happy to have a conversation about it if that's easier.

climakitae/unit_conversions.py
nicolejkeeney: degR is removed! Good point. I felt a weird affinity toward that strange unit but it's probably too confusing for users.

climakitae/selectors.py
nicolejkeeney: I would agree with this. Had a hard time figuring out the best way to incorporate this feature into the code. I'll take a look at restructuring the code to read from the catalog instead. Thanks for your input!

nicolejkeeney: I'll let you know once I've implemented this change so you can give it another look

nicolejkeeney: These changes have now been implemented! :D

thenaomig (approved):
Looks good.

climakitae/unit_conversions.py
```diff
@@ -0,0 +1,74 @@
...
+    # Pressure units
+    elif native_units == "Pa":
+        if selected_units == "hPa":
+            da = da / 100.
+            da.attrs["units"] = selected_units
+        elif selected_units == "mb":
+            da = da / 100.
+            da.attrs["units"] = selected_units
+        elif selected_units == "inHg":
+            da = da / 3386.39
```
ckingdon95, line 56: was there any discussion about precision for these? I think this one is an approximation, and just want to check that we are intentionally choosing our number of sig figs!

climakitae/unit_conversions.py
```diff
@@ -0,0 +1,74 @@
...
+            da.attrs["units"] = selected_units
+
+    # Wind units
+    elif native_units == "m/s":
+        if selected_units == "knots":
+            da = da * 1.94
```
ckingdon95, line 62: same for this one, just want to check that we are intentionally choosing only two decimal places!

ckingdon95 (commented):
This is amazing! I tested this out in a notebook and it looks awesome, great job!! I just had one question about the level of precision for the "inHg" and "knots" conversions, which I think are the only conversions that are approximations (see comments)

climakitae/unit_conversions.py
vicford: Same thing here: I'll update with the official NWS formula (https://www.weather.gov/media/epz/wxcalc/windConversion.pdf) which goes out to 7 decimal places.

climakitae/unit_conversions.py
vicford: This is a very good catch @ckingdon95! This conversion matches what the NWS uses in their own calculations (https://www.weather.gov/media/epz/wxcalc/pressureConversion.pdf), but is the opposite direction (will fix now!) so I would feel confident in also doing so, once I've fixed it.

ckingdon95: Ah okay great! (I think you have the direction correct here though?) And maybe we could add links to these sources as a comment at the top of the file?

vicford: @ckingdon95 and I confirmed that multiplying by 0.000295300 is the same operation as dividing by 3386.39, but for our purposes to better follow NWS guidelines that the multiplication option is the way to go. 

Links have been put in the docstring for documentation purposes.

ckingdon95 (approved):

PR #59, opened 2022-07-29
Remove unit test due to unresolved conflict

nicolejkeeney:
**This PR makes only one change: it deletes a unit test that is throwing unresolved errors.** 

The test that throws the error will still be in our git history should someone else want to take another stab at this sometime in the future, but at the moment, I don't think it's a priority. The create_test_dataset script runs fine, this error only impacts the unit test. 

More information about this error is included in the PR comments.


nicolejkeeney:
Information for future reference (should another brave soul try to resolve this mysterious error): I built a script (create_test_dataset.py) to generate a testing dataset by breaking down the functions used in core.py. This code relies on retrieving the catalog contents from AWS, in order to grab the desired data. My current theory is that this information is not always stored in a permanent fashion, and that in the process of trying to grab the data, the catalog contents change somehow. This is based on the fact that the error is thrown at seemingly random times (i.e. not reproducible), making me think there's some sort of timing out issue going on. The error message it throws is shown below. The test that throws the error will still be in our git history should someone else want to take another stab at this sometime in the future, but at the moment, I don't think it's a priority. The create_test_dataset script runs fine, this error only impacts the unit test. 

test error

ckingdon95 (approved):

PR #60, opened 2022-08-09
Threshold tools `app.explore2.thresholds()` function

ckingdon95:
This PR implements the thresholds explore panel, to be called by:
```python
import climakitae as ck
app = ck.Application()
app.explore2.thresholds()
```

The various user options are currently better explained by the four examples in this notebook: https://github.com/cal-adapt/cae-notebooks/blob/exceedance_ntbk/threshold_tools_exceedance.ipynb


climakitae/threshold_tools.py
```diff
@@ -884,3 +887,346 @@ def return_period(ams):
...
+        groups = [math.floor(d / group_len) for d in days] # group the day numbers based on user-specified group length
+        date_ids = dates[[groups.index(i) for i in list(set(groups))]] # save one date value for each group to use as the time index after summing
+        da["time"] = groups # set the group numbers as the time dimension for summing
+        group_totals = da.groupby("time").sum() # sum across the groups
+        group_totals["time"] = pd.to_datetime(date_ids) # reset the time dimension to the saved date values
+
+    else:
+        raise ValueError("Groupby options other than 'day' not yet implmented.")
+
+    return group_totals
```
ckingdon95, line 1053: This is the helper function I would like someone to review and possibly discuss with!

bkg: Aggregating on the temporal dimension can be handled with [.resample()](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.resample.html) which works with various date offset aliases from Pandas and Xarray. You can sum over 3 days:
``` python
da2 = da.resample(time='3D').sum()
```
For annual, `A` and `Y` are synonymous:
``` python
da2 = da.resample(time='A').sum()
```

ckingdon95: perfect, thanks so much! will work on switching to this instead.

vicford:
This tracks Cora! Tested all 4 exceedance examples on the hub with the following options:
- Hourly, 2020-2050
- SSP3-7.0 with no historical data appended
- Air temp in degF
- 9km spatial res, area averaged
- County level: Fresno;  State: CA

I think the different examples of how to use the tool will be exceedingly useful. 

Just going to note that this fails on 45km spatial resolution for me, likely due to the ongoing spatial subsetting issues and should be automatically resolved. This also fails when historical data is included (either on its own, or being appended to a future scenario, likely due to the presence of NaN values).

ckingdon95:
Thanks for testing this out @vicford !

> This also fails when historical data is included (either on its own, or being appended to a future scenario, likely due to the presence of NaN values).

Just checking, is this failure you're referencing with historical data also during `app.retrieve` or `compute` or is it something you're seeing during `explore_exceedance`?

vicford:
> historical data also during `app.retrieve` or `compute`

@ckingdon95 It's at the app.retrieve() stage - and I'm fairly certain it's related to the ongoing geospatial subsetting issues that we're seeing in main and elsewhere right now.

vicford (approved):
Historical and 45km data now working -- looks great Cora!

climakitae/threshold_tools.py
```diff
@@ -8,6 +8,9 @@
 import pandas as pd
 import xarray as xr
 from scipy import stats
+import panel as pn
+import param
+import math
 
 import lmoments3 as lm
 from lmoments3 import distr as ldistr
```
nicolejkeeney, line 16: On this branch in the AE hub, I'm getting an import error when trying to call `import climakitae as ck` due to the import lines here. 

Error message: `AttributeError: module 'scipy.stats._continuous_distns' has no attribute 'frechet_r_gen'`

nicolejkeeney (changes requested):
I'm unable to load climakitae in the AE hub due to package dependencies in threshold tools. The conflicting package import is: `from lmoments3 import distr as ldistr`

I'm surprised this is only happening to me. When you run in the hub, does this package conflict come up for you as well?

tests/test_exceedance.py
```diff
@@ -0,0 +1,120 @@
...
+
+from climakitae import threshold_tools
+
+#------------- Data for testing -----------------------------------------------
+
+@pytest.fixture
+def T2_monthly(test_data):
+    """ Monthly RAINC data for one scenario and one simulation 
+    (pulled from the general test data set)"""
+    return test_data["RAINC"].isel(scenario=0, simulation=0)
```
nicolejkeeney, line 15: It's probably a good time to add the test dataset to the repo now, since we are close to merging and want to ensure the tests pass. That is, unless you are still making changes to that dataset?

climakitae/threshold_tools.py
vicford: I had this same issue: you just have to pip install lmoments in the notebook cell above the import climakitae as ck line and it should resolve :)

vicford: @nicolejkeeney 

!pip install git+https://github.com/OpenHydrology/lmoments3.git

tests/create_test_dataset.py
```diff
@@ -149,11 +167,7 @@ def progressBar(i, tot):
 print("Filename: {0}".format(filename))
 
 if download_data==True: 
-    try: 
-        test_dataset.to_netcdf(path=filepath, mode='w') # Output 
-        print("File saved to: {0}".format(filepath))
-    except:
-        print("You already have a file with the same name: "+filename+"\nDelete that file to output a new one with the same name.")
```
nicolejkeeney, line 156: I did kind of like having a descriptive print statement shown if the download failed due to a pre-existing file with the same name. Am open to removing it like you did here, but curious why you didn't like it?

climakitae/threshold_tools.py
```diff
@@ -884,3 +889,523 @@ def return_period(ams):
...
+        "Examples: for an annual timeseries, select '1-year'. For a seasonal timeseries, select '3-month'.",
+        pn.layout.Divider(margin = (-10,0,-10,0)),
+
+        # GROUP
+        "Optional aggregation: I'm interested in the number of ___ that contain at least one occurance.",
+        choices.group_row,
+
+        # DURATION 2
+        "After aggregation, I'm interested in occurances that last for . . .",
+        choices.duration2_row,
```
nicolejkeeney, line 1393: It's not clear to me by the wording for this selection if this is being performed or not. It says "optional", but a selection is shown. The default shows 2hr, and 2hr, but is labelled as an optional selection, making me unsure whether this aggregation is occurring even if I haven't modified the values?

climakitae/threshold_tools.py
```diff
@@ -884,3 +889,523 @@ def return_period(ams):
...
+                self.param.smoothing,
+                self.param.num_timesteps, 
+                width=375
+            )
+        else:
+            smooth_row = pn.Row(
+                self.param.smoothing,
+                width=375
+            )
+        return pn.Card(smooth_row, title = "Smoothing", collapsible = False)
```
nicolejkeeney, line 1320: The smoothing card looks a little strange a lonely hanging out on its own below the threshold selections panel. I would suggest incorporating it into the threshold events selection Card.

climakitae/explore2.py
```diff
@@ -1,21 +1,67 @@
...
+        )
+
+        plot_panel = _exceedance_visualize(thresh_data, option) # display the holoviz panel
+
+        return pn.Column(
+            pn.Row(
+                data_options_card, description_box
+            ),
+            plot_panel
+        )
```
nicolejkeeney, line 67: My preference would be for this view function to be part of threshold_tools.py, to keep the AppExplore class clean.

nicolejkeeney:
Is the threshold_tools.py file only code for the thresholds panel generated by explore2? If not, I think it's best to break up that module, because it's a bit long and unwieldily. Any code for the thresholds panel should be in a separate module (i.e. threshold_panel.py), and helper functions or non-panel code could be in a threshold_tools.py module.

climakitae/threshold_tools.py
```diff
@@ -884,3 +889,523 @@ def return_period(ams):
...
+    duration2_length = param.Integer(default = 1, bounds = (0, None), label = "")
+    duration2_type = param.ObjectSelector(default = "hour", objects = ["year", "month", "day", "hour"], label = "")
+    smoothing = param.ObjectSelector(default="None", objects=["None", "Running mean"], label = "Smoothing")
+    num_timesteps = param.Integer(default=10, bounds=(0, None), label = "Number of timesteps")
+
+    units2 = param.ObjectSelector(objects=dict())
+    # variable2 = param.ObjectSelector(default="Air Temperature at 2m", objects=["Air Temperature at 2m"])
+    variable2 = param.ObjectSelector(default="Air Temperature at 2m", objects=dict())
+    cached_area2 = param.ObjectSelector(default="Sacramento County", objects=dict())
+    area_subset2 = param.ObjectSelector(default="CA counties", objects=["CA counties", "states"])
```
nicolejkeeney, line 1176: I don't think the default should be Sacramento county (sorry, Owen & Beth). It should probably be CA state, since that will be what more users are interested in seeing.

climakitae/threshold_tools.py
```diff
@@ -884,3 +889,523 @@ def return_period(ams):
...
+    plot_obj = exceedance_count.hvplot.line(
+        x="time", 
+        widget_location="bottom", 
+        by="simulation", 
+        groupby=["scenario"],
+        title = "",
+        fontsize = {'ylabel': '10pt'},
+        legend = 'right',
+    )
+    return pn.Column(plot_obj)
```
nicolejkeeney, line 1104: This plot definitely needs a title. You have some descriptive text above the plot; that would make for a nice title! You could just paste that in here. When you save this plot using the bokeh toolbar, it lacks a title, making it hard to understand what the data represents. I also just think plots look a bit unbalanced without a title, even though you have the two lines of text above the plot.

climakitae/threshold_tools.py
```diff
@@ -884,3 +889,523 @@ def return_period(ams):
...
+        """If the button was clicked and the location, variable, or units were 
+        changed, reload the data from AWS"""
+        if self.changed_loc_and_var:
+            self.da = _read_from_catalog(selections = self.selections, location = self.location, cat = self._cat).compute()
+            self.changed_loc_and_var = False
+        if self.changed_units:
+            self.da = _convert_units(da=self.da, selected_units=self.units2)
+            self.threshold_value = round(self.da.mean().values.item())
+            self.param.threshold_value.label = f"Value (units: {self.da.units})"
+            self.changed_units = False
```
nicolejkeeney, line 1248: It's not clear to me what the purpose of the unit options are. If they are just used in computation, it shouldn't matter if you convert units, because the output plot doesn't show the units from the original variable anywhere (just count vs. time). And, even when I change units (say, from K to degF), the thresholds panel still shows the data in K. 

Unless I'm missing something (very possible), I would suggest removing the unit conversion option all together. 

Or, is it there for the purpose of data downloading, after the panel? If so, I still think it's clearer to just remove the units option all together, unless you can loop it in with the panel options entirely. 

units mismatch

climakitae/threshold_tools.py
```diff
@@ -884,3 +889,523 @@ def return_period(ams):
...
+    def view(self):
+        try:
+            to_plot = self.transform_data()
+            obj = plot_exceedance_count(to_plot)
+        except Exception as e:
+            # Display any raised Errors (instead of plotting) if any of the 
+            # user specifications are incompatible or not yet implemented.
+            return e
+        return pn.Column(
+            pn.widgets.Button.from_param(self.param.reload_plot, button_type="primary", width=150, height=30),
```
nicolejkeeney, line 1298: I think it would make more sense to just have one reload button, instead of two. The reload_plot button should reload the data, and reload the threshold tools selection. 

In the warming levels panel, we've set up a workflow that only reloads the data if the user has made changes to certain selections related to the data. You could do the same here: have one button (reload plot) that reloads the data ONLY IF the user has made changes to the data selections, and updates the plot with the threshold event options always. 

The code from the warming levels panel that I'm referencing is below. It will reload the postage stamp data when changes to the location subset and variable are made, but NOT if the warming level is changed. If only the warming level is changed, the data (loaded in a previous step, and saved in the variable postage_data) is not reloaded, but the anomalies with that data are recalculated. Thus, avoiding reloading the same data, but allowing for only one button to represent the two changes. 


```
@param.depends("area_subset2","cached_area2","variable2", watch=True)
    def _updated_bool_loc_and_var(self):
        """Update boolean if any changes were made to the location or variable"""
        self.changed_loc_and_var = True

    @param.depends("reload_data2", watch=True)
    def _update_postage_data(self):
        """If the button was clicked and the location or variable was changed,
        reload the postage stamp data from AWS"""
        if self.changed_loc_and_var == True:
            self.postage_data = _get_postage_data(area_subset2=self.area_subset2, cached_area2=self.cached_area2, variable2=self.variable2, location=self.location)
            self.changed_loc_and_var = False
        self._warm_all_anoms = get_anomaly_data(data=self.postage_data, warmlevel=self.warmlevel)
```

climakitae/explore2.py
```diff
@@ -1,21 +1,67 @@
...
+                ),
+        title="Data Options", collapsible=False, width=700, height=_first_row_height
+        )
+
+        description_box = pn.Card(
+            _thresholds_tool_description,
+            title = "About this tool", collapsible = False,
+            # width=400, 
+            height = _first_row_height
+        )
```
nicolejkeeney, line 58: I'm a big fan of this description box! Nicely done

tests/test_exceedance.py
ckingdon95: adding now!

tests/create_test_dataset.py
ckingdon95: mm this wasn't working for me properly. I think the file paths weren't the same for me for some reason. Also it was catching all errors (not just file-already-exists errors)... but I don't have strong thoughts, I can add it back and we can revisit later!

climakitae/explore2.py
ckingdon95: mm okay yeah I will do this!

climakitae/threshold_tools.py
ckingdon95: I think that this tool doesn't make much sense at a state level, because it uses an area average. I think it makes more sense for local analysis. But maybe I should just default to the first alphabetical county?

climakitae/threshold_tools.py
ckingdon95: do you know how to add title and subtitle? I need both, which is why I left it as is for now (with the text outside the plot, I agree this is not ideal), but I couldn't find a great explanation for how to add subtitles.

climakitae/threshold_tools.py
```diff
@@ -884,3 +889,523 @@ def return_period(ams):
...
+    """
+    Plots each simulation as a different color line.
+    Drop down option to select different scenario.
+    Currently can only plot for one location, so is expecting input to already be subsetted or an area average.
+    """
+    plot_obj = exceedance_count.hvplot.line(
+        x="time", 
+        widget_location="bottom", 
+        by="simulation", 
+        groupby=["scenario"],
```
nicolejkeeney, line 1099: I don't think I like this scenario dropdown, because I would expect that there's only one scenario option always-- right? Is that a correct assumption? If so, I would make that dropdown disappear, and add the scenario information as a subtitle to the plot. If that's not a correct assumption, and you'd expect that with different data selections (or in future iterations of this panel), there would be more than one scenario option, definitely leave the dropdown in.  

only one dropdown option

climakitae/explore2.py
ckingdon95: yay, thanks Grace!

climakitae/threshold_tools.py
nicolejkeeney: I only know how to add a title unfortunately. Just put it into where you have ```title = ""``` (although I assume you knew this already, sorry if that was mansplaining!). Victoria, Grace, and I were chatting [on slack](https://cal-adaptdp.slack.com/archives/C021EAYF0MC/p1664301987204829) about this issue as well though; the issue being, how to we get all the appropriate data into these plots, without having a super long and messy title? Hopefully we can all get together post- code freeze and discuss a good method for this, since it's something we want in all the explore panels.

climakitae/threshold_tools.py
nicolejkeeney: Yep, that works!

nicolejkeeney: It might make more sense to include `_exceedance_visualize` as part of `ThresholdDataParams`. That way, you could have the value shown (i.e. Value (units: K)) update in a function with `param.depends("units",watch=True)` so it dynamically updates.

nicolejkeeney (changes requested):

nicolejkeeney (changes requested):
All of your code looks great. I just set it to "request changes" since we need to wait on getting lmoments3 in before you merge. Brian [has a PR up now](https://github.com/cal-adapt/climakitae/pull/88) and I think he is working on fixing that.

nicolejkeeney (approved):

PR #61, opened 2022-08-10
Reorganize tests

nicolejkeeney:
This PR just organizes the test files and puts them into folders matching the modules that each test is written for. I was thinking it would be nice to organize our tests this way going forward, to avoid having a massive tests/ folder with tons of files in it. This way, we can know which module each test file corresponds to.


vicford (approved):
Looks good to me - we love organization, especially if test file(s) will become more extensive with future dev

ckingdon95 (approved):

PR #63, opened 2022-08-23
Use S3 based parquet files to run subsetting plus code streamlining and cleanup

elehmer:
This PR tackles replacing the current boundary data with parquet files based within an intake catalog on CA:AE S3 bucket. It modifies `selectors.py` and `data_loaders.py`. In selectors in replaces the Boundary data refs with the S3 parquet files and redos how the states dictionary is created removing regionmask and numpy as dependencies. The watersheds dataset also switches to using the index instead of the objectid as a primary key. This is so ALL subset area types can be handled in a similar way and the code compacted into a function call and logic streamlined. The same treatment of code compacting and streamlining is done in `data_loaders` and logic is same as in `selectors`.

### DIFF
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -140,6 +140,9 @@ def _read_from_catalog(selections, location, cat):
 
     assert not selections.scenario == [], "Please select as least one scenario."
 
+    def set_subarea(boundary_dataset):
+        return boundary_dataset[boundary_dataset.index == shape_index].iloc[0].geometry
+
     if location.area_subset == "lat/lon":
         geom = _get_as_shapely(location)
         if not geom.is_valid:
@@ -154,12 +157,12 @@ def _read_from_catalog(selections, location, cat):
         shape_index = int(
             location._geography_choose[location.area_subset][location.cached_area]
         )
-        if location.area_subset == "CA watersheds":
-            shape = location._geographies._ca_watersheds
-            shape = shape[shape["OBJECTID"] == shape_index].iloc[0].geometry
+        if location.area_subset == "states":
+            shape = set_subarea(location._geographies._us_states)
         elif location.area_subset == "CA counties":
-            shape = location._geographies._ca_counties
-            shape = shape[shape.index == shape_index].iloc[0].geometry
+            shape = set_subarea(location._geographies._ca_counties)
+        elif location.area_subset == "CA watersheds":
+            shape = set_subarea(location._geographies._ca_watersheds)
         ds_region = regionmask.Regions(
             [shape], abbrevs=["geographic area"], name="area mask"
         )
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -157,10 +157,6 @@ def _read_from_catalog(selections, location, cat):
         if location.area_subset == "CA watersheds":
             shape = location._geographies._ca_watersheds
             shape = shape[shape["OBJECTID"] == shape_index].iloc[0].geometry
-            wgs84 = pyproj.CRS('EPSG:4326')
-            psdo_merc = pyproj.CRS('EPSG:3857')
-            project = pyproj.Transformer.from_crs(psdo_merc, wgs84, always_xy=True).transform
-            shape = transform(project, shape)
         elif location.area_subset == "CA counties":
             shape = location._geographies._ca_counties
             shape = shape[shape.index == shape_index].iloc[0].geometry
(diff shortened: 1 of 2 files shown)


climakitae/data_loaders.py
```diff
@@ -140,30 +140,24 @@ def _read_from_catalog(selections, location, cat):
...
-    elif location.area_subset == "states":
-        shape_index = int(
-            location._geography_choose[location.area_subset][location.cached_area]
-        )
-        ds_region = location._geographies._us_states[[shape_index]]
     elif location.area_subset != "none":
         shape_index = int(
             location._geography_choose[location.area_subset][location.cached_area]
         )
-        if location.area_subset == "CA watersheds":
```
elehmer, line 157: Reprojection no longer required. All boundary data are in EPSG:4326

climakitae/selectors.py
```diff
@@ -6,8 +6,6 @@
 import matplotlib.ticker as ticker
 import cartopy.crs as ccrs
 import cartopy.feature as cfeature
-import regionmask
```
elehmer, line 9: regionmask and numpy no longer needed in selectors

climakitae/selectors.py
```diff
@@ -56,62 +54,53 @@
...
-        self._ca_counties = gpd.read_file(
-            "https://tigerweb.geo.census.gov/arcgis/rest/services/TIGERweb/State_County/MapServer/1/query?where=STATE='06'&f=geojson"
-        )
-        self._ca_counties = self._ca_counties.sort_values("NAME")
-
-        self._ca_watersheds_file = "https://gis.data.cnra.ca.gov/datasets/02ff4971b8084ca593309036fb72289c_0.zip?outSR=%7B%22latestWkid%22%3A3857%2C%22wkid%22%3A102100%7D"
-        self._ca_watersheds = gpd.read_file(self._ca_watersheds_file)
-        self._ca_watersheds = self._ca_watersheds.sort_values("Name")
+        self._cat = intake.open_catalog("https://cadcat.s3.amazonaws.com/parquet/catalog.yaml")
+        self._us_states = self._cat.states.read()
```
elehmer, line 58: states isn't sorted here because it needs a custom sorter (not alpha)

climakitae/selectors.py
```diff
@@ -56,62 +54,53 @@
...
+            "UT",
+            "MT",
+            "ID",
+            "AZ",
+            "CO",
+            "NM",
+            "WY"
+        ]
+        _us_states_subset = self._us_states.query("abbrevs in @_states_subset_list")[["abbrevs"]]
+        _us_states_subset["abbrevs"] = pd.Categorical(_us_states_subset["abbrevs"], categories=_states_subset_list)
```
elehmer, line 80: This is a cool feature of pandas, a Categorical can be sorted using a list to create a custom sort order!

climakitae/selectors.py
```diff
@@ -262,40 +251,29 @@ def view(self):
             color="k",
         )
         mpl_pane = pn.pane.Matplotlib(fig0, dpi=144)
+
+        def plot_subarea(boundary_dataset, extent):
```
elehmer, line 255: Compacted repeated code into function...

climakitae/selectors.py
```diff
@@ -56,62 +54,53 @@
...
             self._ca_counties.index, index=self._ca_counties["NAME"]
         ).to_dict()
 
     def get_ca_watersheds(self):
         """
-        Returns a lookup dictionary for CA watersheds that references the shapefile.
+        Returns a lookup dictionary for CA watersheds that references the geoparquet file.
         """
         return pd.Series(
-            self._ca_watersheds["OBJECTID"].values, index=self._ca_watersheds["Name"]
```
elehmer, line 109: No longer using OBJECTID but using the internal index to conform with how the other subareas are retrieved.

climakitae/selectors.py
```diff
@@ -262,40 +251,29 @@ def view(self):
             color="k",
         )
         mpl_pane = pn.pane.Matplotlib(fig0, dpi=144)
+
+        def plot_subarea(boundary_dataset, extent):
+            ax.set_extent(extent, crs=xy)
+            subarea = boundary_dataset[boundary_dataset.index == shape_index]
+            df_ae = subarea.to_crs(crs_proj4)
+            df_ae.plot(ax=ax, color="b", zorder=2)
```
elehmer, line 259: This code plots `states` with a solid blue symbology like the other 2 datasets instead of a blue outline. Let me know if you want to keep these different.

nicolejkeeney (changes requested):
I believe this will cause the same issues with explore.py as I noted in my comment for PR #64

elehmer:
Going for a merge so that Nicole can adapt to these changes...

PR #64, opened 2022-08-24
Replace regionmask with rioxarray.clip for masking of data

elehmer:
This PR completely removes all dependency on `regionmask` by replacing the remaining calls with `rioxarray.clip`. This particular code is dependent on the changes in #63. `_read_from_catalog` now passes a geometry collection instead of a Region object to `_open_and_concat` where the mask is done with one command! It appears to work just as before and I have not seen much difference in performance, which is unfortunate but at least we have one less non-pangeo dependency.

It seems to fail gracefully without the "Insufficient Gridcells" assertion. When I query for the 3km data with Washington state which is non-intersecting I get: `NoDataInBounds: No data found in bounds. Data variable: T2` from `rioxarray`. We could enhance this change by adding some custom error handling for this.

### DIFF
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -140,6 +140,9 @@ def _read_from_catalog(selections, location, cat):
 
     assert not selections.scenario == [], "Please select as least one scenario."
 
+    def set_subarea(boundary_dataset):
+        return boundary_dataset[boundary_dataset.index == shape_index].iloc[0].geometry
+
     if location.area_subset == "lat/lon":
         geom = _get_as_shapely(location)
         if not geom.is_valid:
@@ -154,12 +157,12 @@ def _read_from_catalog(selections, location, cat):
         shape_index = int(
             location._geography_choose[location.area_subset][location.cached_area]
         )
-        if location.area_subset == "CA watersheds":
-            shape = location._geographies._ca_watersheds
-            shape = shape[shape["OBJECTID"] == shape_index].iloc[0].geometry
+        if location.area_subset == "states":
+            shape = set_subarea(location._geographies._us_states)
         elif location.area_subset == "CA counties":
-            shape = location._geographies._ca_counties
-            shape = shape[shape.index == shape_index].iloc[0].geometry
+            shape = set_subarea(location._geographies._ca_counties)
+        elif location.area_subset == "CA watersheds":
+            shape = set_subarea(location._geographies._ca_watersheds)
         ds_region = regionmask.Regions(
             [shape], abbrevs=["geographic area"], name="area mask"
         )
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -3,7 +3,7 @@ import cartopy.crs as ccrs
 import pyproj
 from shapely.geometry import box
 from shapely.ops import transform
-import regionmask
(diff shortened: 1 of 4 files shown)


climakitae/data_loaders.py
```diff
@@ -140,33 +127,25 @@ def _read_from_catalog(selections, location, cat):
...
         geom = _get_as_shapely(location)
         if not geom.is_valid:
             raise ValueError("Please go back to 'select' and choose a valid lat/lon range.")
-        ds_region = regionmask.Regions([geom], abbrevs=["lat/lon box"], name="box mask")
-    elif location.area_subset == "states":
-        shape_index = int(
-            location._geography_choose[location.area_subset][location.cached_area]
-        )
-        ds_region = location._geographies._us_states[[shape_index]]
+        ds_region = [geom]
```
elehmer, line 137: This is the change from Region to geometry

climakitae/data_loaders.py
```diff
@@ -140,33 +127,25 @@ def _read_from_catalog(selections, location, cat):
...
         elif location.area_subset == "CA counties":
-            shape = location._geographies._ca_counties
-            shape = shape[shape.index == shape_index].iloc[0].geometry
-        ds_region = regionmask.Regions(
-            [shape], abbrevs=["geographic area"], name="area mask"
-        )
+            shape = set_subarea(location._geographies._ca_counties)
+        elif location.area_subset == "CA watersheds":
+            shape = set_subarea(location._geographies._ca_watersheds)
+        ds_region = [shape]
```
elehmer, line 148: Change from Region to geometry

climakitae/data_loaders.py
```diff
@@ -80,21 +80,8 @@ def _open_and_concat(file_list, selections, cat, ds_region):
...
-
-            data = data.sel(x=slice(np.nanmin(output[:,0]), np.nanmax(output[:,0])),
-                y=slice(np.nanmin(output[:,1]), np.nanmax(output[:,1])))
-
-            mask = ds_region.mask(data.lon, data.lat, wrap_lon=False)
-            assert (
-                False in mask.isnull()
-            ), "Insufficient gridcells are contained within the bounds."
-            data = data.where(np.isnan(mask) == False)
+            data = data.rio.clip(geometries=ds_region, crs=4326, drop=True)
```
elehmer, line 84: This seems to be the only needed operation and it runs as fast as all this other code.

climakitae/data_loaders.py
```diff
@@ -3,7 +3,7 @@
 import pyproj
 from shapely.geometry import box
 from shapely.ops import transform
-import regionmask
```
elehmer, line 6: Removal of final regionmask dependency!

nicolejkeeney (changes requested):
This PR will cause a conflict with the current version of explore.py, which was written using the regionmask library. The function _get_postage_data from explore.py modified some code from selectors.py to pull data from AWS and then perform spatial subsetting. I'm not sure the best way to go about resolving this conflict, since the explore.py that exists in the PR is the older version from several weeks ago before we built in the warming levels panel.

nicolejkeeney (approved):

elehmer:
Merging so that Nicole can adapt to these changes...

PR #66, opened 2022-09-06
app.explore(): Interactive plotting & improvements to GMT context plot

nicolejkeeney:
I've made some small(ish) changes since WG 4, listed below: 

1. Postage stamp plots are now interactive. They are made with hvplot.image(). Everything is a tiny bit slower than with the static plots, but I think it's worth it, because these maps are rad. 
interactive maps

3. The GMT context plot includes all the SSPs now (1-1.9 and 1-2.6 as well). It includes options to show all SSPs on the same plot, or to just show the selected SSP. I've also added labels to the confidence interval lines that improve the readability of the plot (at least, I think so!). 

https://user-images.githubusercontent.com/66140951/188938967-e47ebac6-936a-4133-a2a9-f169edb9b793.mov

### DIFF
--- a/climakitae/explore.py
+++ b/climakitae/explore.py
@@ -65,7 +65,7 @@ hist_data = pd.read_csv(hist, index_col='Year')
 
 
 
-def read_cached_tmy_df(cached_tmy_files, variable, warmlevel, cached_area):
+def _read_cached_tmy_df(cached_tmy_files, variable, warmlevel, cached_area):
     """Read in cached tmy file corresponding to a given variable, warmlevel, and cached area.
     Returns a dataframe"""
 
@@ -93,6 +93,11 @@ def read_cached_tmy_df(cached_tmy_files, variable, warmlevel, cached_area):
 
     # Read in file as pandas dataframe
     df = pd.read_csv(tmy[0], index_col=0)
+    
+    # Name columns and index 
+    df.columns.name = "Hour of Day"
+    df.index.name = "Day of Year"
+    
     return df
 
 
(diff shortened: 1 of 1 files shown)


climakitae/explore.py
```diff
@@ -188,12 +193,30 @@ def get_anomaly_data(data, warmlevel=3.0):
...
+    return _plot
+
+
 class WarmingLevels(param.Parameterized):
 
     ## ---------- Reset certain DataSelector and LocSelectorArea options ----------
     def __init__(self, *args, **params):
         super().__init__(*args, **params)
-
+
```
elehmer, line 219: Unnecessary white space...

vicford:
Conceptually - I have some mixed feelings on the way that the confidence interval is now displayed because shading is not included. I think many of our stakeholders may be confused as to the solid lines now without the shading present. 

I do like the mean crossing line now explicitly stating the year of exceedance, as well as the new interactive postage stamp plots.

nicolejkeeney:
> 

@vicford Do you think we ought to either (a) remove the confidence interval from the label or (b) convert the lines to shading, as is typical for a CI on a plot? 

Also I definitely did not mean to remove the shading around the SSP line, I will add that back in now. Thanks for pointing that out, I hadn't even noticed!

vicford:
> we ought to either (a) remove the confidence interval from the label or (b) convert the lines to shading, as is typical for a CI on a plot

I think including the 90% interval is useful for our purposes, and I think shading accomplishes that as long as we are clear in our description of what it is. This might be a larger conversation about whether the solid lines to illustrate that range (as well as providing the "x years" as the range) is intentionally useful though or just confusing. I'm leaning towards shading but no solid lines?

nicolejkeeney:
> 

Awesome. I think I've resolved this issue :D

nicolejkeeney:
I believe Beth, Eric, and Victoria have all tested this so I am considering it "done" and moving onto the next PR. Merge pending, following resolution of inconsistencies w/ Eric's PR and waiting on monthly data in AWS.

vicford (approved):

PR #68, opened 2022-09-07
app.explore(): Documentation improvements & Added 4 deg warming option

nicolejkeeney:
This PR has expanded documentation in the GMT context plot and both postage stamp tabs, as well as a new panel to show info about the anomaly calculations. It also adds a "tip" box in the first postage stamp tab telling users how to use the plotting toolbar. Lastly, I've added back in the 4 deg warming option! I took some creative liberties here, trying to develop features requested by stakeholder, and would love some feedback on what people think. 

### Tip box in first postage stamp panel: 
tip box

### New panel "Anomaly computation details": 
30-yr-window

### DIFF
--- a/climakitae/explore.py
+++ b/climakitae/explore.py
@@ -24,8 +24,6 @@ import matplotlib.colors as mcolors
 import logging
 logging.getLogger("param").setLevel(logging.CRITICAL)
 
-#test
-
 # Import package data
 ssp119 = pkg_resources.resource_filename('climakitae', 'data/tas_global_SSP1_1_9.csv')
 ssp126 = pkg_resources.resource_filename('climakitae', 'data/tas_global_SSP1_2_6.csv')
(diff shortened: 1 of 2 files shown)


ckingdon95 (approved):

vicford (approved):
I'd also recommend putting the language from the 'maps of individual panels' regarding that the anomalies are calculated based on the 30 years centered on the warming level crossing and the 1981-2010 baseline on the model statistics mean/median/min/max as well just for explicit statements. 

Love the tip pop up too.

nicolejkeeney:
> I'd also recommend putting the language from the 'maps of individual panels' regarding that the anomalies are calculated based on the 30 years centered on the warming level crossing and the 1981-2010 baseline on the model statistics mean/median/min/max as well just for explicit statements.
> 
> Love the tip pop up too.

Done, thanks for your input!

PR #69, opened 2022-09-09
New feature: app.view()

nicolejkeeney:
### Basic info about this PR 
This new feature is designed to allow the user a quick glance at their selected data. It generates a map for data with spatial coordinates, and a line plot for data with only a time coordinate. 

At the moment it only works for DataArrays, but we could later add functionality to work with Datasets if we envision that as a part of our workflow. 

**Examples (screenshots) of the new feature is at the bottom!**

### Additional Information
As an example, the steps to call app.view() from the getting_started notebook would be: 
`app.select()`
`data = app.retrieve()`
`data = data.compute()`
`app.view(data) `

You don't _need_ to call `data.compute()` beforehand, but `app.view()` is quite slow without it. Thus, I've set up the code to raise a warning if you try to pass it a zarr store. It will still generate a map/plot, just warn the user that it may take a while. 

This PR includes a new (and beefy) function that reprojects 1D x,y data to 1D lat, lon data. This is used to generate the map if app.view is fed data with spatial coordinates. I had to do this instead of using the native lat,lon coords because the native coords are curvilinear, and the plotting library hvplot.image() requires 1D coordinates. 

For some reason, the reprojection occasionally fails for certain data selections. I've set `app.view` to plot the data anyways on the native x,y grid, and raise a warning that the projection didn't occur. 

### Function Arguments 
`app.view` allows for several user arguments to modify the plot/map shown. These arguments are: 
1. height & width of plot:  modify dimensions of map/plot
2. lat/lon: option to show the data in lat/lon or x/y (default to lat/lon) 
3. cmap: modify the colorbar (applies to maps only)

The colormap defaults to inferno_r (we could easily change this) 

### Examples
(1) Spatial data: 45km, surface pressure, entire grid

45km surface pressure

(2) Temporal data: Area-averaged radiation data 

line_plot

(3) Spatial data: 45km, 2m air temp with user-defined width, height, cmap, & projection (x,y)
ca_map

### DIFF
--- a/climakitae/core.py
+++ b/climakitae/core.py
@@ -48,7 +48,7 @@ class Application(object):
         return _read_from_catalog(self.selections, self.location, self._cat)
     
     # === View =====================================
-    def view(self, data, lat_lon=True, width=None, height=None, cmap="viridis"): 
+    def view(self, data, lat_lon=True, width=None, height=None, cmap="inferno_r"): 
         """Create a generic visualization of the data
     
         Args: 
--- a/climakitae/core.py
+++ b/climakitae/core.py
@@ -48,14 +48,25 @@ class Application(object):
         return _read_from_catalog(self.selections, self.location, self._cat)
     
     # === View =====================================
-    def view(self, data): 
-        """Create generic visualization of data"""
-        return _visualize(data)
+    def view(self, data, lat_lon=True, width=None, height=None): 
+        """Create a generic visualization of the data
+    
+        Args: 
+            data (xr.DataArray)
+            lat_lon (boolean): reproject to lat/lon coords? (default to True) 
+            width (int): width of plot (default to hvplot.image default) 
+            height (int): hight of plot (default to hvplot.image default) 
+        
+        Returns: 
+            hvplot.image()
+
+        """
+        return _visualize(data, lat_lon=lat_lon, width=width, height=height)
     
     # === Explore ===================================
     def explore(self): 
         """
-        Calls a method to display a plot of the data 
+        Display warming levels panel 
         """
         return _display_warming_levels(self.selections, self.location, self._cat)
     
--- a/climakitae/core.py
+++ b/climakitae/core.py
@@ -48,7 +48,7 @@ class Application(object):
         return _read_from_catalog(self.selections, self.location, self._cat)
     
     # === View =====================================
-    def view(self, data, lat_lon=True, width=None, height=None): 
+    def view(self, data, lat_lon=True, width=None, height=None, cmap="viridis"): 
         """Create a generic visualization of the data
     
         Args: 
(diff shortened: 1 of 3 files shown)


ckingdon95 (approved):
Love this, looks awesome! I appreciate the defaulting to map or timeseries depending on whether it's spatial or not. I could see options later on where if a user has x-y data they still might want to see some kind of timeseries (either area average or selected cell). But I think that's something to chat about later!

nicolejkeeney:
> Love this, looks awesome! I appreciate the defaulting to map or timeseries depending on whether it's spatial or not. I could see options later on where if a user has x-y data they still might want to see some kind of timeseries (either area average or selected cell). But I think that's something to chat about later!

Good idea regarding timeseries option for spatial data. I could see this being something as simple as "timeseries=True" or "timeseries=False" as function arguments.

vicford (approved):
This looks great! I primarily tested using monthly temp in degF for 2008-2015 at 45/9/3km for CA, Fresno County, and Clearcreek watershed and overall didn't run into any issues, just varying compute times. Some testing of reprojection observations:
- monthly relhumid 2002-2015 9 and 3km for Alameda county both took 2-5 minutes and had reprojection errors
- monthly 3km for california took 10.5 minutes, but did run successfully (temp 2008-2015)
- monthly 9km for california (temp 1985-2010) took 16 minutes, but also ran successfully
- On relative humidity: CRS not found. Please set the CRS with 'rio.write_crs()'. Data variable: REL_HUMIDITY (but could be related to most recent PR pushes)

I also really like that there is default plotting of map vs. time series going on too. I have some thoughts about the colormapping, but that's a separate thing. Nicely done!

nicolejkeeney:
> This looks great! I primarily tested using monthly temp in degF for 2008-2015 at 45/9/3km for CA, Fresno County, and Clearcreek watershed and overall didn't run into any issues, just varying compute times. Some testing of reprojection observations:
> 
> * monthly relhumid 2002-2015 9 and 3km for Alameda county both took 2-5 minutes and had reprojection errors -- this might have been before the most recent PR merges though
> * monthly 3km for california took 10.5 minutes, but did run successfully (temp 2008-2015)
> * monthly 9km for california (temp 1985-2010) took 16 minutes, but also ran successfully
> * On relative humidity: CRS not found. Please set the CRS with 'rio.write_crs()'. Data variable: REL_HUMIDITY (but could be related to most recent PR pushes)
> 
> I also really like that there is default plotting of map vs. time series going on too. I have some thoughts about the colormapping, but that's a separate thing. Nicely done!

I've noticed this issue with Relative Humidity with another workflow. I will investigate further. Going to PM you about the colormap stuff, as I'm not sure what is best for the default cmap.

nicolejkeeney:
Relative humidity will be fixed in a separate PR

PR #70, opened 2022-09-12
fixed bug that attempted to slice coordinate instead of dimension

bethmints:
fixing this bug exposed another one associated with exporting metadata attributes to a text file for csv/geotiff. This will be a part of a separate pull request (or potentially a larger conversation re: metadata).

### DIFF
--- a/climakitae/data_export.py
+++ b/climakitae/data_export.py
@@ -62,28 +62,42 @@ def _export_to_user(user_export_format,data_to_export,
     # we have to ensure that each non-NetCDF has only one simulation
     # and scenario per file. 
     if ("NetCDF" not in req_format):
-        if ('simulation' in data_to_export.coords): # if simulation coord exists
-            if ('simulation' not in kwargs): # and one simulation is not supplied by user
-                if (np.size(data_to_export.coords['simulation'].values) > 1): # and there is > 1
+        
+        if ('simulation' in data_to_export.dims): # if simulation dim exists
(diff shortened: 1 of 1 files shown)


nicolejkeeney:
The export data function isn't acting as I would expect regarding data coords/dims. If I feed the export_dataset my simulation of choice as a kwarg, I would think that it would "slice" the data and only output a netcdf containing my simulation of choice. But, no slicing occurs and the entire dataset is downloaded. 
 
kwargs

nicolejkeeney:
The function throws an error when you try to download the data using a filename that already exists in your directory. It would be nice to have this throw a custom error instead of the default scary red python error message. The actual error message that is thrown is `PermissionError: [Errno 13] Permission denied: b'/home/jovyan/cae-notebooks/cesm.nc'`, which occured because I already had a file named "cesm.nc" in my directory. Perhaps just a custom error, warning, or print message is better, i.e. `"A file named "cesm.nc" already exists in your directory. Please choose a different filename."`

error
error_2

nicolejkeeney (approved):
This all looks good to me. I left additional some comments about related issues, but you could address those in a separate PR.

bethmints:
Ah, this is a good catch.

vicford (approved):
Specifically was able to raise the desired errors, but going to second @nicolejkeeney comment about unique filenames, that would be useful in a future iteration!

PR #72, opened 2022-09-13
Add explore class to Application

ckingdon95:
This creates an additional class so that we can call various explore methods with:
```python
app.explore2.TMY()
app.explore2.thresholds()
```
I have named the attribute `explore2` for now until we decide how to handle or rename the existing app.explore() method.


nicolejkeeney (approved):
This looks good to me. I've tested in the hub and it performs as expected, and doesn't break the existing app.explore() panel (yay!)

vicford (approved):
Looks great to me! Only thought is for consistency's purposes, the TMY option could also be lowercase (so app.explore2.tmy) because all of our other methods/functions are lowercase too

PR #74, opened 2022-09-14
Branded colormap options

vicford:
Hi all, 

I've created three Cal-Adapt Analytics Engine branded colormaps for us to use when displaying our data. Options: 
1. **ae_orange** : broadly represents increasing positive values, but also temperature-related/radiation variables, my vote for our default colormap
2. **ae_blue** : broadly represents negative values, but can also be for moisture-related variables
3. **ae_div** : diverging colormap option that utilizes both ae_orange and ae_blue colors; can be used for wind-related variables, pressure (while pressure doesn't diverge, it is commonly. plotted as such for high/low pressure), as well as any anomaly display

Because visualizations touch on many files, I'll need guidance/assistance on 'behind-the-scenes' implementation in climakitae (thinking specifically the explore/explore2.py, visualize.py, and threshold_tools.py scripts, could be missing others). But for now, to implement the new colormaps, you can add the following code (example using ae_orange): 

```
import numpy as np
import matplotlib.colors as mcolors
ae_orange = pkg_resources.resource_filename('climakitae', 'data/cmaps/ae_orange.txt')
ae_orange_data = np.loadtxt(ae_orange)
ae_orange_cm = mcolors.LinearSegmentedColormap.from_list('ae_orange_cmap', ae_orange_data, N=256)
```

**Examples of what this looks like from explore_warming notebook below:**

Screen Shot 2022-09-14 at 1 38 05 PM
Screen Shot 2022-09-14 at 12 05 30 PM
Screen Shot 2022-09-14 at 12 22 51 PM


elehmer (approved):
These look great!

elehmer:
Maybe add a CR+LF to each of these files to comply with our style guide?

nicolejkeeney (approved):
All looks great to me!

PR #75, opened 2022-09-14
Add back in grid_mapping attribute for relative humidity

nicolejkeeney:
**Information:** The grid_mapping attribute is lost for some reason by the metpy function that computes relative humidity in our derive_variables module. This is causing issues with reprojecting the data, as grid_mapping is used by rioxarray to determine information about the data's CRS. In this PR, I just add back in that attribute. 

**To test:** Try just selecting relative humidity from app.select() and make sure the attribute "grid_mapping" appears.

### DIFF
--- a/climakitae/derive_variables.py
+++ b/climakitae/derive_variables.py
@@ -46,8 +46,8 @@ def _compute_relative_humidity(pressure, temperature, mixing_ratio, variable_nam
     # For some reason, the grid_mapping attr is lost by the metpy function
     # We want to add it back in to allow for projecting the data
     for var in [pressure, temperature, mixing_ratio]:
-        if “grid_mapping” in var.attrs:
-            rel_hum.attrs[“grid_mapping”] = var.attrs[“grid_mapping”]
+        if "grid_mapping" in var.attrs:
+            rel_hum.attrs["grid_mapping"] = var.attrs["grid_mapping"]
             
     return rel_hum
 
--- a/climakitae/derive_variables.py
+++ b/climakitae/derive_variables.py
@@ -43,6 +43,12 @@ def _compute_relative_humidity(pressure, temperature, mixing_ratio, variable_nam
     rel_hum.name = variable_name
     rel_hum.attrs["description"] = "Relative humidity"
 
+    # For some reason, the grid_mapping attr is lost by the metpy function
+    # We want to add it back in to allow for projecting the data
+    for var in [pressure, temperature, mixing_ratio]:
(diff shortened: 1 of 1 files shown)


vicford (approved):
grid_mapping = Lambert_Conformal 

back in, whoop

elehmer:
Hmm, yeah looking at the `metpy` docs  it looks like a dimension-less quantity (relative humidity) is returned and so they didn't think the attrs would be needed? I don't see any handles inside the library to change any of this behavior. Are all attrs lost or just the `grid_mapping` one? Anyway, this might be the most expedient way to deal with this issue.

nicolejkeeney:
> 

Weirdly, it seems like only the grid_mapping attr is lost.

nicolejkeeney:
> Hmm, yeah looking at the `metpy` docs it looks like a dimension-less quantity (relative humidity) is returned and so they didn't think the attrs would be needed? I don't see any handles inside the library to change any of this behavior. Are all attrs lost or just the `grid_mapping` one? Anyway, this might be the most expedient way to deal with this issue.

I just checked and the variables that go into that metpy function all have the grid_mapping function, yet the output does not? It maintains the other attributes though. I don't understand why.

elehmer:
It must happen at the `calc` stage as the `dequantify` does a `copy` which is a deep copy by default so the attrs should come with. Have you tried integrating `parse_cf` into the code somewhere?

https://unidata.github.io/MetPy/latest/api/generated/metpy.xarray.html?highlight=dequantify#metpy.xarray.MetPyDataArrayAccessor.dequantify

It's on this page. Also maybe the `assign_crs` method might help?

nicolejkeeney:
> It must happen at the `calc` stage as the `dequantify` does a `copy` which is a deep copy by default so the attrs should come with. Have you tried integrating `parse_cf` into the code somewhere?
> 
> https://unidata.github.io/MetPy/latest/api/generated/metpy.xarray.html?highlight=dequantify#metpy.xarray.MetPyDataArrayAccessor.dequantify
> 
> It's on this page. Also maybe the `assign_crs` method might help?

What do you mean? Is this an argument to metpy.calc.relative_humidity_from_mixing_ratio? 

I shouldn't need to assign a crs. I really just need grid_mapping as an attribute set to Lambert_Conformal. You're right that something weird is going on under the hood where the function removes that attribute.

bethmints (approved):
the attribute grid_mapping is in the data array obtained by app.select()

PR #76, opened 2022-09-14
Simplified export

bethmints:
### About this pull request 

- The export() function now assumes that users have subsetted and sliced their data array already (ie, it accepts no arguments to do so).
- It will only accept multi-variate datasets for exporting to netcdf (not to csv or geotiff).
- It is no longer writing out a metadata txt file for csv or geotiff.
- Output netcdfs still violate CF conventions by setting _FillValue to NaN for coordinates - this is xarray default behavior.
- Raises an error if a file with the same name exists in the workspace.

### To do:

- [ ] Attach metadata to csv and geotiff
- [ ] Figure out that annoying coordinate issue

### DIFF
--- a/climakitae/data_export.py
+++ b/climakitae/data_export.py
@@ -8,6 +8,34 @@ import numpy as np
 xr.set_options(keep_attrs=True)
 import os
 
+def export_to_netcdf(data_to_export,save_name,**kwargs):
+    print("Alright, exporting specified data to NetCDF. This might take a while - "+
+         "hang tight!")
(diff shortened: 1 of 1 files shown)


vicford (approved):
This works for me - I also appreciate the warning for duplicate file names!

elehmer (approved):
This looks good. You seem to have trapped for the most likely issues I can think of.

Including the commented out code isn't necessary as the git history will have the old code. You can cut out the commented code and put it in an uncontrolled file file if you think you'll need to consult it.

Thanks!

PR #77, opened 2022-09-15
Mapping improvements

nicolejkeeney:
This PR incorporates the AE branded colormaps into app.explore and app.view. 

Each variable has been assigned a default colormap, which was added as an extra column in variable_descriptions.csv. A new utils function aids in reading in the cmap text files.


vicford (approved):
Love the colormaps in action!

elehmer (approved):
Now we cooking with gas!!! LOL

Looks great!!

elehmer:
I'm wondering if there is a way to only store the RGB colormap and convert to HEX on the fly? That way we only have to maintain one set and they won't diverge over time by accident. This could be an improvement we make later after launch...

Here is one possible way I think:
https://stackoverflow.com/questions/33596491/extract-matplotlib-colormap-in-hex-format

nicolejkeeney:
> I'm wondering if there is a way to only store the RGB colormap and convert to HEX on the fly? That way we only have to maintain one set and they won't diverge over time by accident. This could be an improvement we make later after launch...
> 
> Here is one possible way I think: https://stackoverflow.com/questions/33596491/extract-matplotlib-colormap-in-hex-format

Agreed that this would be a better method. Will check this out. Thanks!

elehmer:
This method looks even better: https://www.tutorialspoint.com/extract-matplotlib-colormap-in-hex-format

elehmer:
> > I'm wondering if there is a way to only store the RGB colormap and convert to HEX on the fly? That way we only have to maintain one set and they won't diverge over time by accident. This could be an improvement we make later after launch...
> > Here is one possible way I think: https://stackoverflow.com/questions/33596491/extract-matplotlib-colormap-in-hex-format
> 
> Agreed that this would be a better method. Will check this out. Thanks!

This method looks clean and efficient: https://www.tutorialspoint.com/extract-matplotlib-colormap-in-hex-format

nicolejkeeney:
> > > I'm wondering if there is a way to only store the RGB colormap and convert to HEX on the fly? That way we only have to maintain one set and they won't diverge over time by accident. This could be an improvement we make later after launch...
> > > Here is one possible way I think: https://stackoverflow.com/questions/33596491/extract-matplotlib-colormap-in-hex-format
> > 
> > 
> > Agreed that this would be a better method. Will check this out. Thanks!
> 
> This method looks clean and efficient: https://www.tutorialspoint.com/extract-matplotlib-colormap-in-hex-format

Just implemented it! Thanks for this tip

elehmer (approved):
Awesomeness!! Love the minimizing of code!!

vicford:
Now works on the hub for app.explore without kernel crashing! Huzzah!

PR #80, opened 2022-09-16
Update environment.yml file to current HUB package versions

elehmer:
This PR updates the `environment.yml` file used by the testing environment to the package versions currently on the CA:AE HUB server. Also removes `black` and `flake8` as they are not used at this time.

Also updates `requirements.txt` to match `environments.yml`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -23,7 +23,6 @@ install_requires =
     hvplot
     intake
     intake-xarray
-    jinja2
     lmoments3
     matplotlib
     metpy
@@ -33,7 +32,6 @@ install_requires =
     param
     pyproj
     pytest
-    proj
     rioxarray
     s3fs
     scipy
--- a/setup.cfg
+++ b/setup.cfg
@@ -21,7 +21,9 @@ install_requires =
     geoviews
     holoviews
     hvplot
+    intake
     intake-xarray
+    jinja2
     lmoments3
     matplotlib
     metpy
@@ -29,8 +31,10 @@ install_requires =
     pandas
     panel
     param
+    pyproj
     pytest
-    regionmask
+    proj
+    rioxarray
     s3fs
     scipy
     shapely
--- a/environment.yml
+++ b/environment.yml
@@ -3,24 +3,22 @@ name: climakitae-tests
 channels:
   - conda-forge
 dependencies:
-  - python=3.7
-  - black=22.3.0
+  - python=3.9.7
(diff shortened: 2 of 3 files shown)


elehmer:
@bkg Let me know if we should put all of the `environment.yml` packages in `requirements.txt`.

bkg:
@elehmer thanks for sorting this out. Yes, it would be good if those two package lists are in sync.

elehmer:
> @elehmer thanks for sorting this out. Yes, it would be good if those two package lists are in sync.

Done...

elehmer:
@bkg I think I have this all set right but please let me know if not.

setup.cfg
```diff
@@ -21,16 +21,20 @@ install_requires =
     geoviews
     holoviews
     hvplot
+    intake
     intake-xarray
+    jinja2
```
bkg, line 26: I don't believe this is used anywhere.

setup.cfg
```diff
@@ -21,16 +21,20 @@ install_requires =
...
     matplotlib
     metpy
     numpy
     pandas
     panel
     param
+    pyproj
     pytest
-    regionmask
+    proj
```
bkg, line 36: I think `pyproj` is used directly, but not `proj`.

bkg (commented):
The `setup.cfg` should only include direct package dependencies, basically anything `import`ed. So, I don't think that `jinja2` or `proj` are needed there. I believe `jinja2` is present in those other lists to work around incompatible versions when it was implicitly installed, otherwise it wouldn't be needed there either.

elehmer:
> The `setup.cfg` should only include direct package dependencies, basically anything `import`ed. So, I don't think that `jinja2` or `proj` are needed there. I believe `jinja2` is present in those other lists to work around incompatible versions when it was implicitly installed, otherwise it wouldn't be needed there either.

I see. Yes, I agree on both of these packages. I tried to remove `jinja2` from `environment.yml` and `requirements.txt` but panel errored out with import errors.

Ok, I'll remove these packages.

elehmer:
Should we stick with version `0.0.1` in this file or increment it somehow?

bkg:
That only gets bumped when a new package is officially tagged and released.

elehmer:
> That only gets bumped when a new package is officially tagged and released.

Sounds good. Shall we merge this?

bkg:
Yeah, go for it.

PR #85, opened 2022-09-26
Average Meteorological Year via app.explore2.amy()

vicford:
Similar to Cora's thresholds PR #60, this PR implements the average meteorological year (former typical) panel: 

```
import climakitae as ck
app = ck.Application()
app.explore2.amy()
```

Which presents with the following options:
1. Absolute AMY - can select either historical baseline (1981-2010), or future 30-year period based on selected warming level
2. Difference AMY - can select difference between future warming level period minus historical baseline (the difference plots we saw in the WG4 warming level notebook), or _forthcoming_ the severe future period minus historical baseline
2a. Severe AMY will be based on 90th percentile, but should be able to either have full user selection, or a set of pre-determined selection of percentiles, using climtas (https://climtas.readthedocs.io/en/latest/percentile.html)

Screen Shot 2022-09-26 at 4 09 57 PM

### DIFF
--- a/climakitae/explore2.py
+++ b/climakitae/explore2.py
@@ -1,12 +1,76 @@
 # A class for holding the app explore options
 
 import panel as pn
-from .tmy import _display_tmy
+import param
+from .tmy import _tmy_visualize
+
+
+class TMYParams(param.Parameterized):
+    """
+    An object that holds the "Data Options" parameters for the
+    explore.tmy panel.
+    """
+    def __init__(self, *args, **params):
+        super().__init__(*args, **params)
+
+        # Selectors defaults
+        self.selections.append_historical = False
+        self.selections.area_average = True
+        self.selections.resolution = "45 km"
+        self.selections.scenario = ["SSP 3-7.0 -- Business as Usual"]
+        self.selections.time_slice = (1980,2100)
+        self.selections.timescale = "hourly"
+        self.selections.variable = "Air Temperature at 2m"
+
+        # Location defaults
+        self.location.area_subset = 'states'
+        self.location.cached_area = 'CA'
+
+    variable2 = param.ObjectSelector(default="Air Temperature at 2m",
+        objects=["Air Temperature at 2m", "Surface Pressure"]
+    )
+
+    cached_area2 = param.ObjectSelector(default="CA",
+        objects=["CA"]
+    )
+
+    area_subset2 = param.ObjectSelector(
+        default="states",
+        objects=["states", "CA counties"],
+    )
+
+    @param.depends("variable2", watch=True)
+    def _update_variable(self):
+        """Update variable in selections object to reflect variable chosen in panel"""
+        self.selections.variable = self.variable2
+
+    @param.depends("area_subset2", watch=True)
+    def _update_cached_area(self):
+        """
+        Makes the dropdown options for 'cached area' reflect the type of area subsetting
+        selected in 'area_subset' (currently state, county, or watershed boundaries).
+        """
+        if self.area_subset2 == "CA counties":
+            # setting this to the dict works for initializing, but not updating an objects list:
+            self.param["cached_area2"].objects = ["Santa Clara County", "Los Angeles County"]
+            self.cached_area2 = "Santa Clara County"
+        elif self.area_subset2 == "states":
+            self.param["cached_area2"].objects = ["CA"]
+            self.cached_area2 = "CA"
+
+    @param.depends("area_subset2","cached_area2",watch=True)
+    def _updated_location(self):
+        """Update locations object to reflect location chosen in panel"""
+        self.location.area_subset = self.area_subset2
+        self.location.cached_area = self.cached_area2
+
+
+#-----------------------------------------------------------------------
 
 class AppExplore(object):
     """
     A class for holding the following app explore options:
-        app.explore2.TMY()
+        app.explore2.tmy()
         app.explore2.thresholds()
     """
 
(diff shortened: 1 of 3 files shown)


climakitae/tmy.py
```diff
@@ -1,194 +1,346 @@
...
                 )
-        , title="Data Options", collapsible=False, width=460, height=500
+        , title=" How do you want to investigate AMY?", collapsible=False, width=460, height=500
+    )
+
+    mthd_bx = pn.Column(
+        pn.widgets.StaticText(
+            value="An average meteorological year is calculated by selecting the 24 hours for every day that best represent multi-model mean conditions during a 30-year period – 1981-2010 for the historical baseline or centered on the year the warming level is reached. Absolute average meteorolgoical year profiles represent data that is not bias corrected, please exercise caution when analyzing. Warming level future aboslute data is based on 30 year period centered around when multi-model mean exceeds a selected warming level (please see explore_warming notebook for further details). The 'severe' AMY is calculated using the 90th percentile of future warming level data at the selected warming level, and is compared to the historical baseline.",
+            width=400
+        ),
```
nicolejkeeney, line 323: I'd love to see this method box widened a bit to match the width of the panel it's in. Perhaps a width of 500-600 would be more appropriate here.

elehmer: > 

This line is REALLY long lol. Can we do line continuations here?

nicolejkeeney:
I have some design suggestions. I've detailed them in a jupyter notebook, which I'll send to you over slack. 

TLDR; I suggest combining the abs_tmy_options and diff_tmy_options dropdowns into one. I also suggest improving the text descriptions in the panel, and adding a descriptions param to describe the computation method chosen.

Screenshot from the notebook, briefly showing the current workflow and my suggested changes: 

difference

climakitae/tmy.py
```diff
@@ -1,194 +1,346 @@
...
+
+    # TMY options to display
+    tmy_options = param.ObjectSelector(default='Absolute',
+        objects=['Absolute', 'Difference'])
+
+    abs_tmy_options = param.ObjectSelector(default='Historical',
+        objects=['Historical', 'Warming Level Future'])
+
+    diff_tmy_options = param.ObjectSelector(default='Warming Level Future',
+        objects=['Warming Level Future', 'Severe AMY'])
```
nicolejkeeney, line 143: I suggest you remove "Severe AMY" as an option until it is implemented in the code.

nicolejkeeney:
A general comment: On the AE hub with a Gateway Cluster with 8 workers, this takes 7-8 minutes to load/reload for me. Hopefully we can get some help from Naomi and Brian on optimizing the speed of these panels. A reload progress bar or a spinning wheel would be nice (for this and the warming levels panel as well) to more clearly indicate to users that the data is loading.

climakitae/tmy.py
```diff
@@ -1,194 +1,346 @@
...
+            if self.diff_tmy_options == "Warming Level Future":
+                df = df_future - df_hist
+                title = "Average Meteorological Year\nDifference between {} at {}°C and Historical Baseline \n{}".format(self.diff_tmy_options, self.warmlevel, self.cached_area2)
+                clabel = self.variable2 + " (" +self.historical_tmy_data.attrs["units"]+")"
+            else:
+                df = df_future - df_hist # placeholder for now for severe amy
+                title = "Average Meteorological Year\nDifference between {} at 90th percentile and Historical Baseline \n{}".format(self.diff_tmy_options, self.cached_area2)
+                clabel = self.variable2 + " (" +self.historical_tmy_data.attrs["units"]+")"
+        else:
+            title = "Average Meteorological Year\n{}".format(self.cached_area2)
```
nicolejkeeney, line 273: I'm not sure if I like the stacking title set up for the heatmap. The title is a strange width (quite narrow) compared to the width of the plot. I would suggest having a two-line title instead. Perhaps something like this for a few example selections: 

- Average Meteorological Year: Fresno County
  Data: Historical Baseline
- Average Meteorological Year: California 
 Data: 3deg Warming Level Future minus Historical Baseline
- Average Meteorological Year: Santa Clara Watershed 
 Data: 1.5deg Warming Level Future

Although I don't know if I like this that much either.... bleh. I would just go with your gut. Above is just a suggestion.

climakitae/tmy.py
nicolejkeeney: Agree with Eric ^^ just to improve readability of the code

climakitae/tmy.py
nicolejkeeney: @elehmer do you have any opinions about the title of the heatmap? Perhaps if we all put our heads together we can think of a good solution. 3 lines of stacking title feels a bit strange to me, but maybe thats the clearest method.

nicolejkeeney:
This all looks good to me! Nice work, Victoria! I left a few suggestions. You can accept or ignore them as you see fit! :D

climakitae/explore2.py
```diff
@@ -1,21 +1,23 @@
...
+
+#-----------------------------------------------------------------------
 
 class AppExplore(object):
     """
     A class for holding the following app explore options:
-        app.explore2.TMY()
+        app.explore2.amy()
         app.explore2.thresholds()
     """
```
nicolejkeeney, line 15: You and @ckingdon95 both have different versions of this same file. I suggest you communicate between each other about how to best combine these, because we will have merge conflicts with this file.

nicolejkeeney (approved):
This all looks great to me. Nice work, Victoria!

I left a few suggestions, but feel free to ignore them if you don't think they make sense. I also left a comment about the explore2.py file, which conflicts with the explore2.py file in Cora's branch.

climakitae/tmy.py
vicford: Happy to hear any suggestions on this! One of the things I noted from the WG feedback, and my own observations, is that it can be pretty easy to get lost in the weeds about what is being displayed and for what location, which is why I thought putting the location in the title to make it obvious! The nice thing is that when you download the plot as a figure, it does keep that title with location so there's not any confusion about what is **exactly** being displayed.

climakitae/explore2.py
vicford: RE: this, since "TMY" was the old version... i'm wondering if the _order_ in which we merge this might resolve this? As in Cora merges #60 first, and then I merge #85 because mine will update it with the correct "amy" phrasing? Could also just be wilding misunderstanding/misrepresenting how github merges work too

climakitae/tmy.py
vicford: Can do x2!

elehmer: Another possibility is to have an `enumerations.py` where you dump large text blobs and then import them into your various python code

climakitae/explore2.py
nicolejkeeney: @vicford my understanding is that there will be merge conflicts regardless of the who merges first. But, it should be relatively easy to resolve the conflict. Perhaps @ckingdon95 (resident GitHub whiz) can take the lead on resolving the conflict when she merges her branch (i.e. if you are ready to merge yours now before her, I think thats fine).

PR #86, opened 2022-09-26
CSV metadata

bethmints:
### About this pull request

Writes a txt file containing the original xarray attributes when users export to CSV.

To test:
Export a data array to CSV and check out the accompanying .txt


vicford (approved):
I like this one! It's very easy to read through the metadata. Tested with warming levels plots. 

Side note: data outputs meters as metres (while this doesn't matter, it's just curious that it is British spelled metric units)

climakitae/data_export.py
```diff
@@ -34,7 +34,7 @@ def export_to_csv(data_to_export,save_name,**kwargs):
     kwargs: reserved for future use
     '''
     
-    print("WARNING: Exporting to CSV can take a long time,"+
+    print("Exporting to CSV can take a long time,"+
                  " and is not recommended for data with more than 2 dimensions."+
                   " Please note that the data will be compressed via gzip. Even so,"+
          " inherent inefficiencies may result in a file which is too large to save here.")
```
nicolejkeeney, line 40: I recommend removing this statement. I tested with several different data types and the data downloads were all almost instantaneous

climakitae/data_export.py
```diff
@@ -44,8 +44,12 @@ def export_to_csv(data_to_export,save_name,**kwargs):
...
     if (csv_nrows > excel_row_limit):
         print("WARNING: Dataset exceeds Excel limit of "+
-                      str(excel_row_limit)+" rows.")        
-    print("Compressing data... This can take a bit...")
+                      str(excel_row_limit)+" rows.") 
+        
+    
+    metadata_to_file(data_to_export,save_name)
+    print("Converting and compressing data... This can take a bit...")
+
```
nicolejkeeney, line 52: This as well. If the download is fast, it's confusing to notify the user to expect a long wait time

nicolejkeeney (approved):
The metadata looks great to me. One thing I noticed is that the unzipped csv file is missing the .csv file extension. Might be worth looking into for another PR in the future.

bethmints:
thanks for your other comments, and for testing the timing.

re: file extension - that's odd, and I suspect related to the unzipping
itself. I'm used to using gunzip has you specify the extension. users
should be able to add an extension themselves by renaming the file, right?
in any case, thanks for pointing this out. at the very least, maybe I can
put a note in the metadata file letting users know this might be an issue.

bethmints:
> The metadata looks great to me. One thing I noticed is that the unzipped csv file is missing the .csv file extension. Might be worth looking into for another PR in the future.

okay, I've just pushed a fix for this and removed the confusing print statements

elehmer:
This looks good!

PR #88, opened 2022-09-28
Clean up package dependencies

bkg:
Here are some minor packaging and dependency changes that remove a duplicate `lmoments3`, a non-existent `proj`, and separate testing from runtime deps.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -28,5 +28,4 @@ dependencies:
   - xarray=0.19.0
   - pip
   - pip: 
-    - lmoments3
-    - git+https://github.com/OpenHydrology/lmoments3.git@7e19f97c23019ca68cbd526b8bd417c412438f1c#egg=lmoments3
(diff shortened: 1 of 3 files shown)


elehmer (approved):
Ahh, ok thanks for clearing this stuff up. LGTM

nicolejkeeney (changes requested):
I'm still getting an error statement when trying to import the threshold_tools module, where the following import statements are: 
```
import lmoments3 as lm
from lmoments3 import distr as ldistr
from lmoments3 import stats as lstats
```

The full error message: 
import error

bkg:
That's because you're still using the older version of lmoments. You can install the newer version with either `requirements.txt` or `environment.yml`. You can see the newer version of lmoments3 being used in the tests now:

nicolejkeeney (approved):

PR #93, opened 2022-09-30
Remove reprojection from app.explore() to avoid overloading memory

nicolejkeeney:
The reprojection code chunk in app.explore.warming_levels() uses a ton of memory, causing the kernel to fail when you play with `app.explore.warming_levels()` (no bueno!!). 

I've commented out this reprojection code. This seems to solve this issue. The explore.warming_levels panel now uses ~1.6GB of memory, instead of ~3.5GB before removing the reprojection code. 

**To test:** Just call app.explore.warming_levels(). Play around with it. Make sure it doesn't crash your kernel. Try running the entire warming_levels notebook from cae-notebooks, **making sure you replace app.explore() with app.explore.warming_levels()**

### DIFF
--- a/climakitae/warming_levels.py
+++ b/climakitae/warming_levels.py
@@ -100,15 +100,15 @@ def _get_postage_data(area_subset2, cached_area2, variable2, location):
     # Clip data to geometry
     postage_data = postage_data.rio.clip(geometries=ds_region, crs=4326, drop=True)
     
-    # Reproject data to lat/lon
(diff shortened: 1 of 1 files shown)


bethmints:
looking good!

bethmints (approved):

elehmer (approved):
LGTM. We can revisit this at a later date.

vicford (approved):

PR #94, opened 2022-09-30
Switch RH colormap for warming_levels

vicford:
Simple PR to switch the colormap for relative humidity **only*** in warming_levels, because it has both positive and negative values. 

**Caveat**: If/when the warming_levels introduces more variables than just air temperature and relative humidity, a more extensive anomaly colormap setting will need to be in place. All variables have a default colormap in the descriptions csv, however this _should_ be overridden when an anomaly is calculated to display a diverging colormap instead. Midpoint normalizing may need to occur then: https://discourse.holoviz.org/t/diverging-colorbar-around-asymetric-data/3339

To test this, just look at relative humidity in app.explore.warming_levels() and make sure that its the diverging colormap instead of its preset blues.

### DIFF
--- a/climakitae/warming_levels.py
+++ b/climakitae/warming_levels.py
@@ -3,6 +3,7 @@ import hvplot.pandas
 import xarray as xr
 import holoviews as hv
 from holoviews import opts
+import bokeh.palettes
 from matplotlib.figure import Figure
 import numpy as np
 import pandas as pd
@@ -14,7 +15,7 @@ import s3fs
 import pkg_resources
 from .utils import _reproject_data, _read_ae_colormap, _read_var_csv
 
-# Silence warnings 
+# Silence warnings
 import logging
 logging.getLogger("param").setLevel(logging.CRITICAL)
 
@@ -41,7 +42,7 @@ ssp370_data = pd.read_csv(ssp370, index_col='Year')
 ssp585_data = pd.read_csv(ssp585, index_col='Year')
 hist_data = pd.read_csv(hist, index_col='Year')
 
-# Variable descriptions csv with colormap info 
+# Variable descriptions csv with colormap info
 var_descrip = _read_var_csv(var_descrip_pkg, index_col="description")
 
 
@@ -68,12 +69,12 @@ def _get_postage_data(area_subset2, cached_area2, variable2, location):
 
     # Select variable from dataset
     postage_data = pkg_data[variable2].compute()
-    
+
     #================= Modified from data_loaders.py =================
-    
+
     def set_subarea(boundary_dataset):
         return boundary_dataset[boundary_dataset.index == shape_index].iloc[0].geometry
-    
+
     shape_index = int(
             location._geography_choose[location.area_subset][location.cached_area]
         )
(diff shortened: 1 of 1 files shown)


climakitae/warming_levels.py
```diff
@@ -3,6 +3,7 @@
 import xarray as xr
 import holoviews as hv
 from holoviews import opts
+import bokeh.palettes
```
nicolejkeeney, line 6: when is this package used? If it isn't used, please delete!

climakitae/warming_levels.py
```diff
@@ -301,54 +302,64 @@ def _GCM_PostageStamps_MAIN(self):
             vmax_l.append(vmax_i)
         vmin = min(vmin_l)
         vmax = max(vmax_l)
-    
-        # Make each plot 
+
+        # Colormap normalization for hvplot -- only for relative humidity!
+        if self.variable2 == "Relative Humidity":
+            cmap = _read_ae_colormap(cmap="ae_diverging", cmap_hex=True)
```
nicolejkeeney, line 308: looks good to me

nicolejkeeney (approved):
after removing unwanted package, looks good to me

climakitae/warming_levels.py
vicford: Sorry - this was a holdover from doing this manually, but not needed for this version! Will delete!

climakitae/warming_levels.py
```diff
@@ -85,18 +85,18 @@ def set_subarea(boundary_dataset):
...
     for attr_np in ["earth_radius","latitude_of_projection_origin","longitude_of_central_meridian"]:
         postage_data[proj].attrs[attr_np] = postage_data[proj].attrs[attr_np].item()
-    
-    # Add grid-mapping attr (missing from Rel Humidity) 
-    if "grid_mapping" not in postage_data.attrs: 
-        postage_data.attrs["grid_mapping"] = proj 
-    
+
+    # Add grid-mapping attr (missing from Rel Humidity)
+    if "grid_mapping" not in postage_data.attrs:
```
bkg, line 97: Is this missing because RH is a derived variable for hourly? If so, it would be better to fix this upstream so any downstream code doesn't need this check.

climakitae/warming_levels.py
```diff
@@ -301,54 +301,64 @@ def _GCM_PostageStamps_MAIN(self):
...
         clabel = self.variable2 + " ("+self.postage_data.attrs["units"]+")"
         cmap_name = var_descrip[self.variable2]["default_cmap"]
         cmap = _read_ae_colormap(cmap=cmap_name, cmap_hex=True)
         vmin, vmax, sopt = _compute_vmin_vmax(min_data,max_data)
-        
+
+        if self.variable2 == "Relative Humidity":
+            cmap = _read_ae_colormap(cmap='ae_diverging', cmap_hex=True)
+        else:
+            cmap_name = var_descrip[self.variable2]["default_cmap"]
```
bkg, line 359: Looks like this is a repeat of L352 so this one isn't needed.

bkg (commented):
Nice diverging colormap! Just left a few comments for you.

climakitae/warming_levels.py
vicford: Thanks @bkg for all of these catches! I've made those changes

climakitae/warming_levels.py
bkg: Got it, thanks for the explanation. I see that grid mapping is assigned for [other vars](https://github.com/cal-adapt/climakitae/blob/main/climakitae/derive_variables.py#L48), so it should just be a matter of adding RH to that list, but I'll make a separate issue for that.

bkg (approved):

PR #97, opened 2022-10-07
Catalog restructure

nicolejkeeney:
This PR is a massive restructuring of the code to integrate the new data catalog and its new structure to data selection and retrieval. Because so many files have been changed and so much of the code structure has been modified, I am hoping folks can give this a thorough review to avoid any issues I created or missed when writing the code.

Thank you everyone in advance for your time looking through this beefy PR!!! I know it's not always fun! 

## How to review this PR:  
I broke up how different people may want to review this PR so folks aren't overwhelmed trying to review all the changes in all the files. Please ignore my suggestions if you'd prefer to review this in a different way. 

### Everyone
Mess around with selecting and retrieving code for different selection options; timescale, area subset, scenarios, variables, unit conversions, etc. You can do this with: `app.select()` and then `app.retrieve()` and/or with the various explore panels followed with `app.retrieve()`

### Scientists
- Confirm that the unit conversions perform as expected.
- Daily and monthly data are no longer computed on the fly, and have different variables than the hourly data. Review the variable_catalog.csv file in climakitae/data to confirm that the extended_descriptions, display names, and default colormaps for all of the variables are suitable. Please do not edit the columns variable_id, units, or long_name; those all come directly from the catalog. **The column "show" indicates whether or not we want users to play with the variable**; anything set to False will not be accessible to the users through climakitae. Please review that column as well; some variables set to False we may want to set to True. 

### Developers
Confirm that data retrieval performs as expected: data retrieval from AWS and area subsetting. This is in climakitae/data_loaders.py 

### Cora & Victoria 
Along with the tasks for everyone & scientists, please review the code changes I made to tmy.py and thresholds.py to ensure that everything still works as expected. I took the liberty of adding back in the full location options; if you'd like to stick with the subset you had previously set, just let me know and I will switch it back :D 
**Also note that I changed both your default location subsets to Los Angeles County.** Cora, I did this for your panel because the Sacramento county subset at 45km resolution is only 1 or maybe 2 grid cells. LA County is bigger and at 45km has more grid cells, which I think is better to have for analysis. Victoria, I did this because it makes the panel spin up a lot faster (2 min). LMK if either of you don't like this change and I can move it back to the OG subsets.


elehmer (commented):
Just thought I'd post this review prematurely as my first comment would appear to need fixing. Although I haven't run across a situation that it's triggered yet. I'll keep working through the notebooks to see if all is working.

climakitae/catalog_utils.py
```diff
@@ -0,0 +1,43 @@
+"""Helper functions for working with ESM catalog and user data selections"""
+
+def _convert_resolution(resolution, reverse=False): 
+    """Convert resolution format to grid_label format matching catalog names. 
+    Set reverse=True to get resolution format from input grid_label.
+    """
+    res_dict = {"45km":"d01","9km":"d02","3km":"d03"} 
+    
+    if reverse==True: 
+        res_dict = {v: k for k, v in res.items()}
```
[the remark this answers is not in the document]
nicolejkeeney: You have definitely found a typo! I will fix that, thanks for finding the mistake!

bethmints:
so far I've tested data retrieval. I did it quite thoroughly, though not exhaustively. 

- not sure if this is known, but data were missing at 3 km resolution for all tested combinations of time scale, variable, scenario in the following states: CO, NM, WY, ID, MT, WA. fwiw, this occurred whether selecting by state or lat/lon (figured it was worth testing just in case).

otherwise...
- all other tested combinations of spatial range/resolution, time range/resolution, and variable were successful
- temperature, pressure, water vapor mixing ratio, and wind unit conversions look good
- retrieve() was very zippy on my end

bethmints:
alright, checking out the CSV:

- colormaps:
 - should evaporation and evapotranspiration be diverging, since they can be negative? LH flux is, which is good.
 - every variable shows ae_orange in `app.view()`. looking at `view.py`, it looks like this is the default if no cmap is supplied by the user - but I would expect it to read the cmap from the CSV. perhaps that functionality doesn't exist yet, I'm not exactly up-to-date on this.

- long names and extended descriptions look good to me overall. maybe expand on acronyms used in them (e.g., qv).

nicolejkeeney:
> 

Hi Beth, I think I was able to fix the area subset options for the states :D

bethmints:
nice work!

- 3 km data selection is fixed

- re: evaporation cmap - looks like a typo: `ValueError: Supplied cmap ae)diverging not found among matplotlib, bokeh, or colorcet colormaps.`

- for LH flux, the diverging cmap is not centered at 0. I assume this is the case for other diverging variables. hvplot does have the option `symmetric = True`, which is the closest I could find to centering on 0. maybe not super great when negative and positive values have different order of magnitude, but can possibly make negative and positive values clearer.

nicolejkeeney:
> * )

Thanks Beth!! Glad you found the typo!

vicford:
Also going to follow-on/clarify on Beth's comment about the evaporation colormap: 

The evaporation variable _can_ be negative (representing condensation), however by if all values are positive (representing all evaporation), the colormap should be ae_blue (assuming here blues for moisture variables). If negative values are present, it should be ae_diverging and centered around 0.

vicford:
Also just going to point out that some of the variable descriptions get sliced mid-word when looking at the attributes, but not in the app.select panel, which I imagine most of our users will focus on instead of the way it is displayed for users. I don't think this is a bug or anything that should/can(?) be fixed, just going to point it out. Examples below: 

Screen Shot 2022-10-13 at 11 44 18 AM
Screen Shot 2022-10-13 at 11 45 03 AM

vicford:
Are we no longer ordering the variables based off of "priority" in the monthly/daily/hourly lists as they get displayed in the dropdown menus?

nicolejkeeney:
> Also just going to point out that some of the variable descriptions get sliced mid-word when looking at the attributes, but not in the app.select panel, which I imagine most of our users will focus on instead of the way it is displayed for users. I don't think this is a bug or anything that should/can(?) be fixed, just going to point it out. Examples below:
> 
> Screen Shot 2022-10-13 at 11 44 18 AM Screen Shot 2022-10-13 at 11 45 03 AM

Unfortunately I don't think there's anything we can do about this on our end.

nicolejkeeney:
> Are we no longer ordering the variables based off of "priority" in the monthly/daily/hourly lists as they get displayed in the dropdown menus?

I would like to re-order the variables but am not sure the best way to do so. Added that task into catalog backlog a while ago: https://trello.com/c/sx3uC5w5/244-backlock-catalog-to-do

nicolejkeeney:
> Getting this **warning** for daily historical ice water path at 9km for Fresno County as well as for daily evaporation at 9km for 2006-2012 for Fresno County.
> 
> Revised: the warning appears to be happening with finer resolutions for the daily data, but still does retrieve data
> 
> ```
> /srv/conda/envs/notebook/lib/python3.9/site-packages/xarray/core/indexing.py:1226: PerformanceWarning: Slicing is producing a large chunk. To accept the large
> chunk and silence this warning, set the option
>     >>> with dask.config.set(**{'array.slicing.split_large_chunks': False}):
>     ...     array[indexer]
> 
> To avoid creating the large chunks, set the option
>     >>> with dask.config.set(**{'array.slicing.split_large_chunks': True}):
>     ...     array[indexer]
>   return self.array[key]
> ```

This would be a good question for @bkg

vicford (changes requested):
I've now given this a really thorough read through. This is very impressive, and really thorough work @nicolejkeeney -- nicely done!

- app.select & app.retrieve work as expected
- unit conversions also look good to me
- variable catalog looks good to me
- tmy.py changes look good to me, I especially like that units is now included in this because it places the panel into context for degF and that is A+ for user understanding (if used to US units)
- all the explore panels are super speedy now which is really cool to see
- warming_levels panel looks great
- amy panel looks great -- really love the units thrown in there for context
- thresholds panel looks great to the best of my knowledge

A couple of things: 
- **Using a cluster to test the explore panels does not work**. I've check in with Nicole & Brian about this, and it does appear to be occurring when there is a compute() stage behind the scenes. The app.explore panels **do not** need a cluster, however, this will likely cause a problem for all of our other notebooks that do use a cluster once this branch gets merged into main, and therefore should be figured it prior to merging. It is curious to me that we are not using a cluster for the AMY calculations since it is heavily dependent on it for that compute stage. 

- When testing app.explore.warming_levels, I was noticing that the warming_levels panel does not update, **only when** app.explore.amy() has been called after app.explore.warming_levels() and then the user went back to the existing warming_levels panel to continue playing around there. I've attached a few screenshots to demonstrate this -- even after changing the selections in warming_levels panel, the kernel runs for a minute or so, but the resulting postage stamps do not update. While I think that it is unlikely that a user will go back and forth between these panels as they _should_ be used in separate notebooks, it's not unreasonable to think that a user may just play around with the panels in the same notebook if they are just seeing what the panels do. I'm not sure if this can or should be fixed, but something to consider. 
For example, below, I made changes to the data selections panel, hit "reload data" and no maps updated -- you can see a difference between the selections and what is plotted. Again this only happens if app.explore.amy() was called after initiating app.explore.warming_levels() and then the user returned to warming levels. AKA, process: 
1. Initiate app.explore.warming_levels()
2. Initiate app.explore.amy() in new cell, while warming_levels panel is still up. 
3. Return to warming_levels panel
4. Make new data selections in warming_levels panel and hit 'reload data'
Screen Shot 2022-10-17 at 11 11 43 AM

- Differences in AMY (and occasionally in warming_levels) using the ae_diverging colormap are not quite right. This is a separate but should be addressed soon issue. I have a screenshot to demonstrate what's going on here. I don't think this is dependent on a successful merge of this branch because I think this is going to be more involved than a simple switch, and we should think about a more programmatic switch rather than assigning a colormap (this is also on me too). Along the lines of, if the minimum of a dataarray/set is a negative value, then the colormap should be automatically set to ae_diverging **and** centered at 0. Again, I think this is a separate fix. 
For example, below this is the difference in air temp that is displayed with the ae_diverging colormap, even though there are no negative values -- this should be ae_orange instead. 
Screen Shot 2022-10-17 at 11 59 22 AM

elehmer:
> > Are we no longer ordering the variables based off of "priority" in the monthly/daily/hourly lists as they get displayed in the dropdown menus?
> 
> I would like to re-order the variables but am not sure the best way to do so. Added that task into catalog backlog a while ago: https://trello.com/c/sx3uC5w5/244-backlock-catalog-to-do

Ordering data for presentation in the GUI with a custom ordering can be accomplished for Pandas DataFrames by using a similar approach to the one I used in `get_us_states` in `selectors.py`. You can make the data `Categorical` and then you can sort that categorical data using a list.

climakitae/selectors.py
```diff
@@ -295,6 +299,57 @@ def _update_location(self):
...
+        var_catalog (df): variable catalog information. read in from csv file 
+        unique_variable_ids (list of strs): list of unique variable ids from catalog. Used to subset var_catalog 
+        timescale (str): hourly, daily, or monthly 
+        
+    Returns: 
+        variable_options_df (pd.DataFrame): var_catalog information subsetted by unique_variable_ids 
+    
+    """
+    if timescale in ["daily","monthly"]: 
+        timescale = "daily/monthly"
```
thenaomig, line 344: why are these combined as one thing?

climakitae/selectors.py
```diff
@@ -303,64 +358,149 @@ class DataSelector(param.Parameterized):
...
-        _list_of_scenarios = list(self.choices["scenarios"]["45 km"].keys())
-        self.param["scenario"].objects = _list_of_scenarios
-        self.scenario = ["Historical Climate"]
-        self.param["variable"].objects = self.choices["variable_choices"]["hourly"][
-            "Dynamical"
-        ]
-        self.units = self.descrip_dict[self.variable]["native_unit"]
+        
+        # Dataset selection 
+        self.dataset = "WRF"
```
thenaomig, line 390: .dataset is not a good name for this, given that it is also the word used for an xarray Dataset... can we call it 'product'? or 'downscaling_method'?

climakitae/threshold_panel.py
```diff
@@ -29,112 +53,67 @@ class ThresholdDataParams(param.Parameterized):
...
-    area_subset2 = param.ObjectSelector(default="CA counties", objects=["CA counties", "states"])
-
     def __init__(self, *args, **params):
         super().__init__(*args, **params)
         
         # Selectors defaults
         self.selections.append_historical = False
         self.selections.area_average = True
-        self.selections.resolution = "45 km"
+        self.selections.resolution = "45km"
```
thenaomig, line 62: Seems to me that a space is appropriate grammar-wise, since this is used in display text?

thenaomig (commented):
Here are some comments for the time-being.

climakitae/selectors.py
nicolejkeeney: Because the variables (and therefore the descriptions, units, etc) are the same for both daily and monthly data, while the hourly variables are different

climakitae/threshold_panel.py
nicolejkeeney: Personally I prefer it without the space. Perhaps we can ask the larger group and see what everyone thinks?

nicolejkeeney: Addressed in PR https://github.com/cal-adapt/climakitae/pull/114

climakitae/data_loaders.py
```diff
@@ -147,43 +126,156 @@ def set_subarea(boundary_dataset):
             shape = set_subarea(location._geographies._ca_watersheds)
         ds_region = [shape]
     else:
-        ds_region = None
-
-    if selections.append_historical:
-        if not any(['SSP' in s for s in selections.scenario]):
-            raise ValueError('Please also select at least one SSP to '
-                     'which the historical simulation should be appended.')
```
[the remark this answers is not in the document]
nicolejkeeney: Addressed in PR https://github.com/cal-adapt/climakitae/pull/114

climakitae/catalog_utils.py
```diff
@@ -0,0 +1,43 @@
+"""Helper functions for working with ESM catalog and user data selections"""
+
+def _convert_resolution(resolution, reverse=False):
```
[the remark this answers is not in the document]
nicolejkeeney: Addressed in PR https://github.com/cal-adapt/climakitae/pull/114

climakitae/data_loaders.py
```diff
@@ -147,43 +126,156 @@ def set_subarea(boundary_dataset):
...
+            geometries=ds_region, 
+            crs=4326, 
+            drop=True
+        )
+    
+    # Perform area averaging 
+    if selections.area_average == True: 
+        weights = np.cos(np.deg2rad(da.lat))
+        da = da.weighted(weights).mean("x").mean("y")
+
```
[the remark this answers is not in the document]
nicolejkeeney: Addressed in PR https://github.com/cal-adapt/climakitae/pull/115

nicolejkeeney:
> Getting this **warning** for daily historical ice water path at 9km for Fresno County as well as for daily evaporation at 9km for 2006-2012 for Fresno County.
> 
> Revised: the warning appears to be happening with finer resolutions for the daily data, but still does retrieve data
> 
> ```
> /srv/conda/envs/notebook/lib/python3.9/site-packages/xarray/core/indexing.py:1226: PerformanceWarning: Slicing is producing a large chunk. To accept the large
> chunk and silence this warning, set the option
>     >>> with dask.config.set(**{'array.slicing.split_large_chunks': False}):
>     ...     array[indexer]
> 
> To avoid creating the large chunks, set the option
>     >>> with dask.config.set(**{'array.slicing.split_large_chunks': True}):
>     ...     array[indexer]
>   return self.array[key]
> ```

Addressed in PR https://github.com/cal-adapt/climakitae/pull/114

nicolejkeeney:
> Are we no longer ordering the variables based off of "priority" in the monthly/daily/hourly lists as they get displayed in the dropdown menus?

Addressed in PR https://github.com/cal-adapt/climakitae/pull/113

climakitae/data_loaders.py
```diff
@@ -147,43 +126,156 @@ def set_subarea(boundary_dataset):
...
+                sim_list.append(dsets[filename][selections.variable_id])  
+
+        # Concatenate along simulation dimension
+        da = xr.concat(
+            sim_list, 
+            dim="simulation", 
+            coords='minimal', 
+            compat='override'
+        )
+        da_list.append(da.assign_coords({"scenario":da_name}))
```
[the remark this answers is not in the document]
nicolejkeeney: Perhaps nice to tackle in a separate PR?

elehmer:
Minor styling improvement in `warming_levels.py`. Please `git pull` again...

elehmer:
After @bkg looked it over I have introduced some styling issues. This will all be resolved in another PR after merging the other small PRs outstanding. Please live with my styling until then lol

climakitae/data_loaders.py
```diff
@@ -148,42 +134,158 @@ def set_subarea(boundary_dataset):
...
+        "units": da_final.attrs["units"]
+    }
+    return da_final
+
+# ============ Read from catalog function used by ck.Application ===============
+
+def _read_from_catalog(selections, location, cat):
+    """
+    The primary and first data loading method, called by
+    core.Application.generate, it returns a DataArray (which can be quite large)
```
ckingdon95, line 239: `generate` is now `retrieve` right?

climakitae/threshold_panel.py
```diff
@@ -1,161 +1,178 @@
...
+from .threshold_tools import (
+    get_exceedance_count,
+    plot_exceedance_count,
+    _exceedance_plot_title,
+    _exceedance_plot_subtitle
+)
+
+# ============ Class and methods for the explore.thresholds() GUI ==============
+
+def _get_threshold_data(selections, location, cat):
```
ckingdon95, line 16: these modifications to thresholds panel data loading look great to me!

ckingdon95 (commented):
Looks great! I tested some selections on the hub, everything worked as expected for me! Left two small comments here

climakitae/core.py
```diff
@@ -1,72 +1,71 @@
...
     def retrieve(self):
         """
-        Uses the information gathered in 'select' and stored in 'selections' and 'location'
-        to generate an xarray DataArray as specified, and return that DataArray object.
+        Uses the information gathered in 'select' and stored in 'selections' and
+        'location' to generate an xarray DataArray as specified, and return that
+        DataArray object.
         """
-        # to do: insert additional 'hang in there' statement if it's taking a while
+        # TODO: insert additional 'hang in there' statement if it's taking a while
```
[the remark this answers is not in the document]
nicolejkeeney: It was already there, perhaps Eric renamed it when fixing the formatting which is why it's showing up as part of this PR

climakitae/data_loaders.py
nicolejkeeney: yes, i will update that comment. i had just copy/pasted

Issue #98, opened 2022-10-13
DataArray Attributes are lost when running `get_anomaly_data` in warming_levels.py


elehmer:
Data fed to `get_anomaly_data` loses its attrs in the output. This is because a new Xarray DataSet is created and populated but the attrs are not transferred:

https://github.com/cal-adapt/climakitae/blob/308454fa29f4ba75a0cfc95e6ca695694a1d254e/climakitae/warming_levels.py#L120

This causes downstream issues with exporting as projection information is lost.

vicford:
It is worth checking in the explore_amy notebook as to whether this also occurs with the separate tmy_calc calculation, as this follows a similar process. Will check on this now.

elehmer:
Fixed in #99

PR #99, opened 2022-10-14
Copy input attrs to output data in get_anomaly_data

elehmer:
This is a simple fix for #98 by setting the new Xarray DataSet created attributes to the attributes of the input DataArray. @vicford Can you check if all these attrs are appropriate to keep in the output data. If not we can pop out those ones.

### DIFF
--- a/climakitae/warming_levels.py
+++ b/climakitae/warming_levels.py
@@ -111,13 +111,14 @@ def get_anomaly_data(data, warmlevel=3.0):
         warmlevel (float): warming level
 
     Returns:
-        warm_all_anoms (xr.DatArray): warming level anomalies computed from input data
+        warm_all_anoms (xr.DataArray): warming level anomalies computed from input data
     """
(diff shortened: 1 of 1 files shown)


vicford (approved):
Yep - all of these should be kept. I went back and forth on whether the window years should be kept, but as we add more models these will change so I do think it is important for clarity at each step that we are explicit about which years are being used computationally. 

Looks good to me - thanks for handling this

Screen Shot 2022-10-14 at 9 53 15 AM

elehmer:
I probably should have branched off of #97 instead of #main...

nicolejkeeney (approved):
Looks good!

nicolejkeeney:
> I probably should have branched off of #97 instead of #main...

I don't think it should cause any merge conflicts actually!

elehmer:
> > I probably should have branched off of #97 instead of #main...
> 
> I don't think it should cause any merge conflicts actually!

Yes, this is a simple fix and will merge nicely. Going for merge now.

PR #101, opened 2022-10-14
clearer messaging + minor export functionality

bethmints:
### about this pull request

We are currently not accepting multivariate datasets for CSV (could result in massive files, so we're just not recommending this right now) or GeoTIFF (impossible unless variable is the 'band' dim; this was tabled for now to simplify things for the beta launch). This PR throws an error with a more useful message than it had before (see attached image). It also adds the very minor functionality of allowing a single-variable dataset for these formats, should it occur. 

To test, try exporting multivariate and single-variable datasets (not data arrays) as CSV or GeoTIFF. Look for relevant message (in the former case) and success (in the latter).

image

### DIFF
--- a/climakitae/data_export.py
+++ b/climakitae/data_export.py
@@ -174,7 +174,8 @@ def _export_to_user(user_export_format,data_to_export,
     path_and_name = './'+file_name
     req_format = user_export_format.output_file_format
     if req_format is None:
-        raise AssertionError("Please select a file format from the dropdown menu.")
+        raise AssertionError("Please select a file format from the"+
+                             " dropdown menu.")
     
     extension_dict = {'NetCDF' : '.nc',
                       'CSV' : '.csv.gz',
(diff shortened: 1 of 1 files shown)


climakitae/data_export.py
```diff
@@ -238,10 +240,28 @@ def _export_to_user(user_export_format,data_to_export,
     if ("NetCDF" in req_format):
         export_to_netcdf(data_to_export,save_name,**kwargs) 
     else:
-        assert "DataArray" in str(type(data_to_export)),("We are only"+
-                             " converting data arrays to this format"+
-                             " at this time, please pass a data array"+
-                             " (not a dataset). HINT:" ) 
+        if ("DataArray" not in str(type(data_to_export))):            
+            dv_list = list(data_to_export.data_vars)
+            assert (len(dv_list) == 1
```
elehmer, line 245: If can be an `if...raise` instead of `assert`

climakitae/data_export.py
```diff
@@ -204,8 +205,8 @@ def _export_to_user(user_export_format,data_to_export,
         
     assert "xarray" in str(
```
elehmer, line 206: This should really be a `if..raise` instead of `assert` which is only really used in debugging.

Here is an example:

```
x = 10
if x > 5:
    raise Exception('x should not exceed 5. The value of x was: {}'.format(x))
```

elehmer (changes requested):
Please see my comments. It will all run as is, but want to start banging the code into form...

Could you also and a line break at the end of the file?

Thanks!

vicford (approved):
This is really nice, Beth. I also really like the guidance in the assertion error as to how to "fix" it. I tested this in the threshold_basic notebook, which has the "fix" already implemented.

PR #104, opened 2022-10-18
Typical meteorological year data retrieval fix 

nicolejkeeney:
Fix the tmy.py script such that when a user calls app.retrieve() after the panel, the expected data is returned.

### DIFF
--- a/climakitae/tmy.py
+++ b/climakitae/tmy.py
@@ -203,15 +203,6 @@ class AverageMeteorologicalYear(param.Parameterized):
     def __init__(self, *args, **params):
         super().__init__(*args, **params)
 
-        # Selectors defaults
-        self.selections.append_historical = False
-        self.selections.area_average = True
-        self.selections.resolution = "45 km"
-        self.selections.scenario = ["Historical"]  # setting for historical
-        self.selections.time_slice = (1981,2010)
-        self.selections.timescale = "hourly"
-        self.selections.variable = "Air Temperature at 2m"
-
         self.units2 = self.selections.descrip_dict[self.selections.variable]["native_unit"]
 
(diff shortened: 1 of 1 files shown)


elehmer (approved):
LGTM

vicford (commented):
Testing this at the app.retrieve() stage, I'm getting the following error: 

`KeyError: 'Historical'`

It seems like it's not translating to "Historical Climate"? I tried restarting my kernel twice to confirm. Although I see in the code that it should do this now... will check in with @nicolejkeeney about this

elehmer:
> Testing this at the app.retrieve() stage, I'm getting the following error:
> 
> `KeyError: 'Historical'`
> 
> It seems like it's not translating to "Historical Climate"? I tried restarting my kernel twice to confirm. Although I see in the code that it should do this now... will check in with @nicolejkeeney about this

That's weird that I didn't run into it. I guess I always moved off of the defaults. Good catch @nicolejkeeney !

vicford (approved):
With the historical climate fix, it works brilliantly! Much appreciated on this one @nicolejkeeney!!

bethmints:
when selecting _Absolute_ under AMY Type, and _Historical_ under Computation Options, I still see _Warming level_ options, and 1.5'C is automatically selected. Yesterday I noted these outputs:

`app.selections.scenario` 
`['SSP 3-7.0 -- Business as Usual']`

`app.selections.time_slice`
`(2034, 2063)`

and that the panel and amy output did not match. but now everything is consistent with expectations! very nice.

minor note: ideally the Warming Level option won't show if _Historical_ is selected, but that's minor, and I know that panel is tricky. well done!

PR #105, opened 2022-10-18
clearer error messages; formatting

bethmints:
### About this PR:

- Changed errors to exceptions
- Spruced up some error messages related to data dimensionality when exporting to geotiff
- Wrote clearer error message when trying to export multivariate dataset to CSV or geotiff

### To test:

- Attempt to export datasets of various shapes to geotiff
- Attempt to export multivariate datasets to CSV or geotiff.

Let me know how the error messages look.

### DIFF
--- a/climakitae/data_export.py
+++ b/climakitae/data_export.py
@@ -89,7 +89,7 @@ def export_to_geotiff(data_to_export, save_name, **kwargs):
         sim_attrs = {'simulation' : str(data_to_export.coords['simulation'].values)}
         ds_attrs = dict(ds_attrs, **sim_attrs)
     if ((str('simulation') in data_to_export.dims) and
-    (len(data_to_export.simulation)==1)):
+        (len(data_to_export.simulation)==1)):
         sim_attrs = {'simulation' : str(data_to_export.simulation.values)}
         ds_attrs = dict(ds_attrs, **sim_attrs)
         data_to_export = data_to_export.squeeze(dim='simulation')
@@ -148,21 +148,21 @@ def _export_to_user(user_export_format, data_to_export,
 
     if type(file_name) is not str:
         raise Exception("Please pass a string"
-        + " (any characters surrounded by quotation marks)"
-        + " for your file name.")
+                        + " (any characters surrounded by quotation marks)"
+                        + " for your file name.")
 
     path_and_name = './' + file_name
     req_format = user_export_format.output_file_format
     if req_format is None:
         raise Exception("Please select a file format"
-                             + " from the dropdown menu.")
+                        + " from the dropdown menu.")
 
     extension_dict = {'NetCDF' : '.nc',
                       'CSV' : '.csv.gz',
                       'GeoTIFF' : '.tif'}
 
     f_extension = extension_dict[req_format]
-    save_name = path_and_name+f_extension
+    save_name = path_and_name + f_extension
 
     if os.path.exists(save_name):
         raise Exception("File " + save_name + " exists,"
@@ -239,10 +239,10 @@ def _export_to_user(user_export_format, data_to_export,
                 var_name = dv_list[0]
                 data_to_export = data_to_export.to_array()
                 data_to_export.name = var_name
-                
+
         if ("CSV" in req_format):
             export_to_csv(data_to_export, save_name, **kwargs)
-            
+
         elif ("GeoTIFF" in req_format):
             # if x and/or y exist as coordinates
             # but have been squeezed out as dimensions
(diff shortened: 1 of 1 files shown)


climakitae/data_export.py
```diff
@@ -167,28 +146,29 @@ def _export_to_user(user_export_format, data_to_export,
     ndims = len(data_to_export.dims)
```
elehmer, line 146: We might need to have some logic up front here to catch passing non-Xarray data to export. This `dims` call will only work with a Xarray `DataSet` or `DataArray` so the code will bomb here if passed something else. Also the logic later (line 186) won't have a chance to be called...

elehmer (changes requested):
We are getting close! Just a couple of questions. Thanks!

I pushed some formatting commits so make sure you do a `git pull` on your local branch.

vicford (commented):
`Exception: No x dimension or coordinate exists; cannot export to GeoTIFF. Please provide a data array with both x and y spatial coordinates.`

On exporting a timeseries (1x5x288) to geotiff

vicford:
`Exception: We cannot convert multivariate datasets to CSV or GeoTiff at this time. Please supply a dataset or array with a single data variable. A single variable array can be extracted from a multivariate dataset like so: app.export_dataset(ds['var'],'filename') where ds is the dataset or data array you attempted to export, and 'var' is a data variable (in single or double quotes).`

On exporting a multivariate dataset to csv (same with geotiff)

climakitae/data_export.py
bethmints: thanks for pointing this out! I did indeed have the xarray object check too late

climakitae/data_export.py
```diff
@@ -239,27 +218,71 @@ def _export_to_user(user_export_format, data_to_export,
     if ("NetCDF" in req_format):
         export_to_netcdf(data_to_export, save_name, **kwargs) 
     else:
-        assert "DataArray" in str(type(data_to_export)), ("We are only"
-                             + " converting data arrays to this format"
-                             + " at this time, please pass a data array"
-                             + " (not a dataset). HINT:" )
+        if "DataArray" not in str(type(data_to_export)):
```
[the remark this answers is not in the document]
bethmints: `.data_vars` is a `Dataset` property, not a `DataArray` property. a `Dataset` can have multiple data variables. this condition is encountered after the `xarray` object check has passed, so we know it's either a `Dataset` or `DataArray`. 

why am I using a `not`? because I swear that `type(DataArray)` returned `xarray.core.dataset.DataArray`, so using `if "Dataset" in....` as the conditional would not work... but now I'm seeing  `xarray.core.dataarray.DataArray`. 

I think that perhaps I wasn't using `pip install requirements.txt`, so I might have been using the wrong xarray version before. next commit will check for `Dataset` rather than `not DataArray`.

climakitae/data_export.py
bethmints: thanks, I absolutely had that logic check in the wrong place, lol

climakitae/data_export.py
```diff
@@ -163,32 +142,40 @@ def _export_to_user(user_export_format, data_to_export,
...
-        raise AssertionError("Please select a file format"
-                             + " from the dropdown menu.")
+        raise Exception("Please select a file format"
+                        + " from the dropdown menu.")
 
     extension_dict = {'NetCDF' : '.nc',
                       'CSV' : '.csv.gz',
                       'GeoTIFF' : '.tif'}
 
     f_extension = extension_dict[req_format]
```
elehmer, line 171: f_extension isn't used more than once so you can just move the `extension_dict` to the line below...

climakitae/data_export.py
```diff
@@ -163,32 +142,40 @@ def _export_to_user(user_export_format, data_to_export,
...
 
-    assert type(file_name) is str,("Please pass a string"
-        + " (any characters surrounded by quotation marks)"
-        + " for your file name.")
+    if type(file_name) is not str:
+        raise Exception("Please pass a string"
+                        + " (any characters surrounded by quotation marks)"
+                        + " for your file name.")
 
     path_and_name = './' + file_name
```
elehmer, line 161: This is only used once on Line 172 so you can just move this down there

elehmer (changes requested):
Just a few improvements...

climakitae/data_export.py
```diff
@@ -239,27 +221,76 @@ def _export_to_user(user_export_format, data_to_export,
...
+                    data_to_export = data_to_export.expand_dims('y')
+                else:
+                    raise Exception("No y dimension or coordinate exists;"
+                                    + " cannot export to GeoTIFF. Please provide"
+                                    + " a data array with both x and y"
+                                    + " spatial coordinates.")
+
             dim_check = data_to_export.isel(x=0, y=0).squeeze().shape
-            shape_str = str(dim_check).strip('(').strip(')').replace(", ", " x ")
+            print(dim_check)
```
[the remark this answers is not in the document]
bethmints: oops!

elehmer:
Thanks for the updates! I'll run it through its paces and see if it works!

elehmer (approved):
I think we got it!!!! Great job and thanks!! I pushed a minor whitespace/formatting commit.

I've run it on the HUB and it seems good!

vicford (approved):
LGTM

Issue #109, opened 2022-10-20
Heatmap options...can this one be added to line above?


elehmer:
When creating the heatmap object you set `opts` and then set another `opts`. Can this be added to the line above???

https://github.com/cal-adapt/climakitae/blob/984d67e7950f333fc019b7b1c523fbe931b7b641/climakitae/tmy.py#L371

vicford:
Testing now, yes! Will push clean now.

PR #110, opened 2022-10-20
Cleans up toolbar options for AMY heatmap

vicford:
Quick fix for Issue https://github.com/cal-adapt/climakitae/issues/109

Just moves the toolbar statement into a single .opts() call with minor whitespace cleaning

To test: 
Just make sure that the AMY heatmap in explore_amy notebook (or by calling app.explore.amy) works as desired.


elehmer (approved):
LGTM!

PR #115, opened 2022-10-24
Order of operations

nicolejkeeney:
Change order of operations by which the data is subsetted in `data_loaders.py`


climakitae/data_loaders.py
```diff
@@ -245,6 +245,35 @@ def _get_data_one_var(selections, location, cat):
     
     # Read data from AWS.
     data_dict = _get_data_dict_and_names(cat_subset = cat_subset)
+    
+    # Perform subsetting operations
+    for dname, dset in data_dict.items():
```
thenaomig, line 250: is it necessary to loop over these?

thenaomig: I think this would be better done inside process_and_concat, calling a subsetting function before appending each da

nicolejkeeney: Yeah I agree that perhaps it would be more efficient to loop over them in process_and_concat. My concern with doing the subsetting within the process_and_concat loop is that the code is already a bit "messy" (or, not super intuitive) and I like that doing it the way I have in a separate loop is cleaner and easier to read for future developers down the line. 

In general, I have found app.retrieve() to be really fast with this code so I think having the subsetting separate from the historical appending loop is nice for clarity's sake. But perhaps if we notice any weird speed issues with retrieving data we can revisit? Has app.retrieve() been fast for you as well?

vicford: Tested on three different variables/resolutions/timescales and app.retrieve did not take longer than 3.11 seconds for me

thenaomig: The timing of app.retrieve is not the issue though. It's later when you actually go to output something (a plot, or to a file) that it actually does all the data manipulation, so I think we'd need to do that as well to properly benchmark.

nicolejkeeney: Sorry for the confusion. I meant, retrieving data in general is fast, even when calling data.compute()

nicolejkeeney: You might be right that a loop isn't a good home for this but I'd be keen to address that further down the line if we notice it is a problem.

PR #118, opened 2022-10-31
Remove preprocessing function from data loading

nicolejkeeney:
This should hopefully fix the cluster killed worker issue in the catalog_restructure branch

**To test:** Try running this in the AE hub with a cluster. Code for spinning up a cluster is below. 
```
from dask_gateway import GatewayCluster
cluster = GatewayCluster()
client = cluster.get_client()
cluster.scale(8)
```

### DIFF
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -22,13 +22,6 @@ dask.config.set({"array.slicing.split_large_chunks": True})
 
 # ============================ Helper functions ================================
 
-def add_sim_coord(ds):
-    """Add simulation as Dataset coords and dimensions.
-    Used when reading in data from catalog. 
-    """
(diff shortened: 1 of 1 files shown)


elehmer:
Oh, this is interesting. I could have sworn I commented this out and tried to run it and it blew up still. I'll have another look when I'm feeling better...

vicford (approved):
Both warming levels and AMY have no issue with a cluster!

bethmints:
the cluster in the AMY notebook seems to work fine, but I'm seeing a couple errors. 

- Temperature is using the blue precip map, while the precip map itself looks a bit odd to me (attached). 
- I'm also getting an error at the tmy_calc() call (attached).

I'm not sure if these are known issues brought up in another PR, but I wanted to point them out.  

image
image

PR #120, opened 2022-11-08
Link to cached data for AMY panel

nicolejkeeney:
This PR changes the tmy.py code to link to the cached hourly ensemble means in the TMY panel. I haven't noticed that this is speeding up the panel very much, unfortunately. I am curious to see what others find.


vicford (approved):
This does work with the cached data for AMY. 

However, on timings, I also noticed that this did not significantly speed things up. I tested on three different scenarios:
1. absolute historical air temp in K for LA county, all spatial resolutions
2. absolute future air temp at 3°C in K for LA county, all spatial resolutions
3. future warming level at 3°C in K for LA county, all spatial resolutions
In general what I noticed is that working **without** a cluster runs marginally faster than with a cluster, which is fairly curious. 
45 km takes approximately ~1 min 50 seconds for absolute data / ~4 min for differences
9 km takes approximately ~ 2 min 30 seconds for absolute / ~ 4 min for differences
3 km takes approximately ~4 min 20 seconds for absolute / ~4 min for differences
With a cluster, these times added less than a minute of additional time, in general

bkg:
Interesting, thanks for wiring this up and sharing some initial timings. I'm seeing faster execution for larger areas such as CA w/ 45km as an example  calling`app.explore.amy()`:

* main branch: 4min 47s
* this branch: 2min 11s

So this does seem to speed things up for larger boundaries and/or higher spatial resolutions, which makes sense as there are 4x fewer chunks to be read for the multi-model ensemble vs the dynamically created one. It sounds like this won't move the needle for smaller spatial slices, although that could change with the introduction of more GCMs, ensemble members, etc.

I'm still checking a few more boundaries like San Bernardino County which is notoriously large in addition to the 3km grid.

PR #121, opened 2022-11-09
Add new app.load() function to read data into memory

nicolejkeeney:
This PR adds a new `app.load()` function that basically just calls .compute() on whatever data you feed it. It checks the available system memory and the memory of the data you want to read into memory, then either reads the input data into memory or raises an error. It will raise an error if there's not enough memory to read in your data. **This avoids mysterious kernel failures, instead raising an error with descriptive information for the user.**

I've set the error to be raised if loading the data would leave less than 256MB of available memory on the system. Perhaps this number should be higher/lower? 

**To test:** Try the workflow laid out below to read in data. Try with different datasets, including some data you expect to be really big (maybe decades of hourly?)-- this should return an error message upon calling `app.load(data)`. Make sure there are no kernel failures EVER when calling `app.load(data)`, and confirm that the warning and print statements returned by `app.load(data)` make sense to you. 
```
app.select() # Select some data 
data = app.retrieve() # Retrieve the data 
data = app.load(data) # Read the data into memory 
```


vicford (approved):
Tried loading hourly data for 1980-2100 for the entire domain at 3km, and very appropriately got a `MemoryError: Your input dataset is too large to read into memory!` error, lol. 

I appreciate that it also informed of the size! 
```
Available memory: 5.85 GB
Total memory of input data: 1872.74 GB
```

climakitae/data_loaders.py
```diff
@@ -20,6 +22,33 @@
...
+    if xr_da.chunks is None: 
+        print("Your data is already loaded into memory") 
+        return xr_da 
+
+    # Get memory information
+    avail_mem = psutil.virtual_memory().available # Available system memory 
+    xr_data_nbytes = xr_da.nbytes # Memory of data 
+
+    # If it will cause the system to have less than 256MB after loading the data, do not allow the compute to proceed. 
+    if avail_mem - xr_data_nbytes < 268435456:
```
elehmer, line 41: This can be represented as: `256 * (1024 ** 2)`

climakitae/utils.py
```diff
@@ -13,6 +13,24 @@
...
+def _readable_bytes(B):
+    """
+    Return the given bytes as a human friendly KB, MB, GB, or TB string.
+    Modified code from stackoverflow: https://stackoverflow.com/questions/12523586/python-format-size-application-converting-b-to-kb-mb-gb-tb
+    """
+    B = int(B)
+    KB = int(1024)
+    MB = int(KB ** 2) # 1,048,576
+    GB = int(KB ** 3) # 1,073,741,824
+
```
elehmer, line 26: Might as well add in TB:

`TB = int(KB ** 4) # 1,099,511,627,776`

climakitae/utils.py
```diff
@@ -13,6 +13,24 @@
...
+    if B < MB:
+        return '{0} {1}'.format(B,'bytes')
+    elif MB <= B < GB:
+        return '{0:.2f} MB'.format(B / MB)
+    else:
```
elehmer, line 31: Add in:

```
elif GB <= B < TB:
    return `{0:.2f} GB'.format(B / GB)
else:
    return `{0:.2f} TB'.format(B / TB)
```

elehmer (approved):
See my comments, but LGTM!

climakitae/utils.py
nicolejkeeney: Just added those conversions in, thanks for the suggestions!

Issue #108, opened 2022-10-20
Test for xarray DataArray a different way and rework asserts into if...raise statements in `timeseriestools.py`


elehmer:
The line below can be rewritten better using `type(data) == xr.core.dataarray.DataArray` 

https://github.com/cal-adapt/climakitae/blob/984d67e7950f333fc019b7b1c523fbe931b7b641/climakitae/timeseriestools.py#L254

Also this whole section should be rewritten to use `if...raise` instead of the debugging `assert` statements...

nicolejkeeney:
See PR https://github.com/cal-adapt/climakitae/pull/123

PR #123, opened 2022-11-10
Timeseries assert --> error fix

nicolejkeeney:
This PR addresses an [issue raised by Eric](https://github.com/cal-adapt/climakitae/issues/108) to replace the assert statements with error messages. I also improved the error messages. 

**To test:** Try running the timeseries_examples notebook and loading the wrong data. The following line should raise an error with an appropriate error message if you feed it "bad" data (i.e. data that can't be used by the timeseries tool) 
```
timeseries = tst.Timeseries(my_data)
```

### DIFF
--- a/climakitae/timeseriestools.py
+++ b/climakitae/timeseriestools.py
@@ -255,12 +255,19 @@ class Timeseries:
     def __init__(self, data):
         
         # Confirm that the user has input a valid dataset
-        if type(data) != xr.core.dataarray.DataArray: 
-            raise ValueError("Please pass an xarray DataArray (e.g. as output by app.retrieve()).")
-        if "lat" in data.coords:
-            raise ValueError("Please pass a timeseries (area average).")
-        if any(["Historical + " in v for v in data.scenario.values]) == False:
-            raise ValueError("Please append the historical period in your data retrieval.")
+        not_xr_da = type(data) != xr.core.dataarray.DataArray # Data is NOT in the form of xr.DataArray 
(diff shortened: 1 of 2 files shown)


vicford (changes requested):
Not quite sure what the desired behavior is if multiple of these errors are relevant. 

I tested with data that was not area averaged, and included SSP3-7.0 but I did not select "append historical". I got one of (presumably) two errors:
`ValueError: Please pass a timeseries (area average).`

vicford (approved):
Better! 
Screen Shot 2022-11-10 at 12 50 44 PM

climakitae/timeseriestools.py
```diff
@@ -253,16 +253,23 @@ class Timeseries:
...
-        ), "Please pass a timeseries (area average or individual station)."
-        assert (
-            True in ["Historical + " in v for v in data.scenario.values]
-        ), "Please append the historical period in your data retrieval."
-
+        
+        # Confirm that the user has input a valid dataset
+        not_xr_da = type(data) != xr.core.dataarray.DataArray # Data is NOT in the form of xr.DataArray 
+        not_area_averaged = "lat" in data.coords # Data is NOT area averaged 
+        not_append_historical = any(["Historical + " in v for v in data.scenario.values]) == False # Append historical = False
```
elehmer, line 260: You can just put a `not` in front of the `any` and take away the `== False`...

elehmer (changes requested):
See my comments. The logic tree needs some work...if you want I can take a stab at it.

PR #124, opened 2022-11-10
Simplify threshold tools code

ckingdon95:
This PR removes unnecessarily repetitive code in `threshold_tools` (code that is repeated within if/elif statements). This does not change the functionality/output of any of these functions. 

I also added a new unit test file: `test/threshold_tools/test_thresholds_returns.py` to make sure the get_return_value and get_return_period functions run without error.

**To review:** I've tested it out on the hub; it would be helpful for someone to just look through the additions/deletions under "Files changed" to make sure the changes look logical and shouldn't effect the functionality!


nicolejkeeney (approved):
Looks good to me, way easier to read the code now! :D

climakitae/threshold_tools.py
```diff
@@ -137,69 +135,32 @@ def get_ks_stat(ams, distr = "gev", multiple_points = True):
...
-                    ams, "gumbel_r", args = (lmoments["loc"], lmoments["scale"])
-                )
-                d_statistic = ks[0]
-                p_value = ks[1]
-            except (ValueError, ZeroDivisionError):
-                d_statistic = np.nan
-                p_value = np.nan
+            cdf = "gumbel_r"
+            args = (lmoments["loc"], lmoments["scale"])
         elif distr == "weibull":
```
odoherty, line 146: I would propose we add a card to the backlog to handle either weibull min or max behavior as a user option.  I agree that wiebull_min is the default/expected behavior and this is fine for now

odoherty (approved):

climakitae/threshold_tools.py
ckingdon95: great, I just made a card!

PR #128, opened 2022-11-16
Add return function to allow users to observe selection options from AMY panel

nicolejkeeney:
This PR adds a new function `app.explore.amy_selections()` that allows users to observe the settings in the AMY panel. Specifically, it returns the AverageMeteorologicalYear object generated in `app.explore.amy()`. I also renamed a few of the params in `tmy.py` to be more descriptive for the users' benefit. 

Workflow I recommend to get info from the amy panel: 
```
app.explore.amy()                                          # Pull up the panel
amy_selections = app.explore.amy_selections()              # Get AverageMeteorologicalYear object generated by amy panel
data_type = amy_selections.data_type                       # Absolute or Difference
computation_method = amy_selections.computation_method     # Warming Level Future or Historical
warmlevel = amy_selections.warmlevel                       # Get warming level selection
```

If you call `app.explore.amy_selections()` **before** calling `app.explore.amy()`, *None* is returned by the function and the following warning is printed: 
AMY error

### DIFF
--- a/climakitae/selectors.py
+++ b/climakitae/selectors.py
@@ -10,6 +10,7 @@ import cartopy.feature as cfeature
 import geopandas as gpd
 import pandas as pd
 import pkg_resources
+import warnings
 from .unit_conversions import _get_unit_conversion_options
 from .catalog_convert import (
     _resolution_to_gridlabel,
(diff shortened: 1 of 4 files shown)


vicford (approved):
PHENOMENAL - thank you @nicolejkeeney for tackling this! Exactly what is needed

Also tested the calling of amy_selections before app.explore.amy() and that also worked

climakitae/selectors.py
```diff
@@ -528,7 +582,7 @@ def _update_cmap_and_extended_description(self):
         self.colormap = var_info.colormap.item()
         self.extended_description = var_info.extended_description.item()
         self.variable_id = var_info.variable_id.item()
-
+
```
elehmer, line 585: Unneeded whitespace

elehmer (approved):
Just one whitespace issue. Looks good and runs fine on the HUB!

PR #129, opened 2022-11-17
Remove precip derived variable. Link code to deaccumulated precip instead. 

nicolejkeeney:
This PR removes the ability for climakitae to compute total precipitation as a derived variable. Instead, the code now links to deaccumulated total precipitation, which is now available as a new variable in our catalog. 

I am still figuring out a good extended description for this variable, but the rest of the code is ready for review. 

**To test:** Just try retrieving the "Precipitation (total)" variable, calling `app.load()` on it, maybe trying the unit conversions to inches. Thanks!


vicford:
Getting a `UnboundLocalError: local variable 'da' referenced before assignment` error on app.retrieve() with the following selections:
- area averaged monthly total precip, **historical + ssp3.7**, 45km, in mm for CA
- area averaged daily total precip, **historical + ssp3.7**, 45km, in inches for CA

A little testing shows that:
- not being caused by area averaging
- converting units is fine
- historical alone is fine
- ssp3-7.0 alone is fine
- area averaging is fine

So looks like **appending historical** to a future scenario is breaking here for total precip, because interestingly enough, calling both historical and ssp3-7 but not appending works just fine

vicford (changes requested):
So looks like appending historical to a future scenario is breaking here for total precip, because interestingly enough, calling both historical and ssp3-7 but not appending works just fine

nicolejkeeney:
> So looks like appending historical to a future scenario is breaking here for total precip, because interestingly enough, calling both historical and ssp3-7 but not appending works just fine

Good find. This should be fixed now :D

elehmer:
I am not able to load this branch on the HUB. There are 4 references to `variable_catalog.csv` still in the code, but it has been renamed to `variable_description.csv`

nicolejkeeney:
> I am not able to load this branch on the HUB. There are 4 references to `variable_catalog.csv` still in the code, but it has been renamed to `variable_description.csv`

great catch. thanks, eric!

elehmer (approved):
Looks good, but at some point I'd like to get rid of all the `assert` statements in the code (doesn't need to be right now).

PR #131, opened 2022-11-18
Set climakitae version dynamically in metadata_update

elehmer:
This simple PR removes the hard coded `climakitae` version number from `metadata_update`. It is now set to whatever the version number is in `setup.cfg`.


nicolejkeeney (commented):
Fantastic. Thanks, Eric!

PR #133, opened 2022-11-28
Add Black lint checking to workflow

elehmer:
After reviewing the GitHub workflow documentation this looks like the best way to implement lint checking on the repo. With this workflow it didn't seem necessary for the other tests to be dependent on the lint. Also, this just checks the code and fails if not compliant. This would require the coder to run `black` before committing and pushing to pass the test. Or would we rather have the workflow reformat the code automatically? I'm not super sure how that would be done, but can look into it.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -8,6 +8,11 @@ name: ci
 on: [push, pull_request]
 
 jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: psf/black@stable
   build:
     strategy:
       fail-fast: false


nicolejkeeney (changes requested):
The tests for lint seem to be failing now, and I'm not sure how to interpret the error messages.

nicolejkeeney (approved):
Sorry, disregard my last comment. I realize that the failing tests are the purpose of this PR, haha! Am still concerned that the failed test error messages are hard to interpret.

elehmer:
> Sorry, disregard my last comment. I realize that the failing tests are the purpose of this PR, haha! Am still concerned that the failed test error messages are hard to interpret.

Yep, basically if the lint test fails you forgot to run `black` before committing and pushing. So, we would just need to change our procedure for updating code.

nicolejkeeney:
> > Sorry, disregard my last comment. I realize that the failing tests are the purpose of this PR, haha! Am still concerned that the failed test error messages are hard to interpret.
> 
> Yep, basically if the lint test fails you forgot to run `black` before committing and pushing. So, we would just need to change our procedure for updating code.

Got it, sounds good!

elehmer:
It looks like we might be able to have it auto-commit `black` formatted code to the repo as an action, but it would be more complicated than just checking:

https://github.com/lgeiger/black-action/issues/1

elehmer:
I'm going merge this. From now on in order to pass the `black` test you will have to run your code change through `black` and commit the changes. See the `black` documentation for more information:

https://github.com/psf/black

Basically, though, just add `black` to your environment and run `python -m black {source_file_or_directory}` and commit changes.

PR #134, opened 2022-11-29
Improve low value contrast on branded colormaps

vicford:
Hi all, we were asked for increased contrast on the low end of the colormaps in order to distinguish better between the lowest values in a colormap and the white background on our maps, primarily focusing on the warming levels notebook, but appropriate everywhere. 

The two main colormaps where this is appropriate is for ae_orange and ae_diverging. I did not make modifications to the low value of ae_blue as the blue-white contrast is fairly high, but also can make adjustments to this colormap as well. 

In ae_orange: the lowest value was removed, and a new highest value was added, effectively shifting the entire colormap to more intense oranges
In ae_diverging: the lowest positive value was removed, with the corresponding highest value in ae_orange added to the max tail end, to also shift the entire colormap to more intense oranges

_These changes are most noticeable in the minimum map_
**Old ae_orange**
old_orange

**New ae_orange**
new_orange

_Most noticeable at the edges of California, closest to background/null values_
**Old ae_diverging**
old_diverging

**New ae_diverging**
new_diverging

If you would like to test: just look at the app.explore.warming_levels panel (or app.view on any variable). If there needs to be higher contrast, please let me know


gdicecco (approved):
This color scale looks good to me!

nicolejkeeney (approved):
Looks good!

PR #135, opened 2022-11-30
Remove assert statements from main code and replace with raise error

elehmer:
This simple PR replaces `assert` statements used in main code and replaces them with `raise` statements. The test logic is reversed to do this.

I left the `assert` statements in the test files as it seem appropriate there.

### DIFF
--- a/climakitae/selectors.py
+++ b/climakitae/selectors.py
@@ -721,10 +721,8 @@ def _display_select(selections, location, location_type = "area average"):
     for location_type -- 'area average' is the only choice, until station-based
     data are available.
     """
-    assert location_type in [
-        "area average",
-        "station",
-    ], "Please enter either 'area average' or 'station'."
+    if location_type not in ["area average", "station"]:
+        raise ValueError("Please enter either 'area average' or 'station'.")
 
     # _which_loc_input = {'area average': LocSelectorArea, 'station': LocSelectorPoint}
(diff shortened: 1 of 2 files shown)


climakitae/selectors.py
```diff
@@ -721,10 +721,8 @@ def _display_select(selections, location, location_type = "area average"):
     for location_type -- 'area average' is the only choice, until station-based
     data are available.
     """
-    assert location_type in [
-        "area average",
-        "station",
-    ], "Please enter either 'area average' or 'station'."
+    if location_type not in ["area average", "station"]:
+        raise ValueError("Please enter either 'area average' or 'station'.")
```
nicolejkeeney, line 725: Glad you noted this as I've been meaning to come back to this. We don't actually ever use this code; I think its old placeholder code from many months ago. Should we just delete these lines?

elehmer: Yes! Reducing unneeded code would be great!

nicolejkeeney: Will do!

nicolejkeeney (approved):
I removed the old code. Looks good to me and tested and it runs fine. Can you double check that it works before merging?

elehmer:
Is this ready to merge?

elehmer:
> I removed the old code. Looks good to me and tested and it runs fine. Can you double check that it works before merging?

Works...merging...

PR #136, opened 2022-12-02
Modernizing app.select & adding utilities/forecast zone subset options

nicolejkeeney:
This is a total rework of app.select to make it look shinier! And fancier! I've added the following things: 

1. New spatial subsetting options: (1) Electric Demand Forecast Zones and (2) IOUs & POUs 
2. Data selections and location subsetting are now part of the same card, in different tabs 
3. New warning if you've selected data outside the temporal range of the selected data 
4. Two new cards to the left of the main data selection card: (1) How to use this panel and (2) Current selections 
5. The time slice is now highlighted in **grey**. This is the reverse of the previous behavior, when the selected time slice was shown as **white** and the not selected time slice was shown as grey. For me personally, this is more clear. 
6. Area average option is now grouped with location subsetting. I updated the wording to be yes/no and made it radio buttons instead, in an effort to make it stick out more to the users. 
7. New dew point derived variable!! YAY! 

I'm excited to see what everyone thinks! A screenshot of what this new panel looks like is below. 

data selections
location subset

### DIFF
--- a/climakitae/selectors.py
+++ b/climakitae/selectors.py
@@ -27,14 +27,14 @@ unit_options_dict = _get_unit_conversion_options()
 
 # Read in utilties shapefile from package data 
 utitilies_file = pkg_resources.resource_filename(
-    "climakitae", "data/Electric_Load_Serving_Entities_(IOU_%26_POU)"
+    "climakitae", "data/Electric_Load_Serving_Entities_(IOU_%26_POU).parquet"
 )
 utilities = gpd.read_parquet(utitilies_file)
 utilties = utilities[["OBJECTID","Acronym","Utility","Type","geometry"]]
 
 # Read in forecast zones 
 forecast_zones_file = pkg_resources.resource_filename(
-    "climakitae", "data/California_Electricity_Demand_Forecast_Zones"
+    "climakitae", "data/California_Electricity_Demand_Forecast_Zones.parquet"
 )
 forecast_zones = gpd.read_parquet(forecast_zones_file)
 
--- a/climakitae/selectors.py
+++ b/climakitae/selectors.py
@@ -27,18 +27,16 @@ unit_options_dict = _get_unit_conversion_options()
 
 # Read in utilties shapefile from package data 
 utitilies_file = pkg_resources.resource_filename(
-    "climakitae", "data/Electric_Load_Serving_Entities_(IOU_%26_POU).zip"
+    "climakitae", "data/Electric_Load_Serving_Entities_(IOU_%26_POU)"
 )
-utilities = gpd.read_file(utitilies_file)
-utilities = utilities.to_crs(4326) # Convert geometries to lat/lon 
+utilities = gpd.read_parquet(utitilies_file)
 utilties = utilities[["OBJECTID","Acronym","Utility","Type","geometry"]]
 
 # Read in forecast zones 
 forecast_zones_file = pkg_resources.resource_filename(
-    "climakitae", "data/California_Electricity_Demand_Forecast_Zones.zip"
+    "climakitae", "data/California_Electricity_Demand_Forecast_Zones"
 )
-forecast_zones = gpd.read_file(forecast_zones_file)
-forecast_zones = forecast_zones.to_crs(4326) # Convert geometries to lat/lon 
+forecast_zones = gpd.read_parquet(forecast_zones_file)
 
 # For Forecast Zones named "Other", replace that with the name of the county
 forecast_zones.loc[forecast_zones["FZ_Name"] == "Other", "FZ_Name"] = forecast_zones["FZ_Def"]
--- a/climakitae/selectors.py
+++ b/climakitae/selectors.py
@@ -286,7 +286,7 @@ class LocSelectorArea(param.Parameterized):
             self.longitude[0], self.latitude[0], self.longitude[1], self.latitude[1]
         )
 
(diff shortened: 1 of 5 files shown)


climakitae/selectors.py
```diff
@@ -146,21 +190,36 @@ class LocSelectorArea(param.Parameterized):
...
-    latitude = param.Range(default=(32.5, 42), bounds=(10, 67))
-    longitude = param.Range(default=(-125.5, -114), bounds=(-156.82317, -84.18701))
-    cached_area = param.ObjectSelector(objects=dict())
+    longitude = param.Range(
+        default = default_lon, 
+        bounds = (-156.82317, -84.18701)
+    )
+    _lat_lon_warning = param.String(
+        default = "", 
+        doc = "Warning if user is messing with lat/lon slider, \
```
elehmer, line 207: I think using a line cont. `\` will introduce extra spaces into your text string on the UI side. Maybe try using the ("","") method instead to split this up.

climakitae/selectors.py
```diff
@@ -397,6 +495,12 @@ class DataSelector(param.Parameterized):
         default = ["Historical Climate"], 
         objects = ["Historical Reconstruction (ERA5-WRF)", "Historical Climate"]
     )
+    _area_average_yes_no = param.ObjectSelector(
+        default = "No",
+        objects = ["Yes","No"], 
+        doc = "Used to make the select panel more readable. \
```
elehmer, line 501: Same again here I think...

climakitae/selectors.py
```diff
@@ -559,28 +707,66 @@ def _update_scenarios(self):
...
+        data_warning = ""
+        bad_time_slice_warning = "You've selected a time slice that is outside the temporal range of the \
+            selected data." 
+        # Warning based on data scenario selections
         if ( # Warn user that they cannot have SSP data and ERA5-WRF data 
             (True in ["SSP" in one for one in self.scenario_ssp]) and 
             ("Historical Reconstruction (ERA5-WRF)" in self.scenario_historical)
         ) :
-            self._data_warning = "Historical Reconstruction (ERA5-WRF) data is not available with SSP data. \
+            data_warning = "Historical Reconstruction (ERA5-WRF) data is not available with SSP data. \
```
elehmer, line 721: And here lol

gdicecco:
Could we tweak the language with the area average selection (which I think is a much-needed clarification) to say something like "Compute an area average across grid cells within your selected region"? Just trying to make it clear to users that they are collapsing the space dimension if they pick this.

Tianchi-Liu:
To set users up for area average selection, explaining that model resolution refers to the grid cell size / the spatial resolution might also help. Loved the two cards on the right!

bethem:
I think it's very snazzy!

elehmer:
I noticed that when you have all dataset checkboxes selected (I know not ideal or sensible) there is no padding on the bottom of the "current selections" box:

image

nicolejkeeney:
> Could we tweak the language with the area average selection (which I think is a much-needed clarification) to say something like "Compute an area average across grid cells within your selected region"? Just trying to make it clear to users that they are collapsing the space dimension if they pick this.

Yes, we can definitely change the wording to that!

nicolejkeeney:
> To set users up for area average selection, explaining that model resolution refers to the grid cell size / the spatial resolution might also help. Loved the two cards on the right!

@Tianchi-Liu how do you suggest incorporating that selection into the panel? Am keen to add it if it clarifies things, but not sure where.

nicolejkeeney:
> I noticed that when you have all dataset checkboxes selected (I know not ideal or sensible) there is no padding on the bottom of the "current selections" box:
> 
> image

Fixed, thanks for finding this!

nicolejkeeney:
> 1. Just thinking about a user not quite fully paying attention/moving quickly through the panel (I include myself here somewhat) - is it possible to include the name of the tabs in the "How to use this panel" card, perhaps bolded and or a different color? Just to highlight those tabs even further, that there are now two tabs. Since the panel is so pretty and contains a lot of information, I do think the eye is drawn to fun visual aspects, and could potentially overlook the tab placement really quickly.

Great idea, just implemented this

nicolejkeeney:
@elehmer I think I fixed all the string issues you mentioned!

nicolejkeeney:
> 2. Might be playing devil's advocate here, but I am not quite sure about the switch from white-to-grey on the data selection. To me, "greyed out" means unavailable. OG users also might be confused at first to the switch. A potential solution (which may or may not be doable): what if where the grey vertical column intersects the colored data bars, that the colored bar itself also gets "greyed out" but still visible. Meaning, that the colored bars only remain colored where the user wants to retain data. Really rough example below (ignore slider bar range - just trying to get grey region right):
>    Screen Shot 2022-12-09 at 10 47 35 AM
This is a really interesting thought. @gdicecco do you have any opinions on this?

gdicecco:
Re the selector plot shading - my two cents is I can’t unsee the mismatch between the shading on the plot image and the time slice slider in the old version, so I’m inclined towards the change in this PR!

vicford:
> Re the selector plot shading - my two cents is I can’t unsee the mismatch between the shading on the plot image and the time slice slider in the old version, so I’m inclined towards the change in this PR!

I think this is a 100% fair point! Having the grey shading line up with the slider bar would be a big plus of the switch!

PR #137, opened 2022-12-02
Timeseries fig updates

nicolejkeeney:
This is a largely formatting PR for the timeseries panel in the timeseries notebook. 
Updates include:
- Setting default reference range to be 1981-2010, matching warming levels baseline period
- Adding useful text blurbs regarding difference between smoothing and resampling
- Fixes y-axis label to be the variable (old label included smoothing parameters when included)
- Updates title to be reflective of transformations (smoothing applied with # of timesteps, extremes with resampling info, etc.)
new

**To test**: play around with options in the timeseries explore panel in **this PR**: https://github.com/cal-adapt/cae-notebooks/pull/28 to see how titles/labels update with user selections.


gdicecco:
I think we need to make the “resampling is required…” text more clear to reflect the methods that are being used (and help point users to appropriate values for their question) - happy to help edit here but I’m not sure what the precise mathematical operation is that’s going on behind the scenes.

Tianchi-Liu:
Titles update as expected with different selections, except for when removing seasonal cycles. The figure takes time to update when selections are changed, so would appreciate a heads-up. Cool tool!

Tianchi-Liu (approved):

PR #142, opened 2023-01-06
Changes related to the DFO notebooks

nicolejkeeney:
This PR pulls some generalized function (such as finding the closest grid cell to a coordinate pair, and computing HDD and CDD) into the climakitae library. I've also added a function that reads and loads catalog data from a config csv file, as demoed in the intro notebook for the DFO. Lastly, I modified the heatmap function in the meteo_yr library so that a default colormap is applied. This update fixes the unit issue we all noticed when running the DFO notebooks on Thur Jan 6. 

**To test:** Just look through the climakitae code changes, and run through the INTRO notebook, available [here](https://github.com/cal-adapt/cae-examples/blob/dfo_notebooks_clean/specific_use_cases/demand_forecast_office/INTRO_annual_consumption_model.ipynb)

### DIFF
--- a/climakitae/core.py
+++ b/climakitae/core.py
@@ -2,11 +2,7 @@ import intake
 from .data_export import _export_to_user
 from .explore import AppExplore
 from .view import _visualize
-from .data_loaders import (
-    _read_from_catalog, 
-    _compute, 
-    _read_data_from_csv
-)
+from .data_loaders import _read_from_catalog, _compute, _read_data_from_csv
 from .selectors import (
     DataSelector,
     _display_select,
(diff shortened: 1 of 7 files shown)


vicford:
Leaving an additional comment that the heatmaps for the meteorological_year notebook default to virdis when it should be ae_orange here instead.

elehmer:
> LGTM, no issues running the entire notebook, no super long waiting times either

I concur. I was able to run through the INTRO, calculate, and met year notebooks no problem except for 2d in met year notebook where the variable names don't match to anything

nicolejkeeney:
> Leaving an additional comment that the heatmaps for the meteorological_year notebook default to virdis when it should be ae_orange here instead.

The defaults will only work for the AMY panel. The heatmap input is a pandas dataframe, which has no information about the variable since it's just a table. I've inputted it programmatically in the AMY panel, which is why the colormap works there. If we want to change the colormap, we should do so in the notebook itself.

nicolejkeeney:
> I concur. I was able to run through the INTRO, calculate, and met year notebooks no problem except for 2d in met year notebook where the variable names don't match to anything

I'll double check the variable name issue now!

nicolejkeeney:
> > LGTM, no issues running the entire notebook, no super long waiting times either
> 
> I concur. I was able to run through the INTRO, calculate, and met year notebooks no problem except for 2d in met year notebook where the variable names don't match to anything

Ah right, sorry I remember you mentioning this in your last review. I'll fix that now. Thanks for pointing it out again!

PR #143, opened 2023-01-09
Fix messed up map offset for location.view

nicolejkeeney:
Quick fix to address the map offset issue with the little location subset map. 

**To test:** Please try pulling up the following GUIs to ensure that the location subsetting map (see below) is in a reasonable location for all the panels. Thanks! 

location map


elehmer (approved):
I have also checked this against getting_started, explore_warming, explore_tmy, and threshold_exceedance. The select location filter map is located well in all.

PR #144, opened 2023-01-10
Global warming levels data update

vicford:
This updates the look up GWL look up tables to account for the CMIP6 archive. warming_levels.py was also updated to ensure that the larger table did not break existing code looking specifically at the AE models.


elehmer (approved):
LGTM.

bethem:
worked just fine for me!

PR #147, opened 2023-01-19
Bug fix in app.retrieve_from_csv

nicolejkeeney:
This fixes a bug I found in `app.retrieve_from_csv()`, where a mysterious pandas error was being raised if any of the cells in a csv were unfilled. `pd.read_csv` was automatically setting these empty cells to NaN, which was causing issues down the line. Here, I set the empty cells to an empty string instead of NaN, which solves the issue! 

**How to review**: Just review the code change. No need to test anything out I think.

### DIFF
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -468,7 +468,7 @@ def _read_data_from_csv(selections, location, cat, csv, merge=True):
     """
 
     df = pd.read_csv(csv)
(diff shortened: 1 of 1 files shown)


vicford (approved):
Without testing anything, looks straightfoward to me!

elehmer:
That should do the trick!

elehmer (approved):
LGTM!

PR #157, opened 2023-01-27
Add uncertainty utils functions/classes to climakitae API in readthedocs

nicolejkeeney:
I added the uncertainy utils to the climakitae API. I also added cmip6_preprocessing (a dependency of uncertain_utils) to our environment.yml file. 

If you want to test locally and build the docs, follow the instructions [here](https://github.com/cal-adapt/climakitae/blob/main/docs/README.md). **You'll also need to install cmip6_preprocessing into your local climakitae-tests conda environment:** ``pip install cmip6_preprocessing``

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -20,6 +20,7 @@ dependencies:
   - pandas=1.3.4
   - panel=0.12.4
   - param=1.11.1
+  - pint-xarray
   - proj=8.1.1
   - psutil=5.8.0
   - pyproj=3.2.1
--- a/environment.yml
+++ b/environment.yml
@@ -5,6 +5,7 @@ channels:
 dependencies:
   - python=3.9.7
   - cartopy=0.20.1
+  - cmip6_preprocessing
   - dask-gateway=0.9.0
   - geopandas=0.10.2
   - geoviews=1.9.2
--- a/environment.yml
+++ b/environment.yml
@@ -5,7 +5,7 @@ channels:
 dependencies:
   - python=3.9.7
   - cartopy=0.20.1
-  - cmip6_preprocessing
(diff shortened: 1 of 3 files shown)


vicford (approved):
Fantastic, looks good to me @nicolejkeeney! Off-handedly wondering if the grab_temp_data function needs to be more clearly documented as to what it is doing, in regards to temperature data....

environment.yml
```diff
@@ -5,6 +5,7 @@ channels:
 dependencies:
   - python=3.9.7
   - cartopy=0.20.1
+  - cmip6_preprocessing
```
bkg, line 8: Please pin `cmip6_preprocessing` to the known and tested version from the image we're using.

environment.yml
```diff
@@ -19,6 +20,7 @@ dependencies:
   - pandas=1.3.4
   - panel=0.12.4
   - param=1.11.1
+  - pint-xarray
```
bkg, line 23: Is pint-xarray a direct dependency? I'm not seeing it imported anywhere. If it is, then the same comment applies about version pinning, otherwise it should be removed.

bkg (changes requested):

environment.yml
nicolejkeeney: I will double check though

environment.yml
nicolejkeeney: Looks like its version in the hub is 0.0.0, I will add that in now!

environment.yml
nicolejkeeney: You're right, I can delete it! Thanks for the sanity check.

bkg (approved):

PR #160, opened 2023-02-03
Fix map

nicolejkeeney:
The interactivity of the cute little map of the area subset (i.e. the map that should update with your location selection) was broken (by me) several weeks ago, without realizing. I fixed it here! 

As a reminder, this is the map I'm referencing: 
map

**To test**: Pull up the following three panels -- app.select(), app.explore.warming_levels(), and app.explore.amy() -- and ensure that the map is functional, updates with different location selections, and an appropriate size.


vicford (approved):
app.select, app.explore.warming_levels, and app.explore_amy all work with new map

bethem (approved):
all three panels worked for me. very nice!

PR #161, opened 2023-02-03
Colorblind-friendly categorical colormap for line plots

vicford:
Adds a colorblind friendly set of colors exclusively for lineplots, based on: https://gist.github.com/thriveth/8560036

**To test**:
Use app.view on area-averaged data. Instead of a vibrant blue line, a single simulation will plot in a darker blue. 

Alternatively, if using hvplot.line in a cell on area-averaged data:
```
cmap = _read_ae_colormap(cmap="categorical_cb", cmap_hex=True)
data.hvplot.line(x="time", by="simulation", color=cmap)
```

Old color-set for line plots
Screen Shot 2023-02-03 at 12 31 56 PM

Color-blind color-set for line plots (will be our default set of colors for line plots)
Screen Shot 2023-02-03 at 12 30 44 PM


bethem:
looks good to me!

nicolejkeeney (changes requested):
The colormap looks fantastic!! Totally love it. The way you've coded it up is a bit confusing though by adding the `color` argument, so now the view function has both `cmap` and `color`, and the user has to intuit which to use to set the colormap. I hope I didn't step on your toes here but I just went ahead and patched this up a bit in the code and removed the `color` argument since I wasn't sure how to explain my suggestion in text. Hopefully you don't mind, I don't mean to overstep! Let me know what you think about the changes.

vicford:
> The way you've coded it up is a bit confusing though by adding the `color` argument, so now the view function has both `cmap` and `color`, and the user has to intuit which to use to set the colormap.

Thanks for updating this @nicolejkeeney, I appreciate the tidying up of the options. I did revise the cmap options further for timeseries, in that if the user passes a colormap option for timeseries data that regardless of choice it is set to the categorical_cb colormap in order to ensure accessibility. Our other colormaps should never be used for line graphs, as they're not visually accessible and next to impossible to distinguish (example if ae_orange was the cmap option passed).
Screen Shot 2023-02-06 at 9 59 25 AM

nicolejkeeney:
>  I did revise the cmap options further for timeseries, in that if the user passes a colormap option for timeseries data that regardless of choice it is set to the categorical_cb colormap in order to ensure accessibility. Our other colormaps should never be used for line graphs, as they're not visually accessible and next to impossible to distinguish

For sure, I see your point. Since you don't want the user to use any of the other colormaps, I'd recommend removing the redundant code and having this instead: 
```
# Default colormap for timeseries data
if cmap is None:
    cmap = "categorical_cb"
    cmap = _read_ae_colormap(cmap=cmap, cmap_hex=True)
```

Also, the nice thing about the code setup now is that the user can pass a single color as `cmap` and customize the line plots easily! Maybe you knew this already, though? For example, setting `cmap="blue"`: 
blue

nicolejkeeney:
Also maybe you noticed this too already, but I added "categorical_cb" as an option for the gridded data as well-- look how neat it looks in a map!! 
categorical

nicolejkeeney (approved):
I left one comment regarding deleting some redundant code, but other than that it looks great to me!
corgi

PR #162, opened 2023-02-03
Add config argument to retrieve

nicolejkeeney:
This PR combines app.retrieve() and app.retrieve_from_csv(). It allows the user to retrieve the data from a configuration csv file using app.retrieve() by setting the argument `config` to the filepath of a file, i.e. "config.csv". For example: 

This will work to retrieve the data as normal: whatever settings are contained in app.location and app.selections (as set in app.select() ) 
```
app.select() # Make your selections 
data = app.retrieve() 
```

This will also work, duplicating the behavior of app.retrieve_from_csv()
```
app.retrieve(config="config.csv")
```

This will also work, **but eventually I will delete this function**. I need to replace it in the existing notebooks though, first. 
```
app.retrieve_from_csv(config="config.csv") 
```

**To test**: Just try running the example code above. I've included an example csv file here: 
[config.csv](https://github.com/cal-adapt/climakitae/files/10583248/config.csv)

### DIFF
--- a/climakitae/core.py
+++ b/climakitae/core.py
@@ -134,22 +134,22 @@ class Application(object):
     # === Retrieve ===================================
     def retrieve(self, config=None, merge=True):
         """Retrieve data from catalog
-        
-        By default, Application.selections and Application.location determines the data retrieved. 
-        To retrieve data using the settings in a configuration csv file, set config to the local 
+
+        By default, Application.selections and Application.location determines the data retrieved.
+        To retrieve data using the settings in a configuration csv file, set config to the local
         filepath of the csv.
         Grabs the data from the AWS S3 bucket, returns lazily loaded dask array.
-        User-facing function that provides a wrapper for _read_from_catalog and _read_data_from_csv. 
-        
+        User-facing function that provides a wrapper for _read_from_catalog and _read_data_from_csv.
+
         Parameters
         ----------
-        config: str, optional. 
+        config: str, optional.
             Local filepath to configuration csv file
             Default to None-- retrieve settings in app.selections and app.location
         merge: bool, optional
             If config is TRUE and multiple datasets desired, merge to form a single object?
             Defaults to True.
-        
+
         Returns
         -------
         xr.DataArray
--- a/climakitae/core.py
+++ b/climakitae/core.py
@@ -132,19 +132,35 @@ class Application(object):
         return _compute(data)
 
     # === Retrieve ===================================
-    def retrieve(self):
+    def retrieve(self, config=None, merge=True):
         """Retrieve data from catalog
-
-        Applications.selections and Applications.location determine data retrieves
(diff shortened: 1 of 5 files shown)


climakitae/core.py
```diff
@@ -132,19 +132,39 @@ def load(self, data):
...
+        config: str, optional. 
+            Local filepath to configuration csv file
+            Default to None-- retrieve settings in app.selections and app.location
+        merge: bool, optional
+            If config is TRUE and multiple datasets desired, merge to form a single object?
+            Defaults to True.
+        
         Returns
         -------
         xr.DataArray
```
thenaomig, line 155: doesn't it return a Dataset instead of a DataArray if merge=True?

climakitae/core.py
```diff
@@ -132,19 +132,39 @@ def load(self, data):
...
 
         """
+        if config is not None:
+            if type(config) == str:
+                return _read_data_from_csv(
+                    self.selections, self.location, self._cat, config, merge
+                )
+            else:
+                raise ValueError(
+                    "To retrieve data from a configuration file, please input the path to your local configuration csv as a string"
```
thenaomig, line 166: "To retrieve data specified in a configuration file..."

climakitae/core.py
```diff
@@ -132,19 +132,39 @@ def load(self, data):
...
+                )
         return _read_from_catalog(self.selections, self.location, self._cat)
 
     def retrieve_from_csv(self, csv, merge=True):
```
thenaomig, line 170: can make this _retrieve_from_csv since the user will no longer ever call it directly

climakitae/core.py
```diff
@@ -132,19 +132,39 @@ def load(self, data):
...
+        
         Returns
         -------
         xr.DataArray
             Lazily loaded dask array
 
         """
+        if config is not None:
+            if type(config) == str:
+                return _read_data_from_csv(
```
thenaomig, line 161: Wishing there were symmetry between this function name and _read_from_catalog... maybe rename them:
_read_catalog_from_select
_read_catalog_from_csv
?

thenaomig (commented):
Awesome. I didn't try running it -- just left some comments on naming and doc-strings.

climakitae/core.py
nicolejkeeney: I left it in core since it's used as app.retrieve_from_csv() in several of our existing DFU notebooks. Will need to replace those calls with app.retrieve(config="blahblah.cvs") in a PR to cae-notebooks, but want to merge the DFU notebooks from Thur to main first.

climakitae/core.py
nicolejkeeney: Yeah, you're right. Forgot to copy over the documentation. Good catch!

climakitae/core.py
nicolejkeeney: yep great idea.

elehmer (approved):
Tested and works great!

PR #165, opened 2023-02-09
Fix bug in retrieve_from_csv

nicolejkeeney:
core.py was importing the wrong function name. Not sure how I missed this! Hopefully this solves the issue.


thenaomig (approved):
Looks good.

PR #166, opened 2023-02-10
Autodoc

nicolejkeeney:
This PR serves three purposes: 
1) Update the docstrings on all user-facing functions and classes 
2) Rework the readthedocs such that all documentation is automated 
3) Rename functions and classes depending on whether or not they are user facing or backend (with backend functions and classes named with an underscore in front)

I also ended up deleting the visualize module (visualize.py) and just moving the two functions in there into threshold_tools.py, since I thought it was weird we had both view.py and visualize.py. The functions in visualize.py make more sense to be in threshold_tools, in my opinion. 

**To test**: Build the docs (instructions for how to do so [here](https://github.com/cal-adapt/climakitae/blob/main/docs/README.md)). Check them out and let me know what you think about the new docstrings.


climakitae/meteo_yr.py
```diff
@@ -1,12 +1,7 @@
...
-Engine using a standard climatological period (1981-2010) for the historical
-baseline, and uses a 30-year window around when a designated warming level is
-exceeded for the SSP3-7.0 future scenario for 1.5°C, 2°C, and 3°C. Additional
-functionality for 4°C is forthcoming. Working Group 4 (Aug 31, 2022) version
-focuses on air temperature and relative humidity, with all variables being
-available for Analytics Engine Beta Launch. The AMY is comparable to a typical
- meteorological year, but not quite the same full methodology.
+Calculates the Average Meterological Year (AMY) and Severe Meteorological Year (SMY) for the Cal-Adapt: Analytics Engine using a standard climatological period (1981-2010) for the historical baseline, and uses a 30-year window around when a designated warming level is exceeded for the SSP3-7.0 future scenario for 1.5°C, 2°C, and 3°C. 
+
+Additional functionality for 4°C is forthcoming. Working Group 4 (Aug 31, 2022) version focuses on air temperature and relative humidity, with all variables being available for Analytics Engine Beta Launch. The AMY is comparable to a typical meteorological year, but not quite the same full methodology.
```
vicford, line 4: IMO we should/can remove the "additional functionality for 4°C...... WG4 all variables being available for AE Beta launch" sentences. First sentence has been backburnered, and second is out-of-date now. I'd keep the third sentence about AMY being somewhat comparable to TMY, but that's a conversation for later too.

vicford (approved):
This all looks fine to me @nicolejkeeney. Big thank you for including one-sentence lines about what each script is, I think that is immensely useful to both docs and developers. Just one comment about the meteo_yr script

Also supportive of moving the visualize functions into thresholds as suggested. I am of the opinion that something very similar will happen with the uncertain_utils once the credibility toolkit is more fleshed out.

climakitae/meteo_yr.py
nicolejkeeney: Sounds good, thanks for taking a look at the description. I figured some of it was outdated but wasn't sure what part.

elehmer (approved):
LGTM!

PR #167, opened 2023-02-10
more functions for the uncertainty notebooks

bethem:
- added specific functions for the precip uncertainty notebook
- also generalized a function used previously just for the temperature uncertainty notebook
-- grab_temp_data is now grab_multimodel_data, to reflect that it is for grabbing one ensemble member from all available models.  It also has an option to sort the models alphabetically. This should be the only change which impacts the temperature notebook!


vicford (approved):
Tested in both uncertainty notebooks, looks all fine to me! Transition from `grab_temp_data` to `grab_multimodel_data` works like a charm

elehmer:
Need to be run through `black` for linting...

climakitae/uncertain_utils.py
```diff
@@ -226,9 +229,25 @@ def _drop_member_id(dset_dict):
...
+    NOTE: assumes regular calendar
+    """
+    ds_attrs = ds.attrs
+    days_month = ds.time.dt.days_in_month
+    seconds_month = 86400*days_month
+    ds = ds*seconds_month
+    ds = xr.where(ds>0.1,ds,np.nan)
+    ds.attrs = ds_attrs
+    ds.pr.attrs["units"] = 'mm'
+    return ds
```
nicolejkeeney, line 246: I think this could be simplified to avoid having to reset all the attributes: 
```
def _precip_flux_to_total(ds):
    """
    converts precip flux units 
    (kg m-2 s-1) to total precip
    per month (mm)
    NOTE: assumes regular calendar
    """
    days_month = ds.time.dt.days_in_month
    seconds_month = 86400*days_month
    ds = ds*seconds_month
    ds = ds.where(ds>0.1,np.nan)
    ds.pr.attrs["units"] = 'mm'
    return ds
```

climakitae/uncertain_utils.py
```diff
@@ -238,8 +257,10 @@ def grab_temp_data(copt):
 
     Attributes
     ----------
-    copt: CmipOpt
-        Selections for variable, area_subset, location, area_average, timescale
+    copt: object
+        Selections: variable, area_subset, location, area_average, timescale
```
nicolejkeeney, line 261: The correct object type for the input `copt` is `CmipOpt`, not `object`, because the function requires the input of the `CmipOpt` object. Hopefully that makes sense? So it should be `copt: CmipOpt`

climakitae/uncertain_utils.py
```diff
@@ -325,6 +361,80 @@ def grab_temp_data(copt):
...
+
+    Searches the CMIP6 data catalog for data from models that have specific
+    ensemble member id in the historical and ssp370 runs. Preprocessing includes
+    subsetting for specific location and dropping the member_id for easier
+    analysis.
+
+    Attributes
+    ----------
+    copt: object
+        Selections: variable, area_subset, location, area_average, timescale
```
nicolejkeeney, line 376: Again this should be `copt: CmipOpt`

climakitae/uncertain_utils.py
```diff
@@ -325,6 +361,80 @@ def grab_temp_data(copt):
...
+        Specific models
+
+    Returns
+    -------
+    List
+        Historical (1981-2010) CMIP6 xr.Datasets concatenated along member_id
+        for each specified model
+    List
+        Warming level CMIP6 xr.Datasets w/ multiple member_ids
+        for each specified model
```
nicolejkeeney, line 389: This should be lowercase `list of xr.Dataset` instead of `List`, with the "l" being lowercase

climakitae/uncertain_utils.py
```diff
@@ -325,6 +361,80 @@ def grab_temp_data(copt):
...
+    analysis.
+
+    Attributes
+    ----------
+    copt: object
+        Selections: variable, area_subset, location, area_average, timescale
+    alpha_sort: bool
+        Set to True if sorting model names alphabetically is desired
+    cmip_names: list
+        Specific models
```
nicolejkeeney, line 380: Should be `cmip_names: list of str`

climakitae/uncertain_utils.py
```diff
@@ -420,3 +530,55 @@ def compute_vmin_vmax(da_min, da_max):
...
+    """Performs a Kolmogorov-Smirnov test at all lat, lon points
+
+    Parameters
+    ----------
+    sample1: xr.Dataset
+        first sample for comparison
+    sample2: xr.Dataset
+        sample against which to compare sample1
+    sig_lvl: Float
+        alpha level for statistical significance
```
nicolejkeeney, line 545: Needs to be lowercase `float`, not `Float`, to match autodoc requirements

climakitae/uncertain_utils.py
```diff
@@ -420,3 +530,55 @@ def compute_vmin_vmax(da_min, da_max):
...
+
+    Returns
+    -------
+    pandas.dataframe
```
nicolejkeeney, line 549: Needs to exactly match the object type, including capitalizations. Should be `pd.DataFrame`

climakitae/uncertain_utils.py
```diff
@@ -69,6 +70,8 @@ def _cmip_clip(self, ds):
         to_drop = [v for v in list(ds.data_vars) if v != variable]
         ds = ds.drop_vars(to_drop)
         ds = _clip_region(ds, area_subset, location)
+        if variable in ("pr"):
+            ds = _precip_flux_to_total(ds)
```
nicolejkeeney, line 74: Variable is a string, so I think it would make more sense for it to be `if variable == "pr"`

nicolejkeeney (changes requested):
Requested a few small docstrings changes, since the object types need to be quite exact regarding naming/capitalizations for the autodocs to read it in. Otherwise it looks good!

climakitae/uncertain_utils.py
bethem: lol, I think Eric told me to do it this way?

climakitae/uncertain_utils.py
bethem: in my experience it did lose attributes when I didn't copy them over manually, even if I used xarray's built-in `xarray.set_options(keep_attrs=True`

climakitae/uncertain_utils.py
bethem: @elehmer am I remembering correctly?

elehmer: Umm, I wouldn't advice to do it this way lol Nicole has it right lol

bethem: that must have been for something else then. wild. I'll change it

climakitae/uncertain_utils.py
```diff
@@ -8,6 +8,7 @@
 import warnings
 from .selectors import Boundaries
 from cmip6_preprocessing.preprocessing import rename_cmip6
+from scipy import stats
```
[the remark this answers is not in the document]
bethem: that might be useful

climakitae/uncertain_utils.py
```diff
@@ -420,3 +530,55 @@ def compute_vmin_vmax(da_min, da_max):
...
+
+    Returns
+    -------
+    pandas.dataframe
+        columns are lat, lon, and p_value;
+        only retains spatial points where
+        p_value < sig_lvl
+    """
+        
+    sample1 = sample1.stack(allpoints=["y", "x"]).squeeze().groupby("allpoints")
```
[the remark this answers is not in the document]
bethem: this is run in the precip notebook. and I'll do that, thanks

climakitae/uncertain_utils.py
bethem: wait, what? this has since been changed... please go to the latest commit

bethem: sorry about that!

climakitae/uncertain_utils.py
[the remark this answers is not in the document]
bethem: yeah I *swear* I used to filter out nans. That line must have been lost in the shuffle.

nicolejkeeney (changes requested):
Just had one comment regarding the naming of the plotting functions, otherwise this looks good!

climakitae/uncertain_utils.py
```diff
@@ -568,17 +657,214 @@ def ks_stat_2sample(sample1, sample2):
...
-    p_df = p_df.loc[:,['lon','lat','p_value']]
+    p_df = p_df[["lat", "lon", "p_value"]]
+    p_df = p_df.loc[:, ["lon", "lat", "p_value"]]
     p_df = p_df[p_df["p_value"] < sig_lvl]
-    
+
     return p_df
+
+
+def get_warm_level(warm_level, ds, multi_ens=False, ipcc=True):
```
vicford, line 678: Getting the following error in both uncertainty notebooks when trying to run get_warm_level _by itself:_ 
`'int' object is not iterable
`

vicford: Additionally, at least in the model uncertainty notebook, when passing a warm_level and a ds (with time, x, y, simulation dims) it returns nothing and doesn't print anything out, and this does not seem to be the desired behavior. Testing over in the precip notebook, I can test what it outputs, but I'm not sure in the compatibility to the other uncertainty notebook.

bethem: thanks for this. I'll check it out on Monday!

climakitae/uncertain_utils.py
```diff
@@ -420,3 +619,252 @@ def compute_vmin_vmax(da_min, da_max):
...
+                )
+            else:
+                year0 = str(int(year_warmlevel_reached) - 14)
+                year1 = str(int(year_warmlevel_reached) + 15)
+                return ds.sel(time=slice(year0, year1))
+
+
+## -----------------------------------------------------------------------------
+## Plotting and helper functions
+## Currently tuned to precip uncertainty
```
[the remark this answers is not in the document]
bethem: I see. I may have been overzealous in my quest to clean up the cells a bit. I think it makes sense to put those specific functions in the notebook itself. And thanks for letting me know about the naming!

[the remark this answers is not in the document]
nicolejkeeney: Perfect-- sounds good! Let me know if you need help with anything.

climakitae/uncertain_utils.py
nicolejkeeney: If I remember correctly, the function actually requires a float input, not an int-- i.e. you need to input `warm_level = 4.0` (a float) vs. `warm_level = 4` (an int)

climakitae/uncertain_utils.py
```diff
@@ -239,7 +311,9 @@ def grab_temp_data(copt):
     Attributes
     ----------
```
nicolejkeeney, line 312: This should be "Parameters" not "Attributes"

nicolejkeeney (changes requested):
Just one comment about the docstrings.

climakitae/uncertain_utils.py
nicolejkeeney: For some reason this has a weirdly large effect on how they are displayed in the readthedocs page haha

bethem: got it, thanks!

vicford (approved):
All good in the temperature notebook too!

climakitae/uncertain_utils.py
elehmer: Should we handle `int` as well as `float` inputs?

climakitae/uncertain_utils.py
```diff
@@ -420,3 +617,135 @@ def compute_vmin_vmax(da_min, da_max):
...
+    ipcc: bool, default=True
+        Set to False if performing warming level analysis with
+        respect to IPCC standard baseline (1850-1900)
+
+    Returns
+    -------
+    xr.Dataset
+        Subset of projected data -14/+15 years from warming level threshold
+    """
+    wtype = type(warm_level)
```
elehmer, line 698: This can replace line 698...704

```
try:
    warm_level = float(warm_level)
except ValueError:
    raise Exception("Please specify warming level as an integer or float.")
```

the `float` function will throw an error for us if it can't convert or is out of range. We are just intercepting the `ValueError` exception and adding our own explanation.

bethem: ah thanks! fixing now

PR #168, opened 2023-02-10
Simplify repeated code in threshold_tools

ckingdon95:
This PR creates a new function `_get_return_variable` that consolidates code that was repeated between the `get_return_value`, `get_return_prob`, and `get_return_period` functions (I think this will make things clearer/easier to keep track of as we change things that affect all three user-facing functions!)

### DIFF
--- a/climakitae/threshold_tools.py
+++ b/climakitae/threshold_tools.py
@@ -241,7 +241,7 @@ def _calculate_return(fitted_distr, data_variable, arg_value):
     float
     """
 
-    try: 
+    try:
         if data_variable == "return_value":
             return_event = 1.0 - (1.0 / arg_value)
             return_value = fitted_distr.ppf(return_event)
@@ -295,9 +295,7 @@ def _bootstrap(ams, distr="gev", data_variable="return_value", arg_value=10):
     try:
         parameters, fitted_distr = _get_fitted_distr(new_ams, distr, distr_func)
         result = _calculate_return(
-            fitted_distr=fitted_distr,
-            data_variable=data_variable,
-            arg_value=arg_value,
+            fitted_distr=fitted_distr, data_variable=data_variable, arg_value=arg_value,
         )
     except (ValueError, ZeroDivisionError):
         result = np.nan
@@ -337,12 +335,7 @@ def _conf_int(
     bootstrap_values = []
 
     for _ in range(bootstrap_runs):
-        result = _bootstrap(
-            ams,
-            distr,
-            data_variable,
-            arg_value,
-        )
+        result = _bootstrap(ams, distr, data_variable, arg_value,)
         bootstrap_values.append(result)
 
     conf_int_array = np.percentile(
@@ -352,6 +345,7 @@ def _conf_int(
     conf_int_lower_limit = conf_int_array[0]
     conf_int_upper_limit = conf_int_array[1]
     return conf_int_lower_limit, conf_int_upper_limit
+
 
 def _get_return_variable(
     ams,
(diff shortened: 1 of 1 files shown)


climakitae/threshold_tools.py
```diff
@@ -400,77 +405,90 @@ def get_return_value(
...
+    Returns dataset with return values and confidence intervals from maximum series.
 
     Parameters
     ----------
     ams: xarray.DataArray
-    threshold: float
+    return_period: float
     distr: str
     bootstrap_runs: int
     conf_int_lower_bound: float
```
nicolejkeeney, line 494: Can you add a brief description of what each input is? Just for user-facing functions I think it's nice to have a bit more info about what each argument is beyond just the object type. For example, something like this: 
```
Parameters
------------
ams: xarray.DataArray
   Annual maximum data, outputted by the function get_ams()
```

climakitae/threshold_tools.py
```diff
@@ -400,77 +405,90 @@ def get_return_value(
...
+def get_return_value(
     ams,
-    threshold,
+    return_period=10,
     distr="gev",
     bootstrap_runs=100,
     conf_int_lower_bound=2.5,
     conf_int_upper_bound=97.5,
     multiple_points=True,
 ):
```
nicolejkeeney, line 483: I don't really see the purpose of this function since it's essentially just a copy of `_get_return_variable`, wrapped in another function with a different name. Why not just rename `_get_return_variable` to `get_return_value`, and have that user-facing function called in `get_return_prob` and `get_return_period`?

nicolejkeeney (changes requested):
Looks good to me, I left a few minor comments about added more description to the docstring of `get_return_value` since it's user facing. I also had a question/comment about that function as well (see review comments)

climakitae/threshold_tools.py
elehmer: These are the main functions called to run the code and have specific parameters needed for each. They now call a generaic internal function `_get_return_variable`  to reduce redundant code.

elehmer (approved):
This looks quite good! Thanks for tackling this as I saw it needed it but didn't want to break it either lol I tested it on the HUB and works!

climakitae/threshold_tools.py
nicolejkeeney: I see your point, but `get_return_value` and `_get_return_variable` have the exact same argument defaults. If they had different default arguments I could maybe see having the two, but as they are now, I still think it makes more sens to just combine them into one. Any thoughts @ckingdon95 ? Definitely open to being wrong about this, haha

nicolejkeeney: Maybe this is a best practice python thing that I'm not aware of? I.e. is it bad practice to use one user-facing function in another user-facing function?

elehmer: `get_return_value`, `get_return_prob`, `get_return_period` are the function exposed to the user and have the same inputs except for the variable to be returned but different names and docstrings. This is a better way to present to the user rather than one function that does different things (that is what happens behind the scenes). This is a good compression of redundant code done here as it moves that code to a single internal function. Does this makes  sense?

ckingdon95: Yeah agree with Eric here. I think my discomfort with the other suggestion is also that 'get_return_value' wouldn't be a good name for the shared function, because it actually can do more than that. So I also think it's important for the function names to make sense for what they do (with `_get_return_variable` being the generic one that can do all three)

climakitae/threshold_tools.py
ckingdon95: yes will do!

climakitae/threshold_tools.py
```diff
@@ -400,77 +405,90 @@ def get_return_value(
...
-    """Creates xarray Dataset with return probabilities and confidence intervals from maximum series
+    """Creates xarray Dataset with return values and confidence intervals from maximum series
 
-    Returns dataset with return probabilities and confidence intervals from maximum series.
+    Returns dataset with return values and confidence intervals from maximum series.
```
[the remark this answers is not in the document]
ckingdon95: yes thank you!

climakitae/threshold_tools.py
nicolejkeeney: Sure, sounds good to me. Makes sense now, sorry for my initial hesitation. Thanks for the explanation!

elehmer:
Lint is failing. @ckingdon95 You might need to update your black version: `pip install black==23.1.0 --upgrade`

nicolejkeeney (approved):
Looks good to me! Thanks for taking the time to fix the docstrings. I just ran this code through black on my end and now it's passing.

PR #169, opened 2023-02-15
Add BSD3 license

nicolejkeeney:
As discussed, this PR adds a license to our repo.


license
```diff
@@ -0,0 +1,28 @@
+BSD 3-Clause License
```
bkg, line 1: Remove this line and the following blank line as they aren't part of the license.

bkg, line 1: Please rename the `license` file to `LICENSE` which is the standard name.

bkg (commented):
Thanks for adding this.

license
nicolejkeeney: I'd be keen to keep this in since it's the default from GitHub, and its the only indicator in the file of what type of license it is.

bkg: It's not part of the [licensing language](https://opensource.org/license/bsd-3-clause/) and GH will automatically interpret the license type and display that info, see an example at <https://github.com/django/django/blob/main/LICENSE>.

nicolejkeeney: Sounds good. I'm curious why that info isn't appearing in our repository after I deleted that line? https://github.com/cal-adapt/climakitae/blob/license/LICENSE

bkg: Looks like it's missing this line: https://github.com/django/django/blob/main/LICENSE#L2

And the formatting is slightly different, but I'm not sure if that's a problem here.

nicolejkeeney: Argh, I just tried literally copy-pasting the license from the link to django's repo-- no luck getting the details to update! The language is a bit different too. I don't know how much this matters. My preference is to use the default language from GitHub, but don't really care too much. What is your preference?

bkg: I'm sure that's fine. Feel free to roll with the GH version, sorry for the complication, ideally they would be the same. It might not be displaying properly until it's merged with main.

PR #171, opened 2023-02-15
Update docs

nicolejkeeney:
The docs work when you build them locally, but I needed to add in some files that had previously been gitignored for it to build on our readthedocs page. I also updated a bad link. 

**to test:** Just look over the code. No way to test this really (at least, that I know of!) until it is merged to main.


.gitignore
```diff
@@ -148,5 +145,3 @@ cython_debug/
 .local
 *.DS_Store
 
-# Generated html files from docs 
-docs/generated/*
```
nicolejkeeney, line 152: For reference: I had been generating some files from a template before everything was automated, following the pattern set in the xarray docs. I ended up deleting this, so this line in the gitignore is no longer neccessary.

elehmer (approved):
Looks fine.

PR #172, opened 2023-02-20
Update app.select

nicolejkeeney:
Redesign app.select(). Add the framework for adding LOCA data down the line. 

**To test:** Try out the new panel and let me know what you think! Try retrieving data as well to make sure everything runs smoothly.

app select


vicford (approved):
Love it! This is literally so minor, it 1% bothers me that the three sub-panels are not the same length. If the area averaging was on the left side of the middle panel, would that even it out?

elehmer (approved):
Wow! This looks and runs great! So much faster!

elehmer:
> Love it! This is literally so minor, it 1% bothers me that the three sub-panels are not the same length. If the area averaging was on the left side of the middle panel, would that even it out?

Yeah, it might look a bit better that way but super minor! Works great and is clear!

nicolejkeeney:
> Love it! This is literally so minor, it 1% bothers me that the three sub-panels are not the same length. If the area averaging was on the left side of the middle panel, would that even it out?

I think it looks a bit unbalanced with the map all alone? What do you think? 
app_select_moved

PR #173, opened 2023-02-27
Remove hard coding of leap year in meteo year code

nicolejkeeney:
Remove the hard coding in the code of setting days_in_year based on whether or not the data is a leap year. Instead, determine whether or not the data has leap days using the xarray datetime accessor [DataArray.dt.is_leap_year](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.dt.is_leap_year.html). This is just a backend change-- if all works well, you shouldn't notice any front end changes. 

This code change should set us up to seamlessly integrate the new data that Brian is working on to remove leap days. 

**To test**: Run app.explore.amy(). Make sure it runs and updates without error.

### DIFF
--- a/climakitae/meteo_yr.py
+++ b/climakitae/meteo_yr.py
@@ -203,7 +203,7 @@ def _format_meteo_yr_df(df):
     return df
 
 
-def compute_amy(data, days_in_year=366, show_pbar=False):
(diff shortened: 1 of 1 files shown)


vicford:
Works just fine at present, but will update + approve once pangeo image is also switched over!

PR #174, opened 2023-02-27
For updated image

thenaomig:
Merge this so that climakitae works with the updated image on the hub. Makes some small changes to account for new versions of a few packages.


elehmer:
I added `intake-esm` to requirements.txt and environment.yml as it is now needed.

thenaomig:
Need to update requirements.txt and environment.yml to match the version numbers here: https://github.com/pangeo-data/pangeo-docker-images/blob/2023.02.08/pangeo-notebook/conda-linux-64.lock

thenaomig:
Someone will need to double-check my version numbers because now tests are failing...

elehmer:
> Someone will need to double-check my version numbers because now tests are failing...

On it...

requirements.txt
```diff
@@ -1,23 +1,24 @@
...
-scipy==1.7.1
-shapely==1.7.1
+xarray==2023.2.0
+cartopy==0.21.1
+dask-gateway==2023.1.1
+geopandas==0.12.2
+geoviews==1.9.6
+holoviews==1.15.4
+hvplot==0.8.2
+intake==0.4.0
```
elehmer, line 8: `intake==0.6.6`

requirements.txt
```diff
@@ -1,23 +1,24 @@
...
+intake-xarray==0.6.1
+intake-esm==2022.9.18
```
elehmer, line 10: Add `intake-geopandas==0.4.0`?

environment.yml
```diff
@@ -4,28 +4,32 @@ channels:
...
-  - shapely=1.7.1
-  - xarray=0.19.0
+  - cartopy=0.21.1
+  - dask-gateway=2023.1.1
+  - dask-geopandas
+  - geopandas=0.12.2
+  - geoviews=1.9.6
+  - holoviews=1.15.4
+  - hvplot=0.8.2
+  - intake=0.4.0
```
elehmer, line 14: `intake==0.6.6`

thenaomig:
Thanks for finding my intake version error, @elehmer. I'm currently tempted to merge a version of this PR without the pinned version-numbers so that something working is on the hub. And then open a separate PR to pin the versions, after we figure out why this test is failing. Thoughts?

elehmer:
> Thanks for finding my intake version error, @elehmer. I'm currently tempted to merge a version of this PR without the pinned version-numbers so that something working is on the hub. And then open a separate PR to pin the versions, after we figure out why this test is failing. Thoughts?

It is worth a shot. I tried bumping python to 3.10.9 like it is on the HUB but just caused another failure. We are in pyproj/proj hell again. I am not sure how it got built on the HUB?

Makefile
```diff
@@ -1,23 +1,7 @@
-PKGNAME = climakitae
```
bkg, line 1: Looks like you overwrote the `Makefile` by accident, could you roll this back please?

thenaomig: Whoops. Fixed.

PR #178, opened 2023-03-01
Install deps on workers, fixes missing dask-geopandas

bkg:


elehmer:
Oh wow, interesting solution!

elehmer:
Does this mean if we are testing a dev branch using cluster on the hub we will have to modify this as well?

vicford:
This works for me, no errors on initalizing a dask cluster followed by the app = ck.Application call. 

Potentially related, but am now getting the following warning when importing climakitae:
```
/srv/conda/envs/notebook/lib/python3.10/site-packages/geopandas/_compat.py:123: UserWarning: The Shapely GEOS version (3.10.3-CAPI-1.16.1) is incompatible with the GEOS version PyGEOS was compiled with (3.11.1-CAPI-1.17.1). Conversions between both will be slow.
  warnings.warn(
```

bkg:
Thanks for testing this. That message isn't a byproduct of this PR, it was introduced from https://github.com/cal-adapt/climakitae/pull/177 where upgrading xclim, downgrades scipy, and shapely, producing this warning. I'll work on addressing that separately.

bkg:
@elehmer good observation, yes it would unfortunately, so it would be better to think of a cleaner solution in that case than having to add the branch name.

It's probably good to mention that this does fix another problem though, where some climakitae module imports threw import errors as the package itself wasn't installed on the workers.

PR #179, opened 2023-03-01
Removed for loop to speed up AMY.

thenaomig:
Updates in less than a minute on my laptop, with no dask cluster, fwiw.


vicford (approved):
Was finally able to test both the AMY and SMY calculations between main and this branch. Everything checks out and is identical. Also really useful to see how the code is sped up with no loops for my own learning, thanks @thenaomig!

nicolejkeeney (approved):
looks good to me

PR #180, opened 2023-03-02
Select station

nicolejkeeney:
Bias corrected station data added to app.select()


climakitae/selectors.py
```diff
@@ -242,10 +216,17 @@ class _LocSelectorArea(param.Parameterized):
         administrative geographic area]
     """
 
+    info_about_station_data = "When you retrieve the station data, gridded model data will be bias-corrected to that point. This process can start from any model resolution."
+
```
thenaomig, line 220: grid-spacing

climakitae/data_loaders.py
```diff
@@ -449,9 +490,304 @@ def _read_catalog_from_select(selections, location, cat):
...
+    if obs_data.units != new_units:
+        obs_data = obs_data + 273.15  # Convert Celcius to Kelvin
+        obs_data.attrs["units"] = "K"
+        if new_units != "K":
+            obs_data = _convert_units(obs_data, new_units)
+    return obs_data
+
+
+def _retrieve_and_format_closest_gridcell(stations_df, station_name, data):
+    """Retrieve the cloeset gridcell to an input weather station
```
thenaomig, line 734: closest

climakitae/data_loaders.py
```diff
@@ -449,9 +490,304 @@ def _read_catalog_from_select(selections, location, cat):
...
+
+    station_ds = xr.open_mfdataset(
+        filepaths,
+        preprocess=_preprocess_hadisd,
+        engine="zarr",
+        consolidated=False,
+        parallel=True,
+        backend_kwargs=dict(storage_options={"anon": True}),
+    )
+    apply_output = station_ds.apply(
```
thenaomig, line 526: would expect this to require a groupby() over each station

thenaomig:
I'm fine merging this (and leaving the 'if loop...' in there for now, just so we can reassure ourselves it's faster without in the presence of a cluster, but then having another quick PR to clean it up so it's only done with the 'apply').

climakitae/selectors.py
nicolejkeeney: resolved, thanks

climakitae/data_loaders.py
nicolejkeeney: resolved, thanks

PR #181, opened 2023-03-07
Incorporate LOCA data selection and retrieval into climakitae

nicolejkeeney:
This PR builds in the functionality to retrieve LOCA data from the AE bucket using `app.select()` and `app.retrieve()`. 

Please test selection and retrieval for as many cases as you have time/energy for, since I modified a ton of the backend I'm worried I may have introduced bugs in places without realizing (although, I tried to be thorough to avoid this, but it's always possible!). Also try visualizing the data with `app.view()`


vicford:
max temp doesn't return anything in app.view and subsetting for any option doesn't matter here (tried time slice, location, area averaged/not)
Screen Shot 2023-04-11 at 1 02 30 PM
Screen Shot 2023-04-11 at 1 07 37 PM

vicford:
Out of curiosity, why does some of the LOCA data have differing numbers of simulations? Noting that max temp has 14, specific humidity has 11, precip has 14

vicford:
> Precipitation when both LOCA and WRF are selected, a couple of things

Ahah! Ignoring scale here (that's a separate issue), looks like the WRF data has 0s, LOCA has nans
Screen Shot 2023-04-11 at 1 49 22 PM
Screen Shot 2023-04-11 at 1 49 30 PM

vicford:
In terms of scale, precipitation is off by 1e5... like 10 meters to 1 mm? Something funky is going on in terms of a unit conversion here

nicolejkeeney:
> In terms of scale, LOCA precipitation in mm is off by 1e5... like 10 meters to 1 mm? Something funky is going on in terms of a unit conversion here

Thanks for figuring out all the weirdness going on with the LOCA precip, I hadn't noticed that. I've removed it from the options for now and we can re-address it in a separate PR.

nicolejkeeney:
> Specific humidity colormap should also be ae_blue

Yep, you're right. Something was going on with the cmaps, I have just fixed it!

nicolejkeeney:
> Do we want the default variable in the LOCA variable dropdown list to be specific humidity? I would vote for max temp, personally

Agreed, just shifted it to max temp. Thansk!

nicolejkeeney:
> only specific humidity returns a plot in app.view Works: specific humidity Doesn't work: max temp, min temp, precip, N-S winds, E-W winds 

Yeah... something is up with some of the LOCA simulations that are causing issues on the data retrieval end. Thanks for pointing this out. I've restricted the LOCA runs to just a few non-problematic simulations. In another PR I can look more into this issue and add back in the other simulations

climakitae/data_loaders.py
```diff
@@ -133,6 +135,49 @@ def _get_cat_subset(selections, cat):
...
+        wrf_on_native_grid = [
+            institution
+            for institution in cat.df.institution_id.unique()
+            if institution != "UCSD"
+        ]
+        cat_subset = cat_subset.search(institution_id=wrf_on_native_grid)
+
+    # Limit simulations if both LOCA and WRF are selected
+    # LOCA has more simulations than WRF
+    # We just want the simulations that are present in both datasets
```
thenaomig, line 155: I'm not sure about that. I think we'd want all of them returned. Hopefully the data structure will allow for that without too much inefficiency.

climakitae/data_loaders.py
```diff
@@ -205,108 +250,111 @@ def _process_and_concat(selections, location, dsets, cat_subset):
...
+                        ) and (
+                            scenario
+                            in [
+                                "SSP 2-4.5 -- Middle of the Road",
+                                "SSP 5-8.5 -- Burn it All",
+                            ]
+                        ):
+                            filename = fnmatch.filter(
+                                list(dsets.keys()),
+                                "*{0}*CESM2*{1}*".format(
```
thenaomig, line 321: I don't love that this is hard-coded at 'CESM2' -- we'd like this to just work if, in principle, another simulation for one of those SSPs was ever added to the catalog.

climakitae/data_loaders.py
```diff
@@ -358,14 +406,27 @@ def _get_data_one_var(selections, location, cat):
...
-        dset = dset.assign_coords({"simulation": dset.attrs["source_id"]})
+        # Break down name into component parts
+        (
+            activity_id,
+            institution_id,
+            source_id,
+            experiment_id,
+            table_id,
+            grid_label,
+        ) = dname.split(".")
```
thenaomig, line 417: Shouldn't this stuff be in the metadata, or how the catalog is organized, so we don't need to be parsing strings?

climakitae/selectors.py
```diff
@@ -543,45 +514,43 @@ def view(self):
...
+        # if "Statistical" in self.selections.downscaling_method:
+        #     # 3km LOCA grid shown whenever LOCA is selected, even if WRF is also selected
+        #     _add_res_to_ax(
+        #         poly=self._loca_bb["3 km"],
+        #         ax=ax,
+        #         color="purple",
+        #         rotation=32,
+        #         xy=(-127, 40),
+        #         label="3 km",
+        #     )
```
thenaomig, line 526: Better not to merge commented-out work-in-progress code. Stash it elsewhere.

climakitae/selectors.py
```diff
@@ -1244,30 +1361,30 @@ def _location_param_to_panel(location):
...
-    data_type = pn.widgets.RadioBoxGroup.from_param(
-        location.param.data_type, inline=True, name=""
-    )
+    # data_type_text = pn.widgets.StaticText(
+    #     value="",
+    #     name="Data type",
+    # )
+    # data_type = pn.widgets.RadioBoxGroup.from_param(
+    #     location.param.data_type, inline=True, name=""
+    # )
```
thenaomig, line 1370: Again to avoid merging commented-out code, stash elsewhere and delete.

climakitae/view.py
```diff
@@ -70,13 +72,13 @@ def _visualize(data, lat_lon=True, width=None, height=None, cmap=None):
                     timescale = data.frequency
                 cmap = var_catalog[
                     (var_catalog["display_name"] == data.name)
-                    & (var_catalog["timescale"] == timescale)
-                ].colormap.item()
+                    # & (var_catalog["timescale"] == timescale)
+                ].colormap.values[0]
```
thenaomig, line 76: this appears less ideal than what it replaces?

thenaomig:
Two things:
- I seem to recall that CESM2 doesn't have LOCA2 daily min and max temperature, but it *should* have specific humidity, yet I don't see it among the simulations even when I select specific humidity
- General preferred behavior would be to return all available LOCA2 simulations if 'statistical' and 'dynamical' are both selected, even though we only return the WRF ones that are bias-corrected on the LOCA grid.

climakitae/data_loaders.py
nicolejkeeney: This is unrelated to the LOCA data actually. It's only used when retrieving ensemble means (see the code directly above it) for SSP 245 and SSP 585.  I don't think our existing TMY code ever uses those two scenarios, and neither scenario has an ensemble mean... so I'm not sure why this code I added this code in ages ago. The TMY stuff has undergone so many iterations, I will need to check in with Victoria on it's current state. 

It's showing up here as a diff because I changed the way the filenames are found for all cases. I used to use list comprehension but prefer the wildcard matching that fnmatch has, so I switched it for all cases in this code chunk.

nicolejkeeney: I'll make a trello card to look into this

nicolejkeeney: https://trello.com/c/ijwDPxxG/373-cesm2-data-retrieval-for-ssp-245-and-585

climakitae/data_loaders.py
nicolejkeeney: It is in the catalog organization actually. The dname is assigned by intake upon retrieval using information from the catalog. This was just an easy way to access the catalog info without needing to dig into the data attributes, which are inconsistent across LOCA/WRF/etc.

climakitae/data_loaders.py
nicolejkeeney: > * I seem to recall that CESM2 doesn't have LOCA2 daily min and max temperature, but it _should_ have specific humidity, yet I don't see it among the simulations even when I select specific humidity

I noticed this too. I'll look into that.

climakitae/data_loaders.py
thenaomig: Okay, this can also be a separate trello card, but I'd rather future-proof other data being added that doesn't follow the same file naming convention. If the same attribute goes by two different names we can have a special function to handle that perhaps that references some look-up, or harmonize them in the catalog itself.

nicolejkeeney: The LOCA data doesn't have these attributes unfortunately, it only has units and long_name if I remember correctly.

nicolejkeeney: > * I seem to recall that CESM2 doesn't have LOCA2 daily min and max temperature, but it _should_ have specific humidity, yet I don't see it among the simulations even when I select specific humidity

Yep, something is up with the CESM2. I must have introduced a bug somewhere that drops it. I'll look into that now, thanks for pointing it out

climakitae/selectors.py
nicolejkeeney: Good catch. Removed!

elehmer (approved):
Looks good.

nicolejkeeney:
> * 9/1/1980

Thinking on this. Thanks for pointing out!

PR #183, opened 2023-03-09
Fix tests by pip installing xclim in the github workflow instead of using conda

elehmer:


nicolejkeeney (approved):
awesome

thenaomig:
Looks very reasonable to me.

PR #187, opened 2023-03-09
Replace ObjectSelector with Selector

elehmer:
This PR replaces ObjectSelector with Selector for Param widgets.


thenaomig:
This is great. Doesn't seem to have messed up any formatting either, since the look in the panel is specified elsewhere. I am good with merging this.

vicford (approved):
Looks good to me, thanks for handling this Eric

PR #188, opened 2023-03-14
Bug fix for bay area stations

nicolejkeeney:
This PR expands the spatial domain that is retrieved when grabbing the Oakland Airport and/or SFO Airport weather stations. It forces the retrieval of the entire model domain, then subsets, such that the gridcells in the water of the bay are included. Ideally we should modify all shapefiles that have gridcells in the bay area to include the water areas, but this fix is fine for now. 

**To test**: Make sure data retrieval at the OAK and SFO stations work.


Tianchi-Liu:
Tried to retrieve the following data:

Screenshot 2023-03-16 103415



None seems to be retrieved:

Screenshot 2023-03-16 103436
Screenshot 2023-03-16 103552

Tianchi-Liu:
Added data from a future model, and switched from 9km to 3km, then succeeded in retrieving.

Tianchi-Liu:
A side note on an unexpected panel behavior: Historical climate data selected to start in 2010, but the start of the time slider temporarily popped back to 1980 while `app.retrieve()` is running. It then corrects itself.

nicolejkeeney:
> A side note on an unexpected panel behavior: Historical climate data selected to start in 2010, but the start of the time slider temporarily popped back to 1980 while `app.retrieve()` is running. It then corrects itself.

This happens because behind the scenes the bias-correction requires that entire time slice. Probably not desired behavior for the time slider on the user-facing side to update while that's happening behind the scenes, though. Good point.

PR #196, opened 2023-04-04
Downgrade image

nicolejkeeney:
Necessary climakitae changes to revert to pangeo image tag 2021.10.19. I left environment.yml and requirements.txt as is. 

You'll also need to pip install xclim: ```pip install xclim==0.36.0```

**To test:**
Try running all the panels- app.select, app.expore.amy, app.explore.thresholds, app.explore.warming_levels and perhaps some of the timeseries stuff if you're so inclined. I believe @vicford is working on reverting some of the notebooks to work for this image.

### DIFF
--- a/climakitae/selectors.py
+++ b/climakitae/selectors.py
@@ -651,7 +651,7 @@ def _get_simulation_options(cat, activity_id, table_id, grid_label, experiment_i
 
     # Get all unique simulation options from catalog selection
     try:
-        simulation_options = cat_subset.unique()["source_id"]["values"]
+        simulation_options = list(cat_subset.df["source_id"].unique())
         if "ensmean" in simulation_options:
             simulation_options.remove("ensmean")  # Remove ensemble means
     except:
@@ -769,7 +769,7 @@ class _DataSelector(param.Parameterized):
             table_id=_timescale_to_table_id(self.timescale),
             grid_label=_resolution_to_gridlabel(self.resolution),
         )
-        self.unique_variable_ids = self.cat_subset.unique()["variable_id"]["values"]
+        self.unique_variable_ids = list(self.cat_subset.df["variable_id"].unique())
         self.variable_options_df = _get_variable_options_df(
             var_catalog=var_catalog,
             unique_variable_ids=self.unique_variable_ids,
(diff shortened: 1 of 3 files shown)


climakitae/selectors.py
```diff
@@ -651,7 +651,7 @@ def _get_simulation_options(cat, activity_id, table_id, grid_label, experiment_i
 
     # Get all unique simulation options from catalog selection
     try:
-        simulation_options = cat_subset.unique()["source_id"]
+        simulation_options = cat_subset.unique()["source_id"]["values"]
```
bkg, line 654: You could use the `DataFrame` here to get unique vals specific to each column, e.g.
``` python
cat_subset.df['source_id'].unique()
```
This has the advantage of the syntax working on both images, plus it's more efficient calling `.unique()` per column instead of the entire catalog for every column.

climakitae/uncertain_utils.py
```diff
@@ -7,7 +7,7 @@
 import intake
 import warnings
 from .selectors import Boundaries
-from xmip.preprocessing import rename_cmip6
+from cmip6_preprocessing.preprocessing import rename_cmip6
```
bkg, line 10: A common way to deal with changes like this is catching the `ImportError`. This allows for supporting older and newer versions of the package.
``` python
try:
    from xmip.preprocessing import rename_cmip6
except ImportError:
    from cmip6_preprocessing.preprocessing import rename_cmip6
```

setup.cfg
```diff
@@ -17,13 +17,11 @@ classifiers =
 packages = find:
 install_requires =
     cartopy
-    dask-geopandas
```
bkg, line 20: I think this package is still required for either image when working with a dask cluster, so it can be left as is.

setup.cfg
```diff
@@ -17,13 +17,11 @@ classifiers =
...
     geopandas
     geoviews
     holoviews
     hvplot
     intake
-    intake-geopandas
```
bkg, line 26: Same comment as above.

bkg (commented):
Thanks for the PR Nicole!

climakitae/selectors.py
nicolejkeeney: great idea, thanks!

climakitae/uncertain_utils.py
nicolejkeeney: awesome, just updated

PR #202, opened 2023-04-20
Fix bug in station selection

nicolejkeeney:
This fixes a bug in the station selection code by adding back in the grid_mapping attr, which I had unknowingly removed.

**To test**: Ensure that retrieval and loading of weather station data works using ``app.select()`` and ``app.retrieve()``.


vicford (approved):
Works just fine for me!

PR #203, opened 2023-04-25
Add in LOCA2 precip with correct units and unit conversions

nicolejkeeney:
ON HOLD FOR NOW: do not review.

This PR adds in the LOCA2 precipitation data with its correct units and unit conversion options. It also modifies how related unit conversions for other precip variables (which have units of mm) are coded in. 

**To test**: Try pulling the precip variables, for LOCA, WRF, and a combination of the two, with different units. Make sure the units and data all look as expected. **TIP**: When testing with the LOCA data, just retrieve and load a year or two of data for a small spatial region to avoid blowing out the memory.

### DIFF
--- a/climakitae/data/variable_descriptions.csv
+++ b/climakitae/data/variable_descriptions.csv
@@ -53,9 +53,9 @@ sfc_runoff,Dynamical,daily/monthly,mm/d,Surface runoff,"The rate at which water 
 subsfc_runoff,Dynamical,daily/monthly,mm/d,Subsurface runoff,"The rate at which  water that has infiltrated the ground surface travels underground until it reaches a body of water (stream, river, etc.)",ae_blue,FALSE
 prec_max,Dynamical,daily/monthly,mm/h,Maximum precipitation,The maximum precipitation rate in a single hour. ,ae_blue,TRUE
 snow,Dynamical,daily/monthly,mm,Snow water equivalent,The amount of water that would be present if all snow melted. ,ae_blue,FALSE
-pr,Statistical,daily/monthly,mm,Precipitation (total),Total precipitation. Computed by summing total gridscale precipitation and total cumulus precipitation.,ae_blue,FALSE
-tasmax,Statistical,daily/monthly,K,Daily maximum air temperature at 2m,The maximum daily air temperature at 2m above the Earth's surface. ,ae_orange,TRUE
-tasmin,Statistical,daily/monthly,K,Daily minimum air temperature at 2m,The minimum daily air temperature at 2m above the Earth's surface. ,ae_orange,TRUE
-uas,Statistical,daily/monthly,m s-1,West-East component of Wind at 10m,"Wind coming from the west, measured at 10m above Earth's surface. By convention, wind coming from the east is negative.",ae_diverging,TRUE
-vas,Statistical,daily/monthly,m s-1,North-South component of Wind at 10m,"Wind coming from the north, measured at 10m above Earth's surface. By convention, wind coming from the south is negative.",ae_diverging,TRUE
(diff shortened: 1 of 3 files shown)


elehmer:
I am getting this error when I try to use both Dynamical and Statistical for Precipitation (Any units, 2098-2100, Aliso San-Onfre WS)

OverflowError: cannot convert float infinity to integer

nicolejkeeney:
> I am getting this error when I try to use both Dynamical and Statistical for Precipitation (Any units, 2098-2100, Aliso San-Onfre WS)
> 
> OverflowError: cannot convert float infinity to integer

That is a super interesting bug you found, Eric! That error is only raised if you have Historical Climate selected. I don't know why this raises an error, but I will investigate in a separate PR, since it's not related to the precip data (the same error occurs with different variables as well).

vicford:
This is for SSP370, 2056-2061 kg/m2/s for SDGE LOCA only.... the size of the plot is changing? 

https://user-images.githubusercontent.com/106171203/234594556-19ce0a2b-88df-4515-80f0-44b1dd0e0bec.mov

elehmer:
> This is for SSP370, 2056-2061 kg/m2/s for SDGE LOCA only.... the size of the plot is changing?
> 
>  Screen.Recording.2023-04-26.at.9.36.47.AM.mov

Looks like due to the changing legend to scientific notation and back. Can we control how those numbers are formatted?

vicford:
> Looks like due to the changing legend to scientific notation and back. Can we control how those numbers are formatted?

Ah, good catch @elehmer.  I think we should also fix the scale/display of what "all zeros" looks like, and that might handle this as well, potentially

vicford:
> colorbar isn't shifting around

Side note, this could potentially be useful across all our geospatial plots, in case someone is making an animation, visual interpretation, etc.

elehmer:
> > colorbar isn't shifting around
> 
> Side note, this could potentially be useful across all our geospatial plots, in case someone is making an animation, visual interpretation, etc.

Yes, when displaying a timeseries it is important to map the symbology to the data in a fixed manner for each time step, otherwise the steps are not easily visually comparable. This can be a struggle if the data ranges vary greatly within the timesteps.

nicolejkeeney:
> Thanks for making these fixes @nicolejkeeney.
> 
> Retrieving data and units all check out for me, tested by looking at WRF, LOCA, and both for various regions, time slices, units, SSPs, etc.
> 
> My one big thought is on how we display the data when it is all zeros (no precip anywhere within the domain at that timeslice). My current thought is that **if it is easy to do**, we assess what the max precip value is per simulation and set the colorbar range between 0 and that max, either for all timeslices or just when the data all displays 0s so that way the colorbar isn't shifting around (potentially might address the comment @elehmer and I are looking at in terms of plot size), the scale is correct for that unit, and the data should all display in the consistent "low end" color.
> 
> I do think we need to address the "sometimes theres a nan sometimes theres a 0" but that's completely separate from this, IMO

Totally agree. Not sure if that is urgent to resolve before merging this PR though, or is more related to how to handle zeroes in the precip data and could be resolved in a separate PR?

vicford:
> Totally agree. Not sure if that is urgent to resolve before merging this PR though, or is more related to how to handle zeroes in the precip data and could be resolved in a separate PR? Would be a change to `app.view()` and should be added for all variables, in my opinion. I agree that the colorbar shifting around is confusing

I was also thinking more on this, and I do think this actually doesn't preclude merging in this PR; it just identified another important thing to address! I'm fine with it being merged in - the main thing was to add the data back in with the correct units!

vicford (approved):
Approving, as my one off comment is now a separate trello card :)

PR #204, opened 2023-04-25
Fix historical + SSP data retrieval bug

nicolejkeeney:
This PR fixes a bug that was raised when the user tries to retrieve historical + SSP data for a time slice that is out of the bounds of the historical time range. For example, ```app.retrieve()``` would raise an error if you had selected "Historical Climate" and one of the SSPs, but only had selected a time range of 2090-2100 (which doesn't include any data in the historical climate range of 1980-2015). 

**To test**: Make sure you can retrieve the data when you have selected both "Historical Climate" and an SSP for a time slice that is completely in the future (i.e 2030-2035).

### DIFF
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -215,14 +215,14 @@ def _process_and_concat(selections, location, dsets, cat_subset):
 
     if True in ["SSP" in one for one in selections.scenario_ssp]:
         if "Historical Climate" in selections.scenario_historical:
-            # Historical climate will be appended to the SSP data
-            append_historical = True
+            if selections.time_slice[0] > 2015:
+                # Historical climate will be appended to the SSP data
+                append_historical = True
             scenario_list.remove("Historical Climate")
         if "Historical Reconstruction (ERA5-WRF)" in selections.scenario_historical:
             # We are not allowing users to select historical reconstruction data and SSP data at the same time,
             # due to the memory restrictions at the moment
             scenario_list.remove("Historical Reconstruction (ERA5-WRF)")
-
     for scenario in scenario_list:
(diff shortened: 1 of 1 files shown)


vicford (approved):
Works just fine for me now!

climakitae/data_loaders.py
```diff
@@ -215,14 +215,14 @@ def _process_and_concat(selections, location, dsets, cat_subset):
 
     if True in ["SSP" in one for one in selections.scenario_ssp]:
         if "Historical Climate" in selections.scenario_historical:
-            # Historical climate will be appended to the SSP data
-            append_historical = True
+            if selections.time_slice[0] > 2015:
```
elehmer, line 218: Just wondering why this is a `>`? Are we fixing a time slice filter earlier in the code and we set this `True` even though we don't actually append historical?

elehmer (approved):
Works for me as well. Just have a question...

climakitae/data_loaders.py
nicolejkeeney: `append_historical` used to be a param of DataSelectors, which is probably what you're thinking of. The code has since been restructured to automatically append historical if Historical Climate and an SSP is selected. The `append_historical` variable in this chunk of code is just used here in the backend data retrieval, not set by the user. 

The code now checks for time slice > 2015 because the Historical period runs until 2015, so if the time slice is 2016+, we don't need to append the historical period. 

Does that clear it up?

elehmer: I get it but just seems like the logic should be reversed?

nicolejkeeney: Yeah... you're right. Ugh. Just opened a new PR!

PR #206, opened 2023-04-26
Fix data retrieval bug for derived hourly variables

nicolejkeeney:
When working on a different feature, I noticed a bug in the data retrieval of the derived variables. The data retrieval was not working for derived variables if you tried to call `app.retrieve` more than once, because the backend code was setting `selections.variable_id` to a string instead of a list. This fixes that! 

**To test**: Try retrieving a derived variable, which could be one of the following: 
1) Hourly relative humidity
2) Hourly, daily, or monthly dew point temperature 
3) Hourly specific humidity 
4) Hourly wind speed

### DIFF
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -582,7 +582,7 @@ def _read_catalog_from_select(selections, location, cat, loop=False):
                         "You've encountered a bug. No data available for selected derived variable."
                     )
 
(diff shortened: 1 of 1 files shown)


vicford (approved):
No issues retrieving derived variables! Tested with relative humidity and dew point temperature, but can see in the code change that it would be applicable to all

bethem (approved):
tried a few variables and had no problems

PR #209, opened 2023-04-27
Fix wildcard matching for zarr retrieval

nicolejkeeney:
This PR fixes a bug in the file retrieval that uses wildcard matching to deduce the filename for a given simulation and scenario. Previously, it incorrectly would pull "EC-Earth3-Veg" data instead of "EC-Earth3" data, because the substring "EC-Earth3" is in both simulation names. In the code, I added a "." to indicate the the wildcard shouldn't continue after the simulation (maybe that makes no sense...)

**To test**: Not sure how to test this, since no error was ever raised. Just look over the code.


vicford (approved):
Makes sense to me, in terms of the filename construction!

elehmer (approved):

PR #212, opened 2023-05-03
Formatting fixes to sync with Pangeo image: 2022.09.21

nicolejkeeney:
## Overview 
This PR is branched off an outstanding branch `mapview`. It provides the formatting changes required to work around poor formatting in the intermediary pangeo image from 2022.09.21. This image appears to function fine but the figure displays are buggy. The resolution is quite bad and the sizing doesn't make much sense (which is why I had to resize the images and change the fontsize). 

## To test
Run this climakitae branch in the Pangeo image tagged for 2022.09.21 (You will need to change the date in your local Makefile). DO NOT RUN IN THE HUB-- the hub image is set to 2021.10.09, and cannot be changed by anyone but Brian. 

## Answers to questions you may have about why the figures/resolution can't be improved further in this image
For those who may ask, modifying the dpi of the images does NOT have any effect on the figure resolutions. The version of panel in the 2022.09.21 is buggy. I have tried modifying the fig sizes and resolutions of the images, but what you see in this branch is the best I could do. For those who may ask, I can't change the scenario figure in the bottom left to be less tall-- doing so just "cuts" the image instead of reducing its height. 

## Examples
The resolution and figure issues are kind of hard to see in these screenshots, but I included them anyways. The issues with the figures should be obvious if you run this branch in the sep 2022 image, though. 
### 2022.09.21 image, from this branch (`pangeo-img-nicole`): 
Sep 2022 image

### 2021.10.09 image, from branch `mapview`: 
Oct 2021 image

### DIFF
--- a/climakitae/selectors.py
+++ b/climakitae/selectors.py
@@ -390,12 +390,12 @@ def _map_view(selections, stations_gpd):
         ),
     }
 
-    fig0 = Figure(figsize=(3.2, 3.2))
+    fig0 = Figure(figsize=(2.25, 2.25))
     proj = ccrs.Orthographic(-118, 40)
     crs_proj4 = proj.proj4_init  # used below
     xy = ccrs.PlateCarree()
     ax = fig0.add_subplot(111, projection=proj)
-    mpl_pane = pn.pane.Matplotlib(fig0, dpi=144)
+    mpl_pane = pn.pane.Matplotlib(fig0, dpi=1000)
 
     # Get geometry of selected location
     subarea_gpd = _get_subarea(
@@ -1175,7 +1175,7 @@ class _DataSelector(param.Parameterized):
         historical_central_year = sum(historical_climate_range) / 2
         historical_x_width = historical_central_year - historical_climate_range[0]
 
-        fig0 = Figure(figsize=(3, 2))
+        fig0 = Figure(figsize=(2, 2))
         ax = fig0.add_subplot(111)
         ax.spines["right"].set_color("none")
         ax.spines["left"].set_color("none")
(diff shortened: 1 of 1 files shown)


vicford (changes requested):
LGTM!!!

Checked: app.explore.amy, app.explore.warming_levels, all good

Breaks: app.explore.thresholds but because of the locations/selections thing: 
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In [3], line 1
----> 1 app.explore.thresholds()

File ~/climakitae/climakitae/explore.py:74, in _AppExplore.thresholds(self, option)
     71 def thresholds(self, option=1):
     72     """Display Thresholds panel."""
     73     thresh_data = _ThresholdDataParams(
---> 74         selections=self.selections, location=self.location, cat=self._cat
     75     )
     76     return _thresholds_visualize(
     77         thresh_data=thresh_data,
     78         selections=self.selections,
     79         option=option,
     80     )

AttributeError: '_AppExplore' object has no attribute 'location'
```

All the resulting figures look just fine to me, so it appears it's just the little map insert and the scroll bar, which I ultimately think is fine for now. 

Noting that the cluster also did not behave in the thresholds exceedance notebook, but that's a separate issue

nicolejkeeney:
> Breaks: app.explore.thresholds but because of the locations/selections thing:

Good catch, that's probably due to it's parent branch ```mapview```. I will investigate!

nicolejkeeney:
> > Breaks: app.explore.thresholds but because of the locations/selections thing:
> 
> Good catch, that's probably due to it's parent branch `mapview`. I will investigate!

Fixed and pushed so everything should work now!

vicford (approved):
all app.explores work now :) 

Noting, getting this warning for app.explore.thresholds:

```
/srv/conda/envs/notebook/lib/python3.9/site-packages/holoviews/core/data/pandas.py:221: FutureWarning: In a future version of pandas, a length 1 tuple will be returned when iterating over a groupby with a grouper equal to a list of length 1. Don't supply a list with a single grouper to avoid this warning.
  data = [(k, group_type(v, **group_kwargs)) for k, v in
```

bethem (approved):
lgtm! ran localization.ipynb and station_8760s.ipynb just fine

elehmer (approved):
Seems to work well on my local environment.

PR #213, opened 2023-05-04
Remove loop in data retrieval of bias corrected station data

nicolejkeeney:
Previously, we had two options for the retrieval of station data (in the back end, not user facing). One was the first method I wrote, which used a loop to append the stations. The second used xr.apply, which I wrote afterward to be more xarray-ish and more optimized (although, I don't think it's actually any faster/slower than the loop). In this PR, I just removed the loop code and now we just have the apply option. I also improved the function documentation. 

**To test:** Try and retrieve and load some station data using ``app.select()``. Make sure everything works as expected and that there aren't any rouge NaNs or errors!


vicford:
Weirdly enough, getting a "no xclim" on the hub in this branch...

nicolejkeeney:
> Weirdly enough, getting a "no xclim" on the hub in this branch...

That's odd, I'll make a mental note of that issue for the future. Just tested on the hub and didn't get that error personally but good to know it may be an issue.

vicford (commented):
Okay, not getting xclim error now (weird, IDK) but getting this warning about loading data in app.view even though app.load was previously called!
Screen Shot 2023-05-08 at 10 20 26 AM
`/home/jovyan/src/climakitae/climakitae/view.py:58: UserWarning: This function may be quite slow unless you call .compute() on your data before passing it to app.view()`

ETA: ah, i see now that this is addressed in PR#215

vicford (approved):
Switching to Approve, since the read in warning gets addressed in PR #215

bethem (approved):
lgtm! pulled data for an IOU area (6 stations) at once and it went smoothly

PR #214, opened 2023-05-04
Remove app.retrieve_from_csv()

nicolejkeeney:
This PR removes ```app.retrieve_from_csv()``` as an option. From my memory, this was legacy code that no notebooks use anymore. **The same functionality can be achieved** by using ```app.retrieve(config="file.csv")```.

Purpose of doing this is just to clean up ```core.py```. 

**To test**: Can someone help me verify that no notebooks use the outdated call ```app.retrieve_from_csv()```? I already checked and couldn't find anything but it's definitely possible I missed something (perhaps in cae-examples?)

### DIFF
--- a/climakitae/core.py
+++ b/climakitae/core.py
@@ -164,33 +164,6 @@ class Application(object):
                 )
         return _read_catalog_from_select(self.selections, self.catalog, loop)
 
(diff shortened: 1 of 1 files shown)


elehmer (approved):

bethem (approved):
it all looked good to me!

PR #215, opened 2023-05-05
Fix bugs in station data 

nicolejkeeney:
This PR retrieves the entire spatial domain of the model data when station data is retrieved, then finds the closest gridcell using the entire dataset. Previously, the model data could have been subsetted depending on if the user had selected a certain area subset in ```app.select``` before picking the station. Hopefully this solves the issue of the station data sometimes returning nans. My suspicion is that that was occurring because sometimes the closest gridcell was one that had been "NaN"ed by the area subsetting. 

This PR also fixes the erroneous "please read your data" warning in ```app.view``` for station data, since the rechunking in the bias correction sets the dask chunks to an empty dask "Frozen" object instead of "None".

**To test:** Retrieve some station data using the area subsetting (i.e. select a spatial region to reduce the station options first). Then retrieve a weather station or two's worth of data. Then call ```app.load(data)``` before ```app.view(data)``` and make sure no warning is raised telling you to read the data into memory.

### DIFF
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -175,9 +175,7 @@ def _get_area_subset(area_subset, cached_area, selections):
             )
         ds_region = [geom]
     elif area_subset != "none":
-        shape_index = int(
-            selections._geography_choose[selections.area_subset][selections.cached_area]
-        )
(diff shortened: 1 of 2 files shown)


Tianchi-Liu (approved):
Looks good to me.

PR #210, opened 2023-04-28
Rewrite map_view to be an attribute of DataSelectors

nicolejkeeney:
This is a purely backend change, the user won't notice anything different. Now, map_view is an attribute of DataSelectors instead of Application. This cleans up the code quite a lot. 

**To test:** Just play around with climakitae (i.e. `app.select()`, `app.view()`, `app.explore`...) and make sure you don't find any weird bugs with the display of the map in the panel.

PR #216, opened 2023-05-08
Fix bug in display of lat/lon selections

nicolejkeeney:
Tianchi had noticed a bug in the map display of lat/lon selection. If you selected too large of a lat/lon region using the slider box thing, ```app.select()``` would throw an error. This PR just resolves that! 

See this comment for the error message (which you should *not* see in this branch!): https://github.com/cal-adapt/climakitae/pull/210#pullrequestreview-1415613228

**To test**: Open up ```app.select()``` in this branch. Try to use the sliders to adjust the lat/lon selection to a large region. Ensure no error is raised.


bethem (approved):
worked for me locally

PR #218, opened 2023-05-09
Fix bug in 3km data retrieval for SSPs 585 and 245 

nicolejkeeney:
We don’t have gridded data at 3km for SSPs 585 and 245 but ```app.select()``` shows them as options for 3km station data, then promptly raises a nonsensical error when you attempt to retrieve the data. 

This PR does the following: 
1) Removes SSP 585 and 245 as options when 3km is selected
2) Modifies the ```data_loaders``` module to raise a descriptive error if no data is found. Hopefully this should never actually be raised, but is in place for us developers in case things fall through the cracks. 

**To test**: Ensure that when you select "station" and "3km", only Historical Climate and SSP 375 are shown as options in ```app.select()```

### DIFF
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -323,12 +323,19 @@ def _process_and_concat(selections, dsets, cat_subset):
                     )
                     sim_list.append(da_sim_member_id)
 
+        # Raise an appropriate error if no data found
+        if len(sim_list) == 0:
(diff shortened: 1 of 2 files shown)


vicford (approved):
Only 370 showing up for 3km station data :)

bethem (approved):
lgtm!

PR #219, opened 2023-05-15
Add xclim

nicolejkeeney:
Add xclim to package requirements

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,4 +22,4 @@ rioxarray==0.13.3
 s3fs==2022.11.0
 scipy>=1.2.0
 shapely==1.8.5
-xclim==0.41.0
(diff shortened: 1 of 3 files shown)


thenaomig:
Looks good to me. Thanks.

requirements.txt
```diff
@@ -22,4 +22,4 @@ rioxarray==0.13.3
 s3fs==2022.11.0
 scipy>=1.2.0
 shapely==1.8.5
-xclim==0.41.0
+xclim==0.36.0
```
bkg, line 25: We've been using `0.42.0` on the hub, so it would be good to go with that.

nicolejkeeney: Good to know-- thanks!

PR #220, opened 2023-05-16
Fix unit conversions for derived variables

nicolejkeeney:
The derived variable retrieval only worked if you selected the default variable. For example, you could grab dew point in Kelvin, but not degF or degC. This PR fixes that issue.


vicford (changes requested):
All derived variables + alternate units look okay _except_ dew point temperature

(wow climate change is getting bad) In all seriousness, I checked out the values for all units for dew point temperature as well as the conversions, and even though the values are way off, the conversion between them is correct. I also double checked the formula, and it's not that either. It looks like dew point temperature calculated in K is doing something funky? I don't see where that changed in the code though. I restarted the kernel several times as well as restarted the hub overall and this didn't fix it, fyi

K
Screen Shot 2023-05-17 at 10 43 24 AM

degF
Screen Shot 2023-05-17 at 10 45 25 AM

degC
Screen Shot 2023-05-17 at 10 44 06 AM

check of formula - values are in correct range
Screen Shot 2023-05-17 at 10 51 38 AM

vicford (approved):
hooray all works now! just needs black formatting

nicolejkeeney:
> hooray all works now! just needs black formatting

Ugh it's actually a failing windows test, unrelated to black, but related to weirdness in the conda testing env? Hopefully will go away once the testing cache is reset after merging.

bethem (approved):
looks right to me

PR #221, opened 2023-05-17
Remove missed loop code

nicolejkeeney:
I thought I had removed all the residual looping code for retrieval of bias correction data in this PR, but I missed a few things. This PR just removes those. 

**To test**: Try retrieving some weather station data. Does it work?


elehmer (approved):
Looks good to me. Works on the HUB for different stations and other variables changed.

PR #222, opened 2023-05-18
app.view() should have a fixed colorbar across simulations

nicolejkeeney:
This PR makes it so app.view() has a fixed colorbar limits across simulations, by computing the min and max (1 and 99th percentile) of the data across the simulation dimension, if the data contains that dimension. I reused the helper function ```_compute_vmin_vmax``` from the warming_levels module, and moved it into the view module, where I think it is more appropriately located. 

**To test**: Make some plots! Make sure the colorbar limits are fixed. Note-- this will only work if you've already loaded the data into memory before calling `app.view(data)`, because computing the min and max will take forever on a dask lazily loaded array.

### DIFF
--- a/climakitae/view.py
+++ b/climakitae/view.py
@@ -16,6 +16,19 @@ var_catalog_resource = pkg_resources.resource_filename(
 var_catalog = pd.read_csv(var_catalog_resource, index_col=None)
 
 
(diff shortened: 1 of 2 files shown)


vicford (approved):
This is a MASSIVE improvement, major kudos @nicolejkeeney!

PR #224, opened 2023-05-18
Add wind direction as derived variable

vicford:
This PR adds wind direction as an hourly variable. 

**To test**
Try retrieving hourly wind direction! Noting, would recommend this for a coarse resolution and/or small location because of the data involved in computing. 

***Note:*** Wind direction is not normally displayed as such. I have a trello card up to add a AE-branded cyclic colormap that I'll get to next, as well as setting the colorbar limits to 0-360. **If** we decide to pursue wind velocity in the future (magnitude + direction), we can investigate barbs for wind direction as an option too.


nicolejkeeney (approved):
Looks great to me, had a few suggestions but that's about it!

bethem (approved):
looks good to me

PR #226, opened 2023-05-22
Add in selection of LOCA2 monthly variables 

nicolejkeeney:
This PR adds in the functionality for selecting and retrieving monthly LOCA2 variables using climakitae app.select() and app.retrieve()

NOTE: You'll see I had to swap the order of the timescale dropdown into an unintuitive order: daily, monthly, hourly. This is due to panel being a bit annoying and filling in the wrong option as they are subsequently restricted and expanded depending on the selection (please PM me if you want a more detailed explanation!). I know it sucks, but I think this is the best fix. 

**To test**: Try retrieving monthly LOCA2 data!


vicford (approved):
All good for me!

elehmer (approved):
Works well on HUB. Tried several different settings.

PR #232, opened 2023-05-29
Add mph conversion for wind speed

nicolejkeeney:
I'm developing some wildfire indices and need windspeed in miles per hour, so I figured I'd add it as a unit conversion option since it's not one already. 

I also deleted a redundant repeat code line. 

**To test**: Confirm that the wind speed in mph looks right.


vicford (approved):
All checks out for me! Checked m/s to mph, and kts to mph

elehmer (approved):

PR #234, opened 2023-05-31
Add Fosberg Fire Weather Index. Clean up data_loaders module

nicolejkeeney:
**Draft PR: Do not review yet**

## Overview
This PR does the following: 
1) Added the Fosberg Fire Weather Index ([ref](https://a.atmos.washington.edu/wrfrt/descript/definitions/fosbergindex.html)) as a new hourly derived variable for WRF 
2) Cleaned up the data_loaders module a bit to make the derived variable code more readable.

## Notes
**This variable takes a long time to load.** Why? The index itself takes three variables: air temp, wind direction, and relative humidity. We need to derive both wind direction and relative humidity. Derivation of wind direction requires u and v, and relative humidity requires mixing ratio, air temp, and pressure. Therefore, **deriving the Fosberg Fire Index on the fly in climakitae requires 5 variables**: air temp, u, v, mixing ratio, and pressure.

### Example loading times
- 9 km, Alameda County, historical 1980-1990 (10 yrs): reading data into memory takes 5m 49.35s
- 45 km, state of CA, SSP 370 2090-2095 (5 yrs): reading data into memory takes 3m 49.82s

## Soliciting opinions
The new variable will appear if you have selected "hourly" and "Dynamical". Do folks think it's location in the dropdown menu is appropriate (see below)? 
fosberg

## Examples of the Fosberg Fire Weather Index in action 
CA_45km


vicford:
2-cent opinion worth nothing: I think indices should be at the bottom of the dropdown menu, not in the mix of data variables. If we intend to add a bunch of other indices, it would be cool to have a separate dropdown menu (variable vs. index) for users to select from, but that could be a bit of a headache

nicolejkeeney:
> 2-cent opinion worth nothing: I think indices should be at the bottom of the dropdown menu, not in the mix of data variables. If we intend to add a bunch of other indices, it would be cool to have a separate dropdown menu (variable vs. index) for users to select from, but that could be a bit of a headache

Yeah, I agree with both of your opinions!

elehmer:
> > > 2-cent opinion worth nothing: I think indices should be at the bottom of the dropdown menu, not in the mix of data variables. If we intend to add a bunch of other indices, it would be cool to have a separate dropdown menu (variable vs. index) for users to select from, but that could be a bit of a headache
> > 
> > 
> > Yeah, I agree with both of your opinions! Adding a separate dropdown would definitely be an option.
> 
> The other annoying issue is that the user doesn't realize we have this variable until they click the "hourly" option and then back up to the variable picklist. I almost feel like the variable picklist should come after the other options. But I guess there isn't really a good way to fix this as the interface of options is quite complicated.

Hmm, for any appreciable size, time duration this is too computationally expensive. We should generate the result and store as a zarr store if it is going to be used a lot.

vicford:
Further thought: Is it too on-the-nose to have the colormap be ae_orange?

thenaomig:
I think that all the functions for dervied variables should live in a different module altogether, so the deriving itself is separate from 1. loading what you need to derive it, and 2. returning a single derived thing. And that way we keep data_loaders.py from getting infinitely longer over time.

nicolejkeeney:
> Further thought: Is it too on-the-nose to have the colormap be ae_orange?

Modified and changed to ae_orange!

nicolejkeeney:
> > > 2-cent opinion worth nothing: I think indices should be at the bottom of the dropdown menu, not in the mix of data variables. If we intend to add a bunch of other indices, it would be cool to have a separate dropdown menu (variable vs. index) for users to select from, but that could be a bit of a headache
> > 
> > 
> > Yeah, I agree with both of your opinions! Adding a separate dropdown would definitely be an option.
> 
> The other annoying issue is that the user doesn't realize we have this variable until they click the "hourly" option and then back up to the variable picklist. I almost feel like the variable picklist should come after the other options. But I guess there isn't really a good way to fix this as the interface of options is quite complicated.

I think I've fixed this with the new dropdown for derived indices :D

vicford (changes requested):
Noticing on the hub that the data loads speedily, but in app.view the map does not change with time or with simulation selection. When looking at a time series, there is variation over time, but again it doesn't change with a change in simulation

climakitae/data/variable_descriptions.csv
```diff
@@ -1,4 +1,5 @@
 variable_id,downscaling_method,timescale,unit,display_name,extended_description,colormap,show
+fosberg_index_derived,Dynamical,hourly,[0 to 100],Fosberg fire weather index,"Fire weather index using meteorological inputs (temperature, relative humidity, and windspeed) to quantify fire risk.",ae_orange,TRUE
```
vicford, line 2: Wondering if we should match the extended description further to match the FFWI reference more?

Something like "Fire weather index using meteorological inputs (temperature, relative humidity, windspeed) to quantify fire risk. Large values imply a higher risk value."

climakitae/indices.py
```diff
@@ -0,0 +1,93 @@
...
+
+    """
+    # Compute the equilibrium moisture constant
+    m_low, m_mid, m_high = _equilibirum_moisture_constant(h=rh_percent, T=t2_F)
+
+    # For RH > 10%, use the low m value.
+    # For RH >= 10%, use the mid value
+    m = xr.where(rh_percent < 10, m_low, m_mid)
+    # For RH > 50%, use the high m value.
+    m = xr.where(rh_percent > 50, m_high, m)
```
vicford, line 39: Just checking that there's not a missing line of code for the low m value here?

climakitae/indices.py
```diff
@@ -0,0 +1,93 @@
...
+    ----------
+    https://a.atmos.washington.edu/wrfrt/descript/definitions/fosbergindex.html
+    https://github.com/sharppy/SHARPpy/blob/main/sharppy/sharptab/fire.py
+    https://www.spc.noaa.gov/exper/firecomp/INFO/fosbinfo.html
+
+    """
+    # Compute the equilibrium moisture constant
+    m_low, m_mid, m_high = _equilibirum_moisture_constant(h=rh_percent, T=t2_F)
+
+    # For RH > 10%, use the low m value.
```
vicford, line 35: Also I think this should be RH < 10% here

climakitae/indices.py
nicolejkeeney: The `low_m` is kind of sneakily in there: 
```
m = xr.where(rh_percent < 10, m_low, m_mid)
```
Where RH < 10%, use m_low. Where it's > 10% use m_mid.

climakitae/indices.py
nicolejkeeney: oops, yeah the code comment is off. fixed

climakitae/data/variable_descriptions.csv
nicolejkeeney: Love it. Just updated

nicolejkeeney:
> After setting station as the data type, Derived Index is visible as a variable type. When Derived Index is selected, although station disappears as a data type, timescale becomes daily on the GUI, which doesn't match the timescale of the index variable. Could the Derived Index option be hidden once data type is set to station? Or could the timescale be set to hourly?

@Tianchi-Liu I think I fixed this just now-- good catch!

climakitae/indices.py
```diff
@@ -0,0 +1,93 @@
...
+
+    # Add descriptive attributes
+    FFWI.name = "Fosberg Fire Weather Index"
+    FFWI.attrs["units"] = "[0-100]"
+
+    return FFWI
+
+
+# Define some helper functions
+def _equilibirum_moisture_constant(h, T):
```
jmlongmate, line 56: *_equilibrium...

climakitae/indices.py
```diff
@@ -0,0 +1,93 @@
...
+    m = xr.where(rh_percent < 10, m_low, m_mid)
+    # For RH > 50%, use the high m value.
+    m = xr.where(rh_percent > 50, m_high, m)
+
+    # Compute the moisture dampening coefficient
+    n = _moisture_dampening_coeff(m)
+
+    # Compute the index
+    U = windspeed_mph
+    FFWI = (n * ((1 + U**2) ** 0.5)) / 0.3002
```
jmlongmate, line 46: Does there need to be an additional line here (or in _convert_units) restricting the maximum possible output to be 100? Currently if I force values of 100F, 0% humidity, and 50mph windspeed, it returns a FFWI value of ~166 

From the a.atmos.washington.edu reference: "The values were designed so that a index rating of 100 is equal to a moisture content of 0 and a wind speed of 30 mph. Larger combinations of these values still result in an index of 100."

climakitae/data_loaders.py
```diff
@@ -664,103 +677,40 @@ def _read_catalog_from_select(selections, cat):
...
-                    da = _compute_specific_humidity(
-                        tdps=dew_pnt_da, pressure=pressure_da  # Kelvin  # Pa
-                    )
-
-                else:
-                    raise ValueError(
-                        "You've encountered a bug. No data available for selected derived variable."
-                    )
-
         da = _convert_units(da, selected_units=orig_unit_selection)
```
jmlongmate, line 703: I'm not sure where native_units gets defined, but I think if the fire index goes into here with native_units as [0-100] then it might get scaled down to cap the max at 100, rather than reassigning all values >100 to 100?

jmlongmate (changes requested):
Mostly checking that the fire weather index maxes out at 100 for very low rh and high windspeeds >30mph, and doesn't get re-scaled.

climakitae/indices.py
nicolejkeeney: good catch lol

climakitae/selectors.py
```diff
@@ -839,16 +858,19 @@ def _update_area_average_based_on_data_type(self):
...
 
-    @param.depends("downscaling_method", "data_type", watch=True)
-    def _update_data_type_options_if_loca_selected(self):
+    @param.depends("downscaling_method", "data_type", "variable_type", watch=True)
+    def _update_data_type_options_if_loca_or_derived_var_selected(self):
         """If statistical downscaling is selected, remove option for station data because we don't
         have the 2m temp variable for LOCA"""
-        if "Statistical" in self.downscaling_method:
+        if (
+            "Statistical" in self.downscaling_method
```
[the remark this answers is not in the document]
nicolejkeeney: I reallllllyyyy wish I could disable the behavior that allows users to deselect both Dynamical and Statistical, but it's not currently an option with the library we are using. It's an unfortunately behavior that I think is confusing to users (as you have pointed out), but it's not something we can change :(

climakitae/indices.py
nicolejkeeney: great catch, I'll fix this!

climakitae/indices.py
```diff
@@ -0,0 +1,99 @@
...
+    # If fosberg index > 100, reset to 100
+    FFWI = xr.where(FFWI < 100, FFWI, 100, keep_attrs=True)
+
+    # If fosberg index is negative, set to 0
+    FFWI = xr.where(FFWI > 0, FFWI, 0, keep_attrs=True)
+
+    # Add descriptive attributes
+    FFWI.name = "Fosberg Fire Weather Index"
+    FFWI.attrs["units"] = "[0 to 100]"
+
```
jmlongmate, line 57: Looks good to me!

jmlongmate (approved):
Looks good!

PR #235, opened 2023-06-01
Update var names for new LOCA2 variables

nicolejkeeney:
Update the variable names in the lookup csv to match the variable and file names of the zarr stores in our catalog for LOCA2 min/max relative humidity. 

**To review**: Makes sure you can retrieve the following variables: 
- LOCA2 minimum relative humidity 
- LOCA2 maximum relative humidity 
- LOCA2 wind speed at 10m


gdicecco (approved):
Retrieved these three variables just fine on the hub!

vicford (approved):
No problems for me, retrieved all three!

elehmer (approved):
Works well on the HUB!

PR #237, opened 2023-06-05
Remove numba from environment.yml so that conda env solvable.

elehmer:
This PR removes numba requirement from `environment.yml` that was added in #190. I am not entirely sure why it is not needed anymore. Shrug. But the conda environment should hopefully be solvable now...


elehmer:
Interesting that `numba==0.53.3` is actually installed as a dependency so all good I guess.

PR #240, opened 2023-06-15
Add flexible/custom extreme event definition functionality to threshold_tools 

corakingdon:
major changes:
- rename `get_ams()` to `get_block_maxima()`, because the function is now more general, allowing for different length block sizes other than annual (default block size is 1 year)
- optional arguments `duration`, `groupby`, and `grouped_duration` are now implemented in the `get_block_maxima()` function for defining custom extreme events
- `get_block_maxima()` checks the average effective sample size of the data in each block, with the new `calculate_ess()` function, and displays a warning if too low
- return value, return period, and return probability calculations are now flexible to different block sizes (in internal function `_calculate_return()`)
- unit tests added for these changes in "test_threshold_returns.py"


vicford:
Is there a plan to update the threshold_tools notebooks (perhaps basics) to illustrate this updated functionality? Based off of testing in threshold_basics, it seems to me that the new function does not work with the current structure (time series vs. spatial data, etc.)

climakitae/threshold_tools.py
```diff
@@ -14,41 +14,162 @@
...
+from warnings import warn
 
 
-def get_ams(da, extremes_type="max"):
-    """Function that converts data into annual maximums
+def calculate_ess(data, nlags=None):
+    """
+    Function for calculating the effective sample size (ESS) of the provided data.
+
+    Input array is assumed to be timeseries data with potential autocorrelation.
```
vicford, line 26: I'd recommend placing a check in here to ensure that it is timeseries data. There's an example in the data_loaders code iirc

climakitae/threshold_tools.py
```diff
@@ -14,41 +14,162 @@
...
+        # In this case, user is interested in extreme events lasting at least
+        # as long as the length of `duration`.
+        dur_len, dur_type = duration
+        if dur_type != "hour" or da_series.frequency not in ["1hr", "hourly"]:
+            raise ValueError(
+                "Current specifications not implemented. `duration` options only implemented for `hour` frequency."
+            )
+
+        # First identify the min (max) value for each window of length `duration`
+        if extremes_type == "max":
```
vicford, line 102: Are the extremes_type == max and == min switched? I'm seeing min applied fo max, and max applied for min

climakitae/threshold_tools.py
```diff
@@ -14,41 +14,162 @@
...
+        # In this case, identify the min (max) value of the grouped values for
+        # each window of length `grouped_duration``. Must be in `days`.
+        dur2_len, dur2_type = grouped_duration
+        if dur2_type != "day":
+            raise ValueError(
+                "`grouped_duration` specification must be in days. example: `grouped_duration = (3, 'day')`."
+            )
+
+        # Now select the min (max) from the duration period
+        if extremes_type == "max":
```
vicford, line 136: Same here with max and min. At Ln 117-120 it's correctly set

corakingdon:
> Is there a plan to update the threshold_tools notebooks (perhaps basics) to illustrate this updated functionality? Based off of testing in threshold_basics, it seems to me that the new function does not work with the current structure (time series vs. spatial data, etc.)

@vicford thanks for checking this out already!! and for catching the last thing I need to update/decide on-- whether/how to implement the check_ess for spatial data as well (I wasn't able to get apply_ufunc to work how I wanted, so I'm going to try using stack/unstack today and if I can't figure it out I'll ask for help with this next week!)

In terms of notebook updates: yes, I will update the existing ones to use the new function name, and I have a new notebook that is specifically just about how to use the new functionality. I haven't created a branch/PR for the notebook updates yet though-- I was going to check in on Monday about the current workflow for doing this!

climakitae/threshold_tools.py
corakingdon: It's a bit confusing, but pretty sure this is right (but not well documented!! I should add some better explanatory comments). Basically, to get the annual _max_ event lasting for a certain duration (ex: 72 hours), you taking the _min_ value for each 72-hour window of that length, and then to get the annual maximum you take the max of each of these mins...

climakitae/threshold_tools.py
```diff
@@ -356,8 +489,106 @@ def _conf_int(
...
+    new_ds["conf_int_upper_limit"].attrs[
+        "confidence interval upper bound"
+    ] = "{}th percentile".format(str(conf_int_upper_bound))
+
+    new_ds.attrs = bms_attributes
+    new_ds.attrs["distribution"] = "{}".format(str(distr))
+    return new_ds
+
+
 def _get_return_variable(
```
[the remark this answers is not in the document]
corakingdon: Oh I think this is a mistake on my part! I think accidentally added the `get_return_variable ` function as a copy of an older version of `_get_return_variable`, I think got messed up in a merge conflict from main at some point.... I should just be using `_get_return_variable`, I'll delete the other now!! thank you!!

climakitae/threshold_tools.py
```diff
@@ -14,41 +14,180 @@
...
+                    input_core_dims=[["time"]],
+                ).unstack("allpoints")
+                all_ess.append(ess)
+        else:
+            # case for not spacial data (no x, y dimensions)
+            for yr in set(bms.time.dt.year.values[0:-1]):
+                ess = calculate_ess(
+                    da_series.sel(time=slice(f"{yr}", f"{yr+block_size-1}"))
+                )
+                all_ess.append(ess)
```
corakingdon, line 174: Looking for feedback on how to re-write this section better (cleaner, more efficient). Just trying to apply the `calculate_ess` function to each dimension (x, y) except for time, which it has to be applied to in separate year (or block_size) chunks (currently as a for loop). I tried to group by year, then call apply, but couldn't get it to work. The current setup uses stack/unstack for the case of x,y data, but am wondering if there is a simpler way to do this.

climakitae/threshold_tools.py
vicford: Okay, if I'm following correctly: 

For an annual max event for events of a certain duration (say there are 5 72-hour events in a year), for each of those 5 events, you take the min value of that event which gives you the lowest value above the threshold, and then for the annual statistic you take the max of those "lowest above threshold" values and this gives you... a baseline of a sorts of the most extreme max event?

For annual min, is it then correspondingly the "highest value below the threshold" (so closest to the threshold) and then the min/lowest of each of those "maxes"? 

So in short, we're not looking for the absolute max value in each max event and taking the max of those maxes, which would just be the absolute extreme (and probably easier to calculate!), but rather the max "baseline" of an event above that threshold, and vice versa for mins?

climakitae/threshold_tools.py
```diff
@@ -14,41 +14,179 @@
...
+        # really only meaningful when coupled with the `grouped_duration` option.)
+        group_len, group_type = groupby
+        if group_type != "day":
+            raise ValueError(
+                "`groupby` specifications only implemented for 'day' groupings."
+            )
+
+        # select the max (min) in each group
+        if extremes_type == "max":
+            da_series = da_series.resample(time=f"{group_len}D", label="left").max()
```
vicford, line 117: Assuming that these ones here are correctly set for max/min?

climakitae/threshold_tools.py
```diff
@@ -14,41 +14,179 @@
...
+            for yr in set(bms.time.dt.year.values[0:-1]):
+                ess = calculate_ess(
+                    da_series.sel(time=slice(f"{yr}", f"{yr+block_size-1}"))
+                )
+                all_ess.append(ess)
+
+        average_ess = np.nanmean(all_ess)
+        if average_ess < 25:
+            print(
+                f"WARNING: The average effective sample size in your data is {round(average_ess, 2)} per block, which is low. This may result in biased estimates of extreme value distributions when calculating return values, periods, and probabilities from this data."
```
vicford, line 178: Without making this a longer error, I think we should modify this somewhat to include the target of 25 (also, statistically, where did 25 come from?), and/or advice on how to increase the ESS, because a warning for example of "the average ESS in your data is 14.23 per block which is low" may mean very little to non-technical users as to how to fix (and maybe this technical user). It was not immediately obvious to me that I needed to pick a location with more grid cells/finer spatial resolution (for spatial xy data), and not increase the length of the time record being analyzed. 

I'm going to amend this, all my attempts to increase ESS have gotten me up to about 15, so what is the recommendation for increasing ESS (or above 25)?

vicford: Noting that this is testing in threshold_basics notebook

nicolejkeeney (changes requested):
I requested a few changes but otherwise it looks good to me! I'm still thinking on that one code chunk but wanted to leave my review first since we can always refactor that code in a different PR.

nicolejkeeney:
Hi Cora, I spent some time working through that tricky code chunk and I think I've cleaned it up a bit and made it more xarray-friendly. I needed to modify the `check_ess` function as well to ensure it returned an xarray object. The code is attached here in a zipped python file. Let me know what you think! 
[nicole_threshold_suggestions.py.zip](https://github.com/cal-adapt/climakitae/files/11887761/nicole_threshold_suggestions.py.zip)

vicford (approved):
This looks fine to me @corakingdon - there may be some modifications needed in the refactor, but like with the notebook, happy to assist there too

elehmer (approved):

elehmer:
Superseded by #288

PR #241, opened 2023-06-19
Wind direction: new colormap for grid plotting + vectorfield standard plotting

vicford:
This PR adds:
- a new colormap: ae_cyclic for wind direction, if it is to plotted by itself
- new functions in order to plot wind direction as is standard in meteorological applications as a vector field of barbs: 
    - _compute_wind_vel, which computes a Dataset with wind speed and direction as variables, because the hvplot.vectorfield function requires these two vars to be in the same space
    - plot_wind_velocity, which plots wind velocity as a gridded map of wind speed overlaid by a vectorfield of wind direction. 
         - **NOTE**: for the vectorfield to be intelligble, wind direction must be plotted in radians. However wind direction is most understandable as degrees. Radians is only provided, thus far, for plotting in this manner. 
         - Wind direction can also be pulled directly via app.select, and will be provided in degrees (only?)
         - Will also note that the coords on this are funky so far, looking for guidance on how to fix!

---
To test:
1. Select a location of interest, at hourly timescales, and load both **u and v winds** and use the following function to plot wind velocity `plot_wind_velocity(u,v)`
Screen Shot 2023-06-19 at 4 37 38 PM

2. To test the new colormap, look at wind direction straight from app.select with app.view()
Screen Shot 2023-06-19 at 4 40 28 PM


vicford:
Code retained locally by @vicford upon revisiting. Branch to be deleted.

PR #245, opened 2023-06-28
Use conda lock files for testing workflow and unpin environment.yml

elehmer:
This PR rewires the testing workflow to use conda lock files which allows for the unpinning of the `environment.yml` file. It also removes caching of the conda env as it is actually much slower on Windows runner.


nicolejkeeney (approved):
Wondering if this will fix the issues in the data_loaders_refactor branch. Thanks for looking into this, Eric!

elehmer:
> Wondering if this will fix the issues in the data_loaders_refactor branch. Thanks for looking into this, Eric!

I'm afraid not. I merged this branch with the data loaders one and still got the same error

elehmer:
> > Wondering if this will fix the issues in the data_loaders_refactor branch. Thanks for looking into this, Eric!
> 
> I'm afraid not. I merged this branch with the data loaders one and still got the same error

Solved with 5e4b2924ecf36cc5c4f1a05898e88b2170d58a0d

PR #247, opened 2023-07-07
Add derived index: NOAA heat index

nicolejkeeney:
This PR builds off the `fire` branch, which adds in the structure for derived indices. Please review [that PR](https://github.com/cal-adapt/climakitae/pull/234) before reviewing this one! 

Add NOAA heat index as a derived variable option for hourly WRF data. To find this option, select "Dynamical" (WRF) and "hourly". Then pick "Derived Index" and select "NOAA heat index" from the dropdown! :D


climakitae/indices.py
```diff
@@ -1,6 +1,73 @@
...
+    HI_lowT_highRH = HI + adj_lowT_highRH  # Add adjustment from HI
+
+    # Use different equation if heat index if the heat index value < 80
+    low_HI = 0.5 * (T + 61.0 + ((T - 68.0) * 1.2) + (RH * 0.094))
+
+    # Adjust heat index depending on different condions for RH, T, and valid range of HI
+    HI = xr.where((RH < 13) & (T > 80), HI_highT_lowRH, HI)
+    HI = xr.where(((RH > 85) & (T < 87) & (T > 80)), HI_lowT_highRH, HI)
+    HI = xr.where(HI < 80, low_HI, HI)
+    HI = xr.where(
```
vicford, line 61: Do we have a reference for this piece for adjustments for temps above 112 deg F being set to nan, or is there a different equation to be used in such cases? Considering LA county might get to 112+ _today_ it definitely is within reason to consider such a scenario now (I can help on this if needed!)

vicford (commented):
Looks good to me! Just had a question about the high end modification

climakitae/indices.py
nicolejkeeney: Ooo good point. There was a lot of confusing wording in the NOAA page that described how to compute it. I misread it and assumed it was invalid above 112 degF, but I'm realizing that just means I don't need to apply an adjustment factor. Good catch!! Will fix that now.

Here's the reference doc if it's of interest: https://www.wpc.ncep.noaa.gov/html/heatindex_equation.shtml

vicford (approved):

Tianchi-Liu:
The regression equation used to compute the heat index is "not valid for extreme temperature and relative humidity conditions beyond the range of data considered by Steadman" according to the reference, and it's unclear to me if using the simpler equation when the regression equation gives an index below 80F makes it valid. That to me warrants a check for the data range Steadman considered.

Tianchi-Liu:
> > The regression equation used to compute the heat index is "not valid for extreme temperature and relative humidity conditions beyond the range of data considered by Steadman" according to the reference, and it's unclear to me if using the simpler equation when the regression equation gives an index below 80F makes it valid. That to me warrants a check for the data range Steadman considered.
> 
> The additional documentation clears this up I think (you just need to scroll down the page): "The Rothfusz regression is not appropriate when conditions of temperature and humidity warrant a heat index value below about 80 degrees F. In those cases, a simpler formula is applied to calculate values consistent with Steadman's results"

Yeah, that's my first reaction too, that using the simpler equation would make the index valid, but that part is not crystal clear to me. The quote about data range appears after it, and makes no reference about the quote about the simpler equation. I could be reading too much into this, but the data range may potentially be something different than the regression equation giving an index less than 80F.

nicolejkeeney:
> > > The regression equation used to compute the heat index is "not valid for extreme temperature and relative humidity conditions beyond the range of data considered by Steadman" according to the reference, and it's unclear to me if using the simpler equation when the regression equation gives an index below 80F makes it valid. That to me warrants a check for the data range Steadman considered.
> > 
> > 
> > The additional documentation clears this up I think (you just need to scroll down the page): "The Rothfusz regression is not appropriate when conditions of temperature and humidity warrant a heat index value below about 80 degrees F. In those cases, a simpler formula is applied to calculate values consistent with Steadman's results"
> 
> Yeah, that's my first reaction too, that using the simpler equation would make the index valid, but that part is not crystal clear to me. The quote about data range appears after it, and makes no reference about the quote about the simpler equation. I could be reading too much into this, but the data range may potentially be something different than the regression equation giving an index less than 80F.

Sorry to belabor this point, but I'm having a hard time following. It seems clear to me from the wording that when the heat index, given my the first equation (given by regression equation of Rothfusz ) gives a value of below 80 deg F, you should instead replace that value with a heat index computed by the simpler equation. 

Can you clarify where you are confused? You may be correct that there is some ambiguity here, but I'm just having a hard time following.

vicford:
> > > > The regression equation used to compute the heat index is "not valid for extreme temperature and relative humidity conditions beyond the range of data considered by Steadman" according to the reference, and it's unclear to me if using the simpler equation when the regression equation gives an index below 80F makes it valid. That to me warrants a check for the data range Steadman considered.
> > > 
> > > 
> > > The additional documentation clears this up I think (you just need to scroll down the page): "The Rothfusz regression is not appropriate when conditions of temperature and humidity warrant a heat index value below about 80 degrees F. In those cases, a simpler formula is applied to calculate values consistent with Steadman's results"
> > 
> > 
> > Yeah, that's my first reaction too, that using the simpler equation would make the index valid, but that part is not crystal clear to me. The quote about data range appears after it, and makes no reference about the quote about the simpler equation. I could be reading too much into this, but the data range may potentially be something different than the regression equation giving an index less than 80F.
> 
> Sorry to belabor this point, but I'm having a hard time following. It seems clear to me from the wording that when the heat index, given my the first equation (given by regression equation of Rothfusz ) gives a value of below 80 deg F, you should instead replace that value with a heat index computed by the simpler equation.
> 
> Can you clarify where you are confused? You may be correct that there is some ambiguity here, but I'm just having a hard time following.

@nicolejkeeney @Tianchi-Liu, If this helps, the NWS has its own handy web-based calculator (https://www.wpc.ncep.noaa.gov/html/heatindex.shtml) for Heat Index. Testing values **below 80F** confirms that the simpler equation is how Heat Index values are calculated with temps below 80F (I tested this against the simple formula and got identical answers). While it is fair that we typically don't see Heat Index values below 80*F in the usual charts, this may just be not clear references ultimately, but is coded correctly here and matches the official method.

Tianchi-Liu:
On a different topic, the reference page also noted at the bottom a different "practical" way to compute the index. The web calculator seems to be using the practical way instead (https://www.ncl.ucar.edu/Document/Functions/Heat_stress/heat_index_nws.shtml). Setting T = 60F and RH = 50% confirms this. If reproducing the results from the web calculator is desirable, the website I just linked has some more helpful information.

Tianchi-Liu:
> > > > > > The regression equation used to compute the heat index is "not valid for extreme temperature and relative humidity conditions beyond the range of data considered by Steadman" according to the reference, and it's unclear to me if using the simpler equation when the regression equation gives an index below 80F makes it valid. That to me warrants a check for the data range Steadman considered.
> > > > > 
> > > > > 
> > > > > The additional documentation clears this up I think (you just need to scroll down the page): "The Rothfusz regression is not appropriate when conditions of temperature and humidity warrant a heat index value below about 80 degrees F. In those cases, a simpler formula is applied to calculate values consistent with Steadman's results"
> > > > 
> > > > 
> > > > Yeah, that's my first reaction too, that using the simpler equation would make the index valid, but that part is not crystal clear to me. The quote about data range appears after it, and makes no reference about the quote about the simpler equation. I could be reading too much into this, but the data range may potentially be something different than the regression equation giving an index less than 80F.
> > > 
> > > 
> > > Sorry to belabor this point, but I'm having a hard time following. It seems clear to me from the wording that when the heat index, given my the first equation (given by regression equation of Rothfusz ) gives a value of below 80 deg F, you should instead replace that value with a heat index computed by the simpler equation.
> > > Can you clarify where you are confused? You may be correct that there is some ambiguity here, but I'm just having a hard time following.
> > 
> > 
> > @nicolejkeeney @Tianchi-Liu, If this helps, the NWS has its own handy web-based calculator (https://www.wpc.ncep.noaa.gov/html/heatindex.shtml) for Heat Index. Testing values **below 80F** confirms that the simpler equation is how Heat Index values are calculated with temps below 80F (I tested this against the simple formula and got identical answers). While it is fair that we typically don't see Heat Index values below 80*F in the usual charts, this may just be not clear references ultimately, but is coded correctly here and matches the official method.
> 
> @nicolejkeeney Apologize for my confusing wording. @vicford Thank you so much for sharing the calculator, super helpful! I totally agree with you both that the implementation of the logic (use Rothfusz regression equation first, if it gives a value < 80, then replace it with simpler equation) is flawless. I had been trying to point out the sentence at the bottom of the reference "The Rothfusz regression is not valid for extreme temperature and relative humidity conditions beyond the range of data considered by Steadman." It seems to be referring to something different from the 80F logic. I just delved into Steadman's paper, and it seems Steadman considered temperature in the range of 20-50C, and so the Rothfusz regression equation is not valid when the temperature is outside of the range, all based on my understanding.

Back to the discussions on the data ranges, I was going to propose setting the index to NaN when the input variables are out of range, but the current implementation can definitely serve as A version of heat index with defensible reference. The reason is that the web calculator actually spits out results when extreme temperatures are given to it, even though the results might be meaningless according to the note below the calculator. Also considering the fact that the data ranges (in Steadman's paper and in the heat index chart below the web calculator) are somewhat arbitrary, plus the fact that there are more than one ways to calculate the index, I'm recognizing that the heat index doesn't enjoy an precise definition as variables like temperature do.

Tianchi-Liu:
Side note: Regardless of which heat index is presented to the users, I think depending on the user expectations, documentation and/or guidance might be appreciated by them, especially those who are interested in the extremes.

vicford:
In terms of next steps: following the notes that @Tianchi-Liu provided

> In practice, the Steadman formula is computed first and the result averaged with the temperature. If this heat index value is 80 degrees F or higher, the full regression equation along with any adjustment as described above is applied. If the ambient temperature is less the 40F (4.4C/277.65K), the heat index is set to to the ambient temperature.

Balanced by:

> Steadman considered temperature in the range of 20-50C, and so the Rothfusz regression equation is not valid when the temperature is outside of the range

20-50C is 68-122F

In short, an approach:
<40F: heat index = ambient air temperature
40-68F:  Simple equation, where Rothfusz eqn is not valid, but simple eqn also follows the UCAR process
68-80F: Simple equation, where Rothfusz eqn is also valid but not used
80+: Full regression equation, where Rothfusz eqn is valid

nicolejkeeney:
After the changes I just made, the function works as follows: 

1. The heat index is computed using the regression equation of Rothfusz (see [NWS documentation](https://www.wpc.ncep.noaa.gov/html/heatindex_equation.shtml))
```
HI = -42.379 + 2.04901523*T + 10.14333127*RH - .22475541*T*RH - .00683783*T*T - .05481717*RH*RH + .00122874*T*T*RH + .00085282*T*RH*RH - .00000199*T*T*RH*RH
```
2. For RH < 13% and 80F < T < 112F, HI regression equation is adjustment is **subtracted** from regression equation of Rothfusz
```
ADJUSTMENT = [(13-RH)/4]*SQRT{[17-ABS(T-95.)]/17}
```
3. For RH > 85% and 80F < T < 87, HI regression equation is adjustment is **added** to the regression equation of Rothfusz
```
ADJUSTMENT = [(RH-85)/10] * [(87-T)/5]
```
4. When the heat index from the regression equation of Rothfusz produces a value in the range 40F < HI < 80F, the simpler formula from Steadman is used instead 
```
HI = 0.5 * {T + 61.0 + [(T-68.0)*1.2] + (RH*0.094)}
```
5. When the heat index from the regression equation of Rothfusz results in a value **LESS** than 40F, the ambient temperature is used (following the [NCAR NCL documentation](https://www.ncl.ucar.edu/Document/Functions/Heat_stress/heat_index_nws.shtml))
```
HI = T
```

nicolejkeeney:
> Side note: Regardless of which heat index is presented to the users, I think depending on the user expectations, documentation and/or guidance might be appreciated by them, especially those who are interested in the extremes.
> 
> Happy to chat more if anything is unclear!

Excellent point, I improved the variable description in `app.select` and the function documentation!

nicolejkeeney:
Thanks @vicford and @Tianchi-Liu for this discussion, it's been interesting to follow and very helpful! Appreciate the thorough review.

nicolejkeeney:
> https://www.wpc.ncep.noaa.gov/html/heatindex.shtml

Good catch, I will fix that now!

nicolejkeeney:
> Thanks Nicole for improving the docs! One minor observation is that in the NCAR documentation, the 40F cutoff is based on the ambient temperature instead of the heat index. Updating the code would probably be easier than justifying the use of heat index.

Good catch, I'll fix that now :D

Tianchi-Liu (approved):

climakitae/indices.py
```diff
@@ -1,6 +1,80 @@
...
+    adj_lowT_highRH = ((RH - 85) / 10) * ((87 - T) / 5)  # Adjustment
+    HI_lowT_highRH = HI + adj_lowT_highRH  # Add adjustment from HI
+
+    # Use different equation if heat index if the heat index value < 80
+    low_HI = 0.5 * (T + 61.0 + ((T - 68.0) * 1.2) + (RH * 0.094))
+
+    # Adjust heat index depending on different condions for RH, T, and valid range of HI
+    HI = xr.where((RH < 13) & (T > 80) & (T < 112), HI_highT_lowRH, HI)
+    HI = xr.where(((RH > 85) & (T < 87) & (T > 80)), HI_lowT_highRH, HI)
+    HI = xr.where((HI < 80) & (HI > 40), low_HI, HI)
```
Tianchi-Liu, line 62: No need to check for HI > 40 per NOAA documentation.

PR #248, opened 2023-07-11
Add Effective Temperature as a derived index option

nicolejkeeney:
## Important note
This PR builds off the `fire` branch, which adds in the structure for derived indices. Please review [that PR](https://github.com/cal-adapt/climakitae/pull/234) before reviewing this one! 

## Description of this PR 
This PR adds effective temperature as a derived index for **daily WRF** data. To see this option, select "Dynamical" as your data type, and "daily" as your timescale. I will note that this derivation sets the first timestep to NaN-- that's not a bug! Since it requires data from yesterday, the first timestep will need to be NaN. 

## How to review 
Please review the derivation of the variable and confirm that the unit conversions and output are as expected (reference document [here](https://www.nationalgas.com/document/132516/download)). It should be in the same range as normal temperature values.


climakitae/data/variable_descriptions.csv
```diff
@@ -26,43 +26,44 @@ rainnc,Dynamical,hourly,mm,Precipitation (grid-scale portion only),"Any form of
...
+rsds,Statistical,"daily, monthly",W/m2,Shortwave flux at the surface,Shortwave radiation from the sun that reaches Earth's surface. ,ae_orange,TRUE
 fosberg_index_derived,Dynamical,hourly,[0 to 100],Fosberg fire weather index,"Fire weather index using meteorological inputs (temperature, relative humidity, windspeed) to quantify fire risk. Large values imply a higher risk value.",ae_orange,TRUE
-noaa_heat_index_derived,Dynamical,hourly,degF,NOAA Heat Index,"A measure of what the the temperature ""feels like"" to the human body. Derived from relative humidity and air temperature. ",ae_orange,TRUE
\ No newline at end of file
+noaa_heat_index_derived,Dynamical,hourly,degF,NOAA Heat Index,"A measure of what the the temperature ""feels like"" to the human body. Derived from relative humidity and air temperature. ",ae_orange,TRUE
+effective_temp_index_derived,Dynamical,daily,K,Effective Temperature,The sum of half of yesterday's daily temperature and half of the current day's temperature. Accounts for previous day's temperature due to consumer behaviour and perception of the weather. ,ae_orange,TRUE
```
vicford, line 69: british spelling here :)

climakitae/indices.py
```diff
@@ -4,6 +4,40 @@
 import numpy as np
 
 
+def effective_temp(T):
```
vicford, line 7: Based on my understanding of this metric, I think the "yesterday" temp needs to be the "yesterday effective" temp, which confirms that yes the first time step will be nan, but to me it looks like the T_yesterday is just grabbing yesterday's temperature _not_ yesterday's effective temperature, if that makes sense

vicford (changes requested):
Out of curiosity, would like more info on how this is used in applications by stakeholders (not a Nicole question though)!

climakitae/indices.py
nicolejkeeney: Hmmm that's not actually how I understood the equation. I wasn't able to find any additional documentation on the metric though. It's described in the [reference doc](https://www.nationalgas.com/document/132516/download) Owen sent me this way: 
> The effective temperature is half of yesterday’s temperature added to half of today’s actual
temperature.

nicolejkeeney: Ooooo you're right! Good catch. I'll update it now
From the reference doc: 
Effective temperature (0.5 * today’s temperature + 0.5 * yesterday’s effective)

vicford (approved):
looks good to me now :)

Tianchi-Liu:
If I'm reading the code correctly, effective temperature is computed as 1/2 today's temperature + 1/4 yesterday's temperature + 1/4 the day before yesterday's temperature, and it is different from my understanding of the effective temperature from the reference document.

Expanding the defining equation of effective temperature gives $1/2  T_t + 1/2 ET_{t-1} = 1/2 T_t + 1/2 (1/2 T_{t-1} + 1/2 ET_{t-2}) = 1/2 T_t + 1/4 T_{t-1} + 1/8 T_{t-2} + 1/16 T_{t-3} + ...$ 
where T = temperature, ET = effective temperature, and the subscripts represent time steps (days) with t being today, t-1 being yesterday etc. 

To make sense of the expansion, today's ET is consist of a portion of the actual temperature of each day up to today--half of today's temp, 1/4 of yesterday's temp, 1/8 of the day before yesterday's temp etc, thus it's "an exponentially smoothed temperature" as stated in the glossary of the reference.

By definition the terms in the equation go back in time infinitely, which makes it impossible to implement, but considering the portion of each previous day's temperature that contributes to the ET decays exponentially, it's reasonable to me to disregard days that are so far back in time that their contributions are trivial. Maybe also consider the desired level of accuracy and technical constraints when deciding how to go about implementation.

nicolejkeeney:
> If I'm reading the code correctly, effective temperature is computed as 1/2 today's temperature + 1/4 yesterday's temperature + 1/4 the day before yesterday's temperature, and it is different from my understanding of the effective temperature from the reference document.
> 
> Expanding the defining equation of effective temperature gives 1/2Tt+1/2ETt−1=1/2Tt+1/2(1/2Tt−1+1/2ETt−2)=1/2Tt+1/4Tt−1+1/8Tt−2+1/16Tt−3+... where T = temperature, ET = effective temperature, and the subscripts represent time steps (days) with t being today, t-1 being yesterday etc.
> 
> To make sense of the expansion, today's ET is consist of a portion of the actual temperature of each day up to today--half of today's temp, 1/4 of yesterday's temp, 1/8 of the day before yesterday's temp etc, thus it's "an exponentially smoothed temperature" as stated in the glossary of the reference.
> 
> By definition the terms in the equation go back in time infinitely, which makes it impossible to implement, but considering the portion of each previous day's temperature that contributes to the ET decays exponentially, it's reasonable to me to disregard days that are so far back in time that their contributions are trivial. Maybe also consider the desired level of accuracy and technical constraints when deciding how to go about implementation.

Awesome, thanks @Tianchi-Liu . This is an excellent point. I just updated the code to consider 4 days in the past. Let me know if the derivation matches the documentation better now!

Tianchi-Liu:
> > If I'm reading the code correctly, effective temperature is computed as 1/2 today's temperature + 1/4 yesterday's temperature + 1/4 the day before yesterday's temperature, and it is different from my understanding of the effective temperature from the reference document.
> > Expanding the defining equation of effective temperature gives 1/2Tt+1/2ETt−1=1/2Tt+1/2(1/2Tt−1+1/2ETt−2)=1/2Tt+1/4Tt−1+1/8Tt−2+1/16Tt−3+... where T = temperature, ET = effective temperature, and the subscripts represent time steps (days) with t being today, t-1 being yesterday etc.
> > To make sense of the expansion, today's ET is consist of a portion of the actual temperature of each day up to today--half of today's temp, 1/4 of yesterday's temp, 1/8 of the day before yesterday's temp etc, thus it's "an exponentially smoothed temperature" as stated in the glossary of the reference.
> > By definition the terms in the equation go back in time infinitely, which makes it impossible to implement, but considering the portion of each previous day's temperature that contributes to the ET decays exponentially, it's reasonable to me to disregard days that are so far back in time that their contributions are trivial. Maybe also consider the desired level of accuracy and technical constraints when deciding how to go about implementation.
> 
> Awesome, thanks @Tianchi-Liu . This is an excellent point. I just updated the code to consider 4 days in the past. Let me know if the derivation matches the documentation better now!

Better indeed!

Tianchi-Liu (approved):
As Nicole pointed out, the reference doesn't document the number of days to consider, and four seems reasonable. I think the derivation is good to merge. Notes for future: This is something that can be tweaked as needed for different use cases. Another note for future: To forecast gas demand, using a gas day (5/6am - 5/6am) might be considered according to the reference.

Tianchi-Liu:
Some other observations apart from the derivation:
- The NOAA heat index seems not available in deg C and K, but the two units are available as options.
- Seen in the screenshot below, the units of x and y coordinates are off and the title is cut off. These are also observed on relative humidity maps.
Screenshot 2023-08-06 120651

nicolejkeeney:
> * Seen in the screenshot below, the units of x and y coordinates are off and the title is cut off. These are also observed on relative humidity maps.

The units issue is due to an underlying issue with the relative humidity derivation, which is going to be fixed with [this PR](https://github.com/cal-adapt/climakitae/pull/264) :D

nicolejkeeney:
> * The NOAA heat index seems not available in deg C and K, but the two units are available as options.

Good catch. Just fixed this!

PR #249, opened 2023-07-11
Fix uncertainity utils renamed function call

elehmer:
This PR fixes broken uncertainity utils function. Uncertainity utils calls a renamed function in data_loaders that does not exist anymore.


nicolejkeeney (approved):
thanks for fixing this!

elehmer:
What other functions got renamed? Should search the codebase to see if they are referenced

PR #252, opened 2023-07-14
Add singleton 'scenario' dimension for 'historical climate'

thenaomig:
This should fix the issue in the AMY panel that popped up after the data_loaders refactor: seems to require that there be a singleton dimension for scenario if 'Historical Climate' is present without any SSP. This PR adds one.

### DIFF
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -479,6 +479,7 @@ def _merge_all(selections, data_dict, cat_subset):
         )
     else:
         all_ssps = all_hist
+        all_ssps = all_ssps.assign_coords({"scenario": selections.scenario_historical})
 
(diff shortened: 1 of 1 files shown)


nicolejkeeney (changes requested):
Doesn't seem to work for me-- still not seeing the scenario dimension/coordinate. Just using the default setting for `app.select()`. Also getting an error still from `app.explore.amy()`

Screenshot 2023-07-14 at 9 02 26 AM

thenaomig:
Previously scenario was added as a coordinate but now it is a dimension too as desired.

nicolejkeeney (approved):
Your recent commit fixed the issue, thanks!

vicford (approved):
Works for me

PR #253, opened 2023-07-18
Station range for consistency.

thenaomig:
The bias-correction already uses the same range in the gridded data as the station data that was read-in. This changes that reference range in one other spot for consistency, where we make sure the gridded data selected is expanded to include the same range as the station obs.


bethem (approved):
lgtm

PR #238, opened 2023-06-09
Bigger refactor of data_loaders.py

thenaomig:
A bigger refactor of data_loaders.py -- 
- replacing `_process_and_concat` and `_subset` with a series of smaller functions
- fewer loops are present in building the multi-dimensional xr.DataArray 
- steps are more modular and testable

PR #254, opened 2023-07-18
Restored retrieval of reanalysis-driven WRF run

thenaomig:
This PR fixes degradation introduced in https://github.com/cal-adapt/climakitae/pull/238, after which 'retrieve' failed to return anything for "Historical Reconstruction" (e.g. the reanalysis-driven WRF simulation). 

The structure of the xr.DataArray in which the ERA5_WRF data is returned is different from before when "Historical Climate" and "Historical Reconstruction" are both selected. The "ERA5_WRF" label is appended along the "simulation" dimension, but the scenario dimension is labeled "Historical" to avoid the inefficiency of a sparse array with two values along the 'scenario' dimension.

### DIFF
--- a/climakitae/data_loaders.py
+++ b/climakitae/data_loaders.py
@@ -115,7 +115,7 @@ def _sim_index_item(ds_name, member_id):
     downscaling_type = ds_name.split(".")[0]
     gcm_name = ds_name.split(".")[2]
     ensemble_member = str(member_id.values)
-    if ensemble_member != 'nan':
+    if ensemble_member != "nan":
         return "_".join([downscaling_type, gcm_name, ensemble_member])
     else:
         return "_".join([downscaling_type, gcm_name])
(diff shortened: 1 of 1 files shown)


vicford (changes requested):
Retrieval of the reconstruction data works as desired, but I have mixed feelings about the differing scenario names based on which combination of data is being selected, and am wondering if we should streamline it further to have just "Historical" as the scenario name regardless of which data product is selected. Potentially a way to ensure that it is clear when both simulation + reanalysis data are selected which is the reanalysis is to append "_Reconstruction" to "WRF_ERA5", but I am aware that will cause other issues within the backend. This doesn't preclude the retrieval piece being merged in though.

thenaomig:
> When both "Historical Climate" and "Historical Reconstruction" are selected, only Historical Climate is returned. The output data here should have two scenarios, not just one.
> 
> only historical climate returned

Yes, this is on purpose. I am changing how it is returned, so that it's not a sparse array. Is there a notebook or other tool that relies on it being the other way?

thenaomig:
Re-separated "Historical Climate" and "Historical Reconstruction" on scenario dimension to match prior behavior. In future may separate these so they are never returned in one merged object.

Another small fix for when ssps are selected with fewer simulations, so only the relevant historical simulation is returned.

nicolejkeeney (approved):
Looks good to me!

vicford (approved):

PR #255, opened 2023-07-21
Fixes to code to jump to 2023-07-05 pangeo image

elehmer:
This PR fixes the code to run on the 2023-07-05 pangeo-notebook image.

To test use in conjunction with `upgd_pangeo` branch of `cae-notebooks` and on local instance of pangeo running `2023-07-05` image


nicolejkeeney (approved):
Looks good to me!

elehmer:
[Fixes needed to update to 2023-07-05 pangeo image](https://trello.com/c/FbUnhQrS/435-fixes-needed-to-update-to-2023-07-05-pangeo-image)

vicford (commented):
Getting these warnings (could be my local?), but doesn't preclude any notebooks running successfully

```
<jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
<jemalloc>: (This is the expected behaviour if you are running under QEMU)
```

elehmer:
> Getting these warnings (could be my local?), but doesn't preclude any notebooks running successfully
> 
> ```
> <jemalloc>: MADV_DONTNEED does not work (memset will be used instead)
> <jemalloc>: (This is the expected behaviour if you are running under QEMU)
> ```

Does this show in the console when loading the docker image?

vicford:
> This is in notebook at the import climakitae as ck stage for me, independent of notebook tested

@elehmer, just tested on the old image locally, and this warning also occurs for me there too, so this might be a my computer thing and not the image update, although I'm not sure why that would be, but did want to provide more context too in testing

nicolejkeeney (approved):
everything works well for me

elehmer:
After lots of errors and fiddling with the test environment I think we are ready to go. I ended up having to downgrade `scipy==1.10.1` from `1.11.1` to deal with distribution function error in the thresholds test.

PR #256, opened 2023-07-21
Add unit tests for derived indices

nicolejkeeney:
This PR adds unit tests for the three new derived indices: effective temperature, fosberg fire index, and the NOAA heat index. Since this code builds off the derived index code, it's branched off of the branch `eft` instead of `main`. 

I also deleted an old file that was used to build one of the existing test datasets used in other unit tests, since the code is nonfunctional now-- climakitae has changed too much. 

This PR also adds two new testing datasets, one with daily data and the other with hourly data. The files are small since I used a small spatial subset and time slice. I also added the scripts I used to generate these files. 

**To test**: Just take a quick look through the changed files. GitHub will automatically check if the tests run without error.


tests/conftest.py
```diff
@@ -18,3 +18,21 @@ def test_data_2022_monthly_45km(rootdir):
     filepath = os.path.join(rootdir, filename)
     ds = xr.open_dataset(filepath)
     return ds
+
+
+@pytest.fixture
```
vicford, line 23: I am wondering as to whether it is useful to include a short one-line comment as to roughly what each of these test datasets represent (maybe not in this script?) since we now have 3 test datasets, could be useful for something like:

test_dataset_Jan2015_LAcounty_45km_daily: daily data for LA county at 45 km for Jan 2015, useful for daily index testing
test_dataset_01Jan2015_LAcounty_45km_hourly: hourly data for LA county at 45 km for 1 Jan 2015, useful for hourly index testing

vicford (commented):
Just one comment as to semi-documentation, but otherwise looks good!

tests/conftest.py
nicolejkeeney: Agreed, thanks!

PR #257, opened 2023-07-24
Refactor of entire codebase

elehmer:
I have adopted suggestions by @thenaomig so the syntax has been shortened. Revised are now below:

This is my beginning of attempt to reorganize and refactor the code...please comment!!

These are the starting classes:
```
import climakitae as ck
selections = ck.Select()
selections.show()
```

Then you can see the `DataParameters` properties:

```
selections._data_catalog
selections.latitude
```

You can set the param properties this way and you can then retrieve the data using:

```
selections.area_subset = 'states'
selections.cached_area = 'CA'
data  = selections.retrieve()
```
You can also retrieve data using only the DataParameters class (attached to class) so that panel can be bypassed.

You can now load data and view data and export data:

```
data = ck.load(data)
ck.view(data)

ck.export("NetCDF", data, 'test_file')
```

Explore can now be used:

```
from climakitae.explore import warming_levels
wl = warming_levels()
wl.show()

from climakitae.explore import thresholds
thres = thresholds()
thres.show()

from climakitae.explore import amy
amy = amy()
amy.show()
```

TimeSeries is accessible this way:

```
import climakitae.explore.timeseries as tst
ts = tst.TimeSeries(data)
ts.explore()
```

Every one of these tools uses a new instance of Parameters so that changing parameters in one won't affect the others


climakitae/core/data_interface.py
```diff
@@ -0,0 +1,799 @@
...
+from climakitae.util.unit_conversions import _get_unit_conversion_options
+from climakitae.core.catalog_convert import (
+    _downscaling_method_to_activity_id,
+    _resolution_to_gridlabel,
+    _timescale_to_table_id,
+    _scenario_to_experiment_id,
+)
+
+
+class DataInterface:
```
elehmer, line 27: I will make this more abstract and not have it load the data and call it on package import I guess. Then it will load data when data is requested...

elehmer:
@thenaomig @bkg I have made real progress. I have made a singleton class `DataInterface` for the data connection (it just loads the data files on init as it isn't that long) and there is now a `DataParameters` class that calls `DataInterface` on init. This has all the param objects. The select panel is called by the `Select` class that piggybacks on `DataParametersWithPanes` which is an extension of `DataParameters` with the panel panes. Data then can be retrieved using the selections in `DataParameters` and it can be called either from `DataParameters` or `Select`.

climakitae/core/constants.py
```diff
@@ -0,0 +1,18 @@
+# constants.py
+"""This module defines package level constants"""
```
nicolejkeeney, line 2: Big fan of this reorganizing! The constants module is a great idea

elehmer: > Big fan of this reorganizing! The constants module is a great idea

Thanks! Yeah, `pkg_resource` is deprecated and going away...

nicolejkeeney:
Is it possible with the current structure to get an output that shows all the current settings in the `selections` object? Previously calling `app.selections` would show you all the params and their current selections, but the selections object doesn't do that. 
selections

nicolejkeeney:
> data_parameters

Answering my own question here: I am seeing that this is possible with this call: `selections.data_parameters`

selections

nicolejkeeney:
This syntax is a huge improvement in my opinion, it makes a lot more sense to retrieve from "selections" than from the Application object. Nice!
```data  = selections.retrieve()```

**Comment by elehmer** [comment]
First sweep through the code is complete! The test work!! I will be going through again and seeing where things can be cleaned up some more.

**Comment by elehmer** [comment]
[REFACTOR entire codebase](https://trello.com/c/hxTayZEt/445-refactor-entire-codebase)

File: climakitae/ui/select.py [hunk 3]
```diff
@@ -0,0 +1,212 @@
...
+import matplotlib.ticker as ticker
+import cartopy.crs as ccrs
+import cartopy.feature as cfeature
+
+from climakitae.core.data_interface import DataParametersWithPanes
+
+
+class Select:
+    def __init__(self):
+        self.data_parameters = DataParametersWithPanes()
```
thenaomig, line 13: I might call this object 'selections' instead of 'data_parameters' for greater consistency with the old version.

thenaomig: Why can't Select inherit from DataParametersWithPanes instead actually?

thenaomig:
In terms of syntax, 
1. it might be nice to wrap those object instantiations and the .show() so that the panel displays with just one line in a notebook
2. `selections.data_parameters.[some_item] =` is a lot for every single assignment. If the Select class inherited from DataParametersWithPanes (as the explore panels do), then we could have `selections.[some_item] =` instead

thenaomig (changes requested):
Just a couple structural thoughts for now. Let's discuss.

elehmer:
> In terms of syntax,
> 
> 1. it might be nice to wrap those object instantiations and the .show() so that the panel displays with just one line in a notebook
> 2. `selections.data_parameters.[some_item] =` is a lot for every single assignment. If the Select class inherited from DataParametersWithPanes (as the explore panels do), then we could have `selections.[some_item] =` instead
#1 I would prefer not to so that properties can be adjusted without the GUI or before you show the GUI.
#2 Makes a lot of sense. I had an initial issue with class inheritance planning, but now it works great!

climakitae/ui/select.py
elehmer: It can! Like I said I initially ran into some conceptual issues but looks like now it works well! I will make the changes for this.

climakitae/core/data_interface.py
```diff
@@ -751,8 +601,19 @@ def __init__(self, **params):
         # Set default values
         super().__init__(**params)
 
+        self.data_interface = DataInterface()
```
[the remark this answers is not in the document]
elehmer: I will look into making these immutable so that they can't be messed with.

climakitae/ui/select.py
elehmer: > I might call this object 'selections' instead of 'data_parameters' for greater consistency with the old version.

I will do this

elehmer: > I might call this object 'selections' instead of 'data_parameters' for greater consistency with the old version.

This will go away with subclassing the GUI classes to the Parameter classes.

elehmer:
@thenaomig I have incorporated your suggestions for  subclassing the GUIs to the Params so that `data_parameters` is no longer needed. I also prevent `DataInterface` attributes from being easily changed by using the `@property` decor.

elehmer:
I have updated the branch to move the code and incorporate export PR

PR #260, opened 2023-07-27
Export file options for TMY data

vicford:
Adds two new fancy formats for how TMY data is used: .epw (energy plus weather) and .tmy (typical meteorological year)
These formats are effectively csv files with specific headers and information included. Sources are included in the header construction functions. 

---
To test:
Run through the AMY_to_TMY notebook in the amy_to_tmy branch on cae-examples (https://github.com/cal-adapt/cae-examples/pull/47) and play around with the file_ext as both "tmy" and "epw" in the export code cell (the very last one)

### DIFF
--- a/climakitae/data_export.py
+++ b/climakitae/data_export.py
@@ -477,9 +477,9 @@ def _tmy_header(location_name, df):
     )
 
     # line 2 - data field name and units, manually setting to ensure matches TMY3 labeling
-    # tmy dataframe headers handles this
+    line_2 = "Air Temperature at 2m (degC),Dew point temperature (degC),Relative humidity (%),Instantaneous downwelling shortwave flux at bottom (W m-2),Shortwave surface downward diffuse irradiance (W m-2),Instantaneous downwelling longwave flux at bottom (W m-2),Wind speed at 10m (m s-1),Wind direction at 10m (deg),Surface Pressure (Pa)\n"
 
-    headers = [line_1]
+    headers = [line_1, line_2]
 
     return headers
 
@@ -513,7 +513,7 @@ def _epw_header(location_name, df):
     )
 
     # line 7 - comments 2, putting the data variables here manually as they are not specified in epw format, and we are not including all
-    line_7 = "COMMENTS 2,Air Temperature at 2m,Dew point temperature,Relative humidity,Instantaneous downwelling shortwave flux at bottom,Shortwave surface downward diffuse irradiance,Instantaneous downwelling longwave flux at bottom,Wind speed at 10m,Wind direction at 10m,Surface Pressure\n"
+    line_7 = "COMMENTS 2,Air Temperature at 2m (degC),Dew point temperature (degC),Relative humidity (%),Instantaneous downwelling shortwave flux at bottom (W m-2),Shortwave surface downward diffuse irradiance (W m-2),Instantaneous downwelling longwave flux at bottom (W m-2),Wind speed at 10m (m s-1),Wind direction at 10m (deg),Surface Pressure (Pa)\n"
 
     # line 8 - data periods, num data periods, num records per hour, data period name, data period start day of week, data period start (Jan 1), data period end (Dec 31)
     line_8 = "DATA PERIODS,1,1,Data,,1/1,12/31\n"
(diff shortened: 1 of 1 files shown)


climakitae/data_export.py
```diff
@@ -457,3 +457,109 @@ def _rchop(s, suffix):
             ):
                 f.write(str(att_keys) + " : " + str(att_values))
                 f.write("\n")
+
+
+## TMY export functions
+def tmy_header(station_name, df):
+    """
```
nicolejkeeney, line 464: Can you rename this as `_tmy_header` since it's an internal (not user-facing) function?

climakitae/data_export.py
```diff
@@ -457,3 +457,109 @@ def _rchop(s, suffix):
...
+    # tmy dataframe headers handles this
+
+    headers = [line_1]
+
+    return headers
+
+
+def epw_header(station_name, df):
+    """
+    Constructs the header for the TMY output file in .epw format
```
nicolejkeeney, line 489: Same for this one `_epw_header`

climakitae/data_export.py
```diff
@@ -457,3 +457,109 @@ def _rchop(s, suffix):
...
+    # line 8 - data periods, num data periods, num records per hour, data period name, data period start day of week, data period start (Jan 1), data period end (Dec 31)
+    line_8 = "DATA PERIODS,1,1,Data,,1/1,12/31\n"
+
+    headers = [line_1, line_2, line_3, line_4, line_5, line_6, line_7, line_8]
+
+    return headers
+
+
+def write_tmy_file(filename_to_export, df, station_name, file_ext="tmy"):
+    """Exports TMY data either as .epw or .tmy file
```
nicolejkeeney, line 527: I think it would be better if `station_name` was an optional argument. Perhaps you could rename it to `location`. Presumably, you could generate a TMY file for a gridcell and not just a weather station. Or, generate a TMY for a point that wasn't a weather station. To be clear, my suggestion is changing `station_name` to `location_name = "location"` or something like that, making it an optional argument.

climakitae/data_export.py
```diff
@@ -457,3 +457,109 @@ def _rchop(s, suffix):
...
+        with open(path_to_file, "w") as f:
+            f.writelines(epw_header(station_name, df))  # writes required header lines
+            df = df.drop(
+                columns=["simulation", "lat", "lon", "scenario"]
+            )  # drops header columns from df
+            dfAsString = df.to_csv(sep=",", header=False, index=False)
+            f.write(dfAsString)  # writes file
+
+    else:
+        print('Please pass either "tmy" or "epw" as a file format for export.')
```
nicolejkeeney, line 565: It might be nice if the function also printed something if the file was successfully exported, i.e 
" TMY file exported to .epw format with filename blahblablah.epw"

climakitae/data_export.py
```diff
@@ -457,3 +457,109 @@ def _rchop(s, suffix):
...
+    filename_to_export (str): Filename string, constructed with station name and simulation
+    df (pd.DataFrame): Dataframe of TMY data to export
+    station_name (str): Station name string
+    file_ext (str, optional): File extension for export, default is .tmy, options are "tmy" and "epw"
+
+    Returns
+    -------
+    None
+    """
+
```
nicolejkeeney, line 540: Possible to add a sanity check here to ensure that the user passes it a DataFrame object? Just something like: 
```
if type(df) != pd.DataFrame: 
    raise ValueError("The function requires a pandas DataFrame object as the data input") 
```
I tried feeding it an xarray object (I assume this might be a somewhat common mistake) and the error the function raised was hard to interpret, so catching that error at the beginning of the function would be cool!

nicolejkeeney (changes requested):
This looks really good, I left a few minor comments for improvement. I really like how you added the data attributes to the epw header.

nicolejkeeney:
I reallllyyy like this part, but is there a way to incorporate units into the comments part for each variable? 
Screenshot 2023-07-28 at 1 59 48 PM

climakitae/data_export.py
vicford: Good call, done!

climakitae/data_export.py
vicford: I agree that `station_name` should be renamed because we do provide the option to have the user select their own location, however TMY/EPW files include the location in the filename, as well as in the header info within the file, which we are matching, so it's not immediately obvious to me how to pass that information by making it optional

nicolejkeeney (approved):
Looks great, thanks!

PR #261, opened 2023-07-28
Limit relative humidity to < 0.5%

nicolejkeeney:
Our derivation was resulting in some really low RH values. RH for CA doesn't go below 0.8. We set the limit here to 0.5. 

**To test:** Try retrieving the relative humidity data. Make sure it looks appropriate.


vicford:
Getting this error on testing in the AMY_to_TMY notebook (most obvious testing location for me) for grabbing RH data, specifically when trying to retrieve data via `get_closest_gridcell`:

`CRSError: Invalid projection: : (Internal Proj Error: proj_create: unrecognized format / unknown name)
`

vicford (approved):
Better, thanks!

PR #262, opened 2023-07-31
Refresh the current documentation. Fix broken environment.yml file

nicolejkeeney:
The readthedocs page doesn't automatically update (it should!! see [this](https://trello.com/c/jTwpvNkP/347-troubleshoot-readthedocs-issue) open trello card). So, here I force-refreshed the documentation to update it with the current code. I branched off of my effective temp branch (eft) so that the index code was included in this. 

I also noticed an issue when trying to rebuild the conda environment, which is used to build the documentation. Just some minor formatting issues, which I easily resolved :D


vicford (approved):
Updated just fine for me, can see indices in there as well no problems

PR #263, opened 2023-07-31
Make plot object a panel Pane object in app.view()

nicolejkeeney:
Troubleshooting the app.view() reloading issues, here I converted the hvplot image returned by `app.view` into a `panel.Pane` object. Let me know if this resolves the issue for you!


vicford (approved):
YAY this works!!

PR #264, opened 2023-08-03
Switches the relative humidity equation

vicford:
Nicole and I noticed some unrealistically, and quantitatively incorrect, high relative humidity values in several applications (AMY; heat index). This PR switches the equation used to address this, as well as sets lower (0.5) and upper (100) limits on the relative humidity results.

To test, retrieve RH and ensure that values look good


thenaomig:
I don't see any reason not to merge this.

PR #265, opened 2023-08-04
Add window as optional argument for warming_levels

nicolejkeeney:
This allows the user to customize the time window they want to generate around the warming levels date. The default is set to 15 years, so no changes are to be expected in the existing functionality. 

### To test 
Make sure the function `get_anomaly_data` works as expected, using the `windows` argument to test that the window changes.


thenaomig:
Looks great, thanks!

PR #278, opened 2023-08-31
Add climate variable names to CSV exported from station data

Tianchi-Liu:
As promised in https://github.com/cal-adapt/climakitae/pull/270, this PR adds to the headers of exported CSV tables, the names of climate variables associated with the weather stations. 

**E.g.** "Air Temperature at 2m" in the 2nd row of the table below.
exported_data_header

### DIFF
--- a/climakitae/core/data_export.py
+++ b/climakitae/core/data_export.py
@@ -7,6 +7,11 @@ import datetime
 import xarray as xr
 import pandas as pd
 from importlib.metadata import version as _version
+from climakitae.util.utils import read_csv_file
+from climakitae.core.paths import (
+    variable_descriptions_csv_path,
+    stations_csv_path
+)
 
 xr.set_options(keep_attrs=True)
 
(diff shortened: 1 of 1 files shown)


vicford (approved):
Works just fine for me!

Minor comment: Thoughts on making the "CSV" argument passed to the export function lowercase, or case-insensitive? While the abbreviation CSV is uppercase, I would imagine most users would think of .csv files as you do in a file browser with the lowercase extension version

gdicecco (commented):
This looks good to me as well!

Tianchi-Liu:
> Works just fine for me!
> 
> Minor comment: Thoughts on making the "CSV" argument passed to the export function lowercase, or case-insensitive? While the abbreviation CSV is uppercase, I would imagine most users would think of .csv files as you do in a file browser with the lowercase extension version

Great idea, that would improve user experience. I've added this to my to-dos.

PR #280, opened 2023-09-06
warming levels upgrade

thenaomig:
This PR does several things:

- updates the code to generate the global warming level look-up tables, for the year at which a given simulation reaches a given global warming level -- now in it's own file in the utils folder
- updates the look-up table files themselves
- updates the warming levels class to create separate panels for initial selection and subsequent exploration:
>  from climakitae.explore import warming_levels
>  wl = warming_levels()
>  wl.choose_data()
>  wl.calculate()
>  wl.visualize()

- adds LOCA2 data as an option (monthly 3km LOCA data for Alameda Co takes ~5min on the calculate step)
- you can extract the data centered around warming levels, or the postage_stamp map data

> regional_response = wl.gwl_snapshots
> gwl_slices = wl.sliced_data

- anomaly not required for LOCA2 (because it's bias-corrected)
- stats tab in visualize panel shows the individual simulation based on min/max/median spatial average


climakitae/core/data_interface.py
```diff
@@ -1174,7 +1174,12 @@ def retrieve(self, config=None, merge=True):
                 raise ValueError(
                     "To retrieve data specified in a configuration file, please input the path to your local configuration csv as a string"
                 )
-        return read_catalog_from_select(self)
+        try:
```
thenaomig, line 1177: this check was present in the warming levels panel, but it seemed like it belonged in the more general case

vicford:
Upon comparison to the older warming levels notebook, seeing a discrepancy in what range of values is produced in that part 2 example that I think is worth checking

Old version:
Screenshot 2023-09-13 at 3 57 56 PM

New version:
Screenshot 2023-09-13 at 3 50 06 PM
- seeing CESM2_r11i1p1f1 3 times?
- 45 deg range is.... alarming

thenaomig:
> Upon comparison to the older warming levels notebook, seeing a discrepancy in what range of values is produced in that part 2 example that I think is worth checking
> 
> ETA: old version is with 3° warming, new is with 2°
> 
> Old version: Screenshot 2023-09-13 at 3 57 56 PM
> 
> New version: Screenshot 2023-09-13 at 3 50 06 PM
> 
>     * seeing CESM2_r11i1p1f1 3 times?
> 
>     * 45 deg range is.... alarming

The reason that CESM2 appears three times is that it now includes the other SSPs. I agree about the magnitudes... degrees Fahrenheit does not account for that.

thenaomig:
> Upon comparison to the older warming levels notebook, seeing a discrepancy in what range of values is produced in that part 2 example that I think is worth checking
> 
> ETA: old version is with 3° warming, new is with 2°
> 
> Old version: Screenshot 2023-09-13 at 3 57 56 PM
> 
> New version: Screenshot 2023-09-13 at 3 50 06 PM
> 
>     * seeing CESM2_r11i1p1f1 3 times?
> 
>     * 45 deg range is.... alarming

This is an issue with order-of-operations in the notebook, not a climakitae issue. Will push a change to cae-notebooks.

vicford (approved):
I definitely like the new split of functions! I do think there needs to be some additional markdown/guidance on some of the new features, but does not preclude this being merged in and implemented. 

The sliced_data example is really clear to me

climakitae/explore/warming.py
```diff
@@ -31,179 +40,223 @@
...
+    - postage stamps from visualize "main" tab are accessible via "gwl_snapshots"
+    - data sliced around gwl window retrieved from "sliced_data"
     """
-    This function pulls data from the catalog and reads it into memory
 
-    Parameters
-    ----------
-    self: WarmingLevelParameters
-        object holding user's selections
+    catalog_data = xr.DataArray()
```
claalmve, line 52: Is there a reason for including these variables outside of the `init` method?

thenaomig: Not necessarily... that's where params are defined but these ones don't have to be.

thenaomig: (feel free to change it yourself)

PR #284, opened 2023-09-20
updated cmap update for redesigned warming

vicford:
1. Colormaps for variables should be ae_diverging instead of their assigned colormap according to the data_descriptions file, with the exception of air temperature and dew point temperature. This fixes an issue Naomi had noticed specifically for precipitation, but upon testing was relevant for all other vars.
2. The assignment to humidity to 100% when displayed had a typo, now fixed.

----

To test:
Run the warming_levels notebook and make sure that:
- air temp and dew point have ae_orange colormaps
- relative humidity is on a 100% scale (range is not -5 - 5)
- any other variable selection has a diverging colormap

### DIFF
--- a/climakitae/explore/warming.py
+++ b/climakitae/explore/warming.py
@@ -172,13 +172,12 @@ def get_sliced_data(y, years, window=15, anom="Yes"):
 
 def _get_cmap(wl_params):
     """Set colormap depending on variable"""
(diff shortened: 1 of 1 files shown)


elehmer (approved):

claalmve (approved):
Looks good to me!

PR #285, opened 2023-09-21
Update HadISD station names

vicford:
Updates the station names in the hadisd_stations.csv to include their ICAO as well as removing the / character in two of the new stations


claalmve:
Looks good to me! Tested with Select and data was properly imported. I think we will eventually need to expand the Select panel to not have the border overlap with station names but that is definitely non-urgent.

claalmve (approved):

PR #287, opened 2023-10-09
GWL Upgrade

claalmve:
**Back-end changes to Updated GWL notebook**

Making changes to GWL framework updated notebook. Main changes include separating warming levels individually, so that data could be properly sliced into different 30-year chunks in `sliced_data`.

To test, please run the corresponding `updated_warming_levels.ipynb` from cae-notebooks.

### DIFF
--- a/climakitae/explore/warming.py
+++ b/climakitae/explore/warming.py
@@ -835,9 +835,9 @@ def warming_levels_visualize(wl_viz):
     postage_stamps_MAIN = pn.Column(
         pn.widgets.StaticText(
             value=(
-                "Panels show the 30-year average centered on the year that each"
-                "GCM run (each panel) reaches the specified warming level."
-                "If you selected 'Yes' to return an anomaly, you will see the difference"
+                "Panels show the 30-year average centered on the year that each "
+                "GCM run (each panel) reaches the specified warming level. "
+                "If you selected 'Yes' to return an anomaly, you will see the difference "
                 "from average over the 1981-2010 historical reference period."
             ),
             width=800,
--- a/climakitae/explore/warming.py
+++ b/climakitae/explore/warming.py
@@ -60,7 +60,6 @@ class WarmingLevels:
     def choose_data(self):
         return warming_levels_select(self.wl_params)
 
-    # @dask.delayed
     def find_warming_slice(self, level, gwl_times):
         """
         Find the warming slice data for the current level from the catalog data.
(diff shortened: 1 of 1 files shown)


climakitae/explore/warming.py
```diff
@@ -74,25 +102,29 @@ def calculate(self):
...
-        self.sliced_data["all_sims"] = relabel_axis(self.sliced_data.all_sims)
-        self.gwl_snapshots = self.sliced_data.reduce(np.nanmean, "time")
-        self.gwl_snapshots = self.gwl_snapshots.compute()
+            self.gwl_times = read_csv_file(gwl_1850_1900_file, index_col=[0, 1, 2])
+        self.gwl_times = self.gwl_times.dropna(how="all")
+        self.catalog_data = clean_list(self.catalog_data, self.gwl_times)
+
+        self.sliced_data = {}
+        self.gwl_snapshots = {}
+        for level in self.warming_levels:
```
claalmve, line 113: This area is where the run-time significantly slows down. This changes run-time from around ~20 min. to now ~45 min., and it has to do with this for-loop being called on each of the individual warming levels. For this first iteration, I figured it would be okay to leave the code as-is ("defined as done"), and in a separate PR later, work on parallelizing this more elegantly.

Just a bit tricky with @dask.delayed calls and/or rechunking. For some reason, any calls I make with @dask.delayed don't seem to work. Happy to hop on a call to discuss ways for dealing with this too.

climakitae/explore/warming.py
```diff
@@ -696,6 +632,183 @@ def warming_levels_select(self):
...
+                + ": for "
+                + str(warmlevel)
+                + "°C Warming Across Models"
+            )  # Add title
+            warm_level_dict[warmlevel] = all_plots.cols(1)
+        # else:
+        # return None
+    return warm_level_dict
+
+
```
vicford, line 811: In `warming_levels_visualize` spaces need to be added to the postage_stamps_MAIN static text at the beginning of each line excluding the first. This is why there's a few typos in here.

vicford (changes requested):

mdkoenig:
[Warming Levels Notebook Debugging and Cleaning up Code](https://trello.com/c/voZZERx2/494-warming-levels-notebook-debugging-and-cleaning-up-code)

climakitae/explore/warming.py
```diff
@@ -151,23 +206,45 @@ def get_sliced_data(y, years, window=15, anom="Yes"):
...
+        # Find the centered year
+        centered_year = pd.to_datetime(center_time).year
+        start_year = centered_year - window
+        end_year = centered_year + (window - 1)
+
+        # TODO: This method will fail for daily data, as leap days create different time indices to be merged on. This will create weird time indices that can be visualized with the `out.plot.line` method in `warming_levels.ipynb`.
+
+        if anom == "Yes":
+            # Find the anomaly
+            anomaly = y.sel(time=slice("1981", "2010")).mean("time")
```
elehmer, line 223: variable naming is a bit confusing here but I get what you are doing.

climakitae/explore/warming.py
```diff
@@ -129,13 +161,36 @@ def clean_list(data, gwl_times):
...
+        time=slice(0, len(warm_data.time) - 1)
+    )  # -1 is just a placeholder for 30 year window, this could be more specific.
+
+    # Cleaning #3
+    # warm_data = warm_data.dropna(dim="all_sims")
+
+    return warm_data
+
+
+def get_sliced_data(y, level, years, window=15, anom="Yes"):
```
elehmer, line 185: The window is really `30` years in the case of `15` so a little confusing naming. Could take the window and then halve it and +/- it to get the time range.

elehmer: I see that you are trying to avoid the fractional year window issue...

elehmer (approved):
Looks good and runs on the HUB. Maybe just needs some polishing up at a later date like you said.

climakitae/explore/warming.py
claalmve: This is a great point. I'm not too sure how to change this for now (what is half a window called?). I'm going to leave it as is because the documentation already listed provides enough detail describing what `window` is, and this method is only using in our back-end, so at least it shouldn't be confusing for users. However, I'm going to include this as another bullet in the `GWL Future Improvements` doc. Thanks for catching this.

PR #289, opened 2023-10-20
Adding timezone conversion function in utils

claalmve:
This PR is focused on localizing data to a specific area's timezone, rather than UTC time, for applications that require hourly granularity and specific local time ranges (i.e. heating degree days of a specific area).

**To test:** Message me for a copy of a Jupyter Notebook that runs creates sampled data and runs the utils function on the given dataset to convert it to the local timezone.


claalmve:
Will need to include `timezonefinder` in Docker image for future usage.

vicford (changes requested):
Tested in the tester notebook provided, and on the hub. A few things:

Tester notebook: 
It appears that there still might be some 'too early' slicing of data with the timezone conversion, in that the last few hours in the converted data are not present (see screenshot)
Screenshot 2023-10-20 at 11 11 19 AM

Hub with WRF area-averaged gridded data, not station data:
This at present doesn't work because the CFTimeIndex requires cftime.datetime objects, and the WRF gridded date has numpy.datetime64. I think there should be some sort of if/else statement in the function that identifies based on the input selections whether it's station data --> use CFtime.datetime index; or gridded data --> process with numpy.datetime64 index
- 2 of the timezone sensitive functions (+ renewables once it's online) will use the gridded WRF data (TMY, heat index), and 1 uses station (degree days), so definitely needs to be flexible for both options!

elehmer:
> Will need to include `timezonefinder` in Docker image for future usage.

timezonefinder            6.0.2              pyhd8ed1ab_0    conda-forge

Looks like it is already on board the HUB image...

climakitae/util/utils.py
```diff
@@ -448,3 +451,68 @@ def hdh_cdh_lineplot(data):
     return data.hvplot.line(
         x="time", by="simulation", title=data.name, ylabel=data.name + " (degF)"
     )
+
+
+def convert_to_local_time(
+    data: xr.Dataset, selections: Select, lat: float, lon: float
```
elehmer, line 457: Can't we get the location information from the data itself? Either with `lat`, `lon` or the station name lookup? Are we trying to convert to the timezone of the data location or to a specific user defined timezone?

climakitae/util/utils.py
```diff
@@ -448,3 +451,68 @@ def hdh_cdh_lineplot(data):
...
+            year=pd_dt.year,
+            month=pd_dt.month,
+            day=pd_dt.day,
+            hour=pd_dt.hour,
+            minute=pd_dt.minute,
+            second=pd_dt.second,
+        )
+
+    # Apply function to Datetime objects and assign back to DataArray
+    total_data["time"] = np.array(localized.map(convert_to_cftime))
```
elehmer, line 509: Looking at this won't the `datetime64` data be converted to `cftime` for the gridded data? or does the `np.array` convert it back and then the station data will be not `cftime`?

elehmer (changes requested):

climakitae/util/utils.py
vicford: Intention is to convert from UTC to the local timezone of the selected location. Because parts of our domain are in mountain timezone (and central at 45km), depending on the location, would need to be able to identify the correct timezone

elehmer: Right, so how do we want to deal with data that spans 2 timezones? Just provide a single point like in the code now and use that or base it on cell/station location?

vicford: This is a good question. For single points/station locations, it's fine to use the timezone that it's in. Both the TMY and heat index applications are for point locations, so that's good. The degree days looks at both a cached area and a point location, and I know the renewables work will need gridded areas, so perhaps whatever the timezone is in the closest-to-central cell is for gridded areas?

elehmer:
Could we just make it simple on us and use a UTC offset parameter instead of the location info? Seems like the end user would know which timezone they would be working in?

climakitae/util/utils.py
```diff
@@ -9,7 +9,10 @@
 import pandas as pd
 import matplotlib.colors as mcolors
 import matplotlib
-
+import copy
+import cftime
+from timezonefinder import TimezoneFinder
+from climakitae.core.data_interface import Select
```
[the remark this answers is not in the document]
elehmer: I am confused why we need an extra year's data added onto the end?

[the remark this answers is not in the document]
elehmer: It also bombs if you try Historical Station data and have the time slice set to the end of the data (2015) as there is no data after that year...

claalmve:
> Tested in the tester notebook provided, and on the hub. A few things:
> 
> Tester notebook: It appears that there still might be some 'too early' slicing of data with the timezone conversion, in that the last few hours in the converted data are not present (see screenshot) Screenshot 2023-10-20 at 11 11 19 AM
> 
> Hub with WRF area-averaged gridded data, not station data: This at present doesn't work because the CFTimeIndex requires cftime.datetime objects, and the WRF gridded date has numpy.datetime64. I think there should be some sort of if/else statement in the function that identifies based on the input selections whether it's station data --> use CFtime.datetime index; or gridded data --> process with numpy.datetime64 index
> 
> * 2 of the timezone sensitive functions (+ renewables once it's online) will use the gridded WRF data (TMY, heat index), and 1 uses station (degree days), so definitely needs to be flexible for both options!

Did you use any specific selections to get the `adj_data`? Or just the ones used in the notebook?

Thanks for testing the WRF data and the station data, I can definitely add this conditional in the function.

climakitae/util/utils.py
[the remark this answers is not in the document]
claalmve: I can remove that typecast for now. That is also a great check Eric, I will definitely debug historical station data too.

climakitae/util/utils.py
```diff
@@ -448,3 +451,68 @@ def hdh_cdh_lineplot(data):
...
+
+    # Apply function to Datetime objects and assign back to DataArray
+    total_data["time"] = np.array(localized.map(convert_to_cftime))
+
+    # 4. Subset the data by the initial time
+    start = data.time[0].item()
+    end = data.time[-1].item()
+    sliced_data = total_data.sel(time=slice(start, end))
+
+    print("Finished!")
```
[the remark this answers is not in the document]
claalmve: Can do!

climakitae/util/utils.py
claalmve: Yes this is true. I will need to debug what happens if you select gridded data.

vicford:
> Could we just make it simple on us and use a UTC offset parameter instead of the location info? Seems like the end user would know which timezone they would be working in?

This would be an easy solution, but needs to be aware of daylight savings, in terms of whether it's UTC-8 or UTC-7 (for CA). Especially since some states in our domain don't abide by DS.

elehmer:
I did some fiddling in a separate branch trying to deal with `datetime64` from gridded data but ran into issues at the end of the code.

I am working on integrating `timezonefinder` into the testing conda environment so the tests won't fail. I'm starting from scratch though as I lost the file I used to generate the lock file when a hard drive failed...

climakitae/util/utils.py
```diff
@@ -448,3 +451,68 @@ def hdh_cdh_lineplot(data):
...
+def convert_to_local_time(
+    data: xr.Dataset, selections: Select, lat: float, lon: float
+) -> xr.Dataset:
+    """
+    Converts the inputted data to the local time of the selection.
+    """
+    # 1. Find the other data
+    start, end = selections.time_slice
+    tz_selections = copy.copy(selections)
+    tz_selections.time_slice = (
```
elehmer, line 465: Just noticed this affects the Select Panel GUI so that it displays this time_slice...

climakitae/util/utils.py
claalmve: I believe this is resolved with the consolidation of only using `datetime64[ns]` time now.

climakitae/util/utils.py
```diff
@@ -520,3 +520,40 @@ def summary_table(data):
...
+    tz_data = tz_selections.retrieve()
+
+    # 2. Combine the data
+    total_data = xr.concat([data, tz_data], dim="time")
+
+    # 3. Change datetime objects to local time
+    new_time = (
+        pd.DatetimeIndex(total_data.time)
+        .tz_localize("UTC")
+        .tz_convert("America/Los_Angeles")
```
claalmve, line 546: Change this to local time zone, not just hard-coded to Los Angeles.

climakitae/util/utils.py
```diff
@@ -520,3 +520,40 @@ def summary_table(data):
...
     return df
+
+
+def convert_to_local_time(data, selections, lat, lon) -> xr.Dataset:
+    """
+    Converts the inputted data to the local time of the selection.
+    """
+    # 1. Find the other data
+    start, end = selections.time_slice
+    tz_selections = copy.copy(selections)
```
claalmve, line 531: Replace with` copy.deepcopy`

climakitae/util/utils.py
```diff
@@ -520,3 +520,40 @@ def summary_table(data):
         df = df.sort_values(by=["year"])
 
     return df
+
+
+def convert_to_local_time(data, selections, lat, lon) -> xr.Dataset:
```
claalmve, line 525: Future iteration: pass in DataArray and location, function should figure it out from there.

climakitae/util/utils.py
claalmve: Potential V2: sense the selections from the data, no need to pass in selections with it

climakitae/util/utils.py
```diff
@@ -520,3 +520,40 @@ def summary_table(data):
...
+    """
+    Converts the inputted data to the local time of the selection.
+    """
+    # 1. Find the other data
+    start, end = selections.time_slice
+    tz_selections = copy.copy(selections)
+    tz_selections.time_slice = (
+        end + 1,
+        end + 1,
```
claalmve, line 534: Test the case where you are extracting data at the end of the time slice (2100).

claalmve: 2015 is end of historical

claalmve: > Temporal range of each dataset:
> historical_climate_range_wrf = (1980, 2015)
> historical_climate_range_loca = (1950, 2015)
> historical_climate_range_wrf_and_loca = (1981, 2015)
> historical_reconstruction_range = (1950, 2022)
> ssp_range = (2015, 2100)

claalmve: At the end of the time slice, don't bother getting the additional data. Just adjust the timezones.

claalmve: Test with daily/monthly data too, instead of just hourly.

claalmve: Try to test slicing on non-year slice. Instead of (2051, 2051), do a timestamp.

climakitae/util/utils.py
vicford, line 525: I'm not seeing where lat and lon are actually used in this function. Presumably this information is needed to determine the local timezone, but at present not being used because the timezone is hardcoded to PST

vicford:
Noticing this now on both the station data load and the convert_to_local_time load (likely because it's grabbing additional data): 

```
srv/conda/envs/notebook/lib/python3.9/site-packages/xarray/coding/times.py:360: FutureWarning: Index.ravel returning ndarray is deprecated; in a future version this will return a view on self.
  sample = dates.ravel()[0]
/home/jovyan/src/climakitae/climakitae/core/data_load.py:957: RuntimeWarning: Converting a CFTimeIndex with dates from a non-standard calendar, 'noleap', to a pandas.DatetimeIndex, which uses dates from the standard calendar.  This may lead to subtle errors in operations that depend on the length of time between dates.
  da_adj["time"] = da_adj.indexes["time"].to_datetimeindex()
```

climakitae/util/utils.py
vicford: > Try to test slicing on non-year slice. Instead of (2051, 2051), do a timestamp.

I would deprioritize this as the last item on the list of getting the timezone conversion functionality working, and would say it's not necessary at the present time. None of our notebooks at present have a finer non-year slice for data pulling.

climakitae/util/utils.py
claalmve: Yep! Fixing this later today.

climakitae/util/utils.py
```diff
@@ -520,3 +521,42 @@ def summary_table(data):
...
+    )  # This is assuming selections passed with be negative UTC time. Also to get the next year of data.
+
+    tz_data = tz_selections.retrieve()
+
+    # 2. Combine the data
+    total_data = xr.concat([data, tz_data], dim="time")
+
+    # 3. Change datetime objects to local time
+    tf = TimezoneFinder()
+    local_tz = tf.timezone_at(lng=lng, lat=lat)
```
vicford, line 545:
```suggestion
def convert_to_local_time(data, selections, lat, lon) -> xr.Dataset:
    """
    Converts the inputted data to the local time of the selection.
    """
    # 1. Find the other data
    start, end = selections.time_slice
    tz_selections = copy.copy(selections)
    tz_selections.time_slice = (
        end + 1,
        end + 1,
    )  # This is assuming selections passed with be negative UTC time. Also to get the next year of data.

    tz_data = tz_selections.retrieve()

    # 2. Combine the data
    total_data = xr.concat([data, tz_data], dim="time")

    # 3. Change datetime objects to local time
    tf = TimezoneFinder()
    local_tz = tf.timezone_at(lon=lon, lat=lat)
```

This is nitpicky, but we use "lon" for longitude exclusively throughout our codebase for consistency and we don't want to confuse our users.

climakitae/util/utils.py
```diff
@@ -520,3 +521,42 @@ def summary_table(data):
...
+        .astype("datetime64[ns]")
+    )
+    total_data["time"] = new_time
+
+    # 4. Subset the data by the initial time
+    start = data.time[0]
+    end = data.time[-1]
+    sliced_data = total_data.sel(time=slice(start, end))
+
+    print("Data converted to local timezone!")
```
vicford, line 560:
```suggestion
    print("Data converted to {}".format(local_tz)
```

Could be nice to confirm which timezone was converted to!

mdkoenig:
[timezone conversion functionality for calcs that must be timezone aware in climakitae and notebooks](https://trello.com/c/Ue2PUi8x/484-timezone-conversion-functionality-for-calcs-that-must-be-timezone-aware-in-climakitae-and-notebooks)

vicford (approved):
LGTM!

PR #291, opened 2023-11-01
Heat Index NB: helper functions

vicford:
1. Modifies the existing `compute_multimodel_stats` to include the multi-model median.
2. Modifies the existing `trendline` to provide an option to calculate the trendline either via the mean or median
3. Includes a new summary table function

### DIFF
--- a/climakitae/util/utils.py
+++ b/climakitae/util/utils.py
@@ -499,7 +499,7 @@ def summary_table(data):
     Returns
     -------
     df: pd.DataFrame
-        df is organized so that the simulations are stacked in individual columns by year
+        df is organized so that the simulations are stacked in individual columns by year/time
     """
 
     # Identify whether the temporal dimension is "time" or "year"
--- a/climakitae/util/utils.py
+++ b/climakitae/util/utils.py
@@ -406,11 +406,27 @@ def compute_multimodel_stats(data):
 
 
 def trendline(data, kind="mean"):
-    """Calculates treadline of the multi-model mean or median"""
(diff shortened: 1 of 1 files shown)


climakitae/util/utils.py
```diff
@@ -448,3 +470,17 @@ def hdh_cdh_lineplot(data):
     return data.hvplot.line(
         x="time", by="simulation", title=data.name, ylabel=data.name + " (degF)"
     )
+
+
+## Heat Index summary table helper
+def summary_table(data, add_stats=True):
```
claalmve, line 476: Is `add_stats=True` still needed?

claalmve:
These utils look good to me! I would suggest adding more information in the docstrings for future readability, to know what are the expected input and output objects for these util functions, but I can also understand if it's difficult to describe the intermediate results between function calls.

One thing that could be useful is a tip on when/where functions are used. Since `trendline` is currently only called after `compute_multimodel_stats` in the codebase, you could include a sentence describing that `trendline` is expecting a dataset shape similar to the output of `compute_multimodel_stats`, and that anyone interested in modifying the `trendline` input in the future should view `compute_multimodel_stats`.

Otherwise, I like the brevity of the methods, they make a lot of sense.

claalmve (approved):

climakitae/util/utils.py
vicford: Ah, good catch, no! Shall remove

Tianchi-Liu (approved):

PR #300, opened 2023-12-11
Update variable_descriptions.csv

claalmve:
Updating `variable_descriptions.csv` with new TMY variables.


vicford (approved):
Thanks @claalmve - DNI now is retrievable! My only request is that the 2 new added vars be added above swddif (row 18) in the csv file so that they are grouped with their complementary variables and for consistency in how we organized the descriptions file, rather than at the end

claalmve:
> Thanks @claalmve - DNI now is retrievable! My only request is that the 2 new added vars be added above swddif (row 18) in the csv file so that they are grouped with their complementary variables and for consistency in how we organized the descriptions file, rather than at the end

Fixed!

PR #302, opened 2023-12-23
WL Generation Debug

claalmve:
Debugging WL generation to allow for warming levels to be generated at a more granular level (0.5 vs 0.01). Fixes the bug of timestamps jumping to abnormal warming levels.

### DIFF
--- a/climakitae/util/generate_gwl_tables.py
+++ b/climakitae/util/generate_gwl_tables.py
@@ -123,18 +123,23 @@ def main():
                         )  # .to_pandas())
         return data_one_model
 
-    def get_gwl(smoothed, degrees):
+    def get_gwl(smoothed, degree):
         """Takes a smoothed timeseries of global mean temperature for multiple
         scenarios, and returns a small table of the timestamp that a given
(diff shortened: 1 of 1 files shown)


climakitae/util/generate_gwl_tables.py
```diff
@@ -151,22 +156,35 @@ def get_table_one_cesm2(
...
-        return gwlevels
+        for level in list(np.linspace(1.5, 4.00, 251)):
+            gwlevels[round(float(level), 3)] = get_gwl(oneModel.T, level)
+
+        # Modifying and returning oneModel to be seen as a WL lookup table with timestamp as index, to get the average WL across all simulations.
+        final_model = oneModel.T
+        final_model.columns = ens_mem + "_" + final_model.columns
+        return gwlevels, final_model
 
     def get_table_cesm2(
```
elehmer, line 167: I haven't been following the development of this as much as I should, but I am wondering why we are putting specific model code in here? Utils are usually generalized not specific to one data set.

PR #305, opened 2024-01-19
Fix for Long Error Message when Downscaling Method checkboxes are all unselected

elehmer:
This PR is an attempt to improve the behavior of when the Downscaling Method checkboxes are all unselected in `selections.show()`

Although not a true fix, this PR at the moment does away with the long error message and the selections behave as if the Dynamical checkbox is checked even though no checkboxes are checked. If someone has an idea how to recheck the checkbox that would be great!

Let me know if this is an improvement...


vicford:
@elehmer I like that the error isn't produced when neither are selected and that it defaults to the `dynamical` settings. For error tracking, is there anyway we can ensure that if a user goes to retrieve data but "forgets" to have a selection for dynamical or statistical selected it will not retrieve anything and will return some sort of statement of "Please select a downscaling method"? As an alternative, I suppose we could add a "Dynamical and Statistical" option for downscaling method and change the check boxes into radio buttons where something must be selected

elehmer:
> @elehmer I like that the error isn't produced when neither are selected and that it defaults to the `dynamical` settings. For error tracking, is there anyway we can ensure that if a user goes to retrieve data but "forgets" to have a selection for dynamical or statistical selected it will not retrieve anything and will return some sort of statement of "Please select a downscaling method"? As an alternative, I suppose we could add a "Dynamical and Statistical" option for downscaling method and change the check boxes into radio buttons where something must be selected

Ooo that's a good idea. Let me look into implementing radio buttons that way...

vicford:
Now getting this error @elehmer on using the `statistical+dynamical` option : 

`ValueError: COULD NOT RETRIEVE DATA: For the provided data selections, there is not sufficient data to retrieve. Try selecting a larger spatial area, or a higher resolution. Returning None.`

elehmer:
> Now getting this error @elehmer on using the `statistical+dynamical` option :
> 
> `ValueError: COULD NOT RETRIEVE DATA: For the provided data selections, there is not sufficient data to retrieve. Try selecting a larger spatial area, or a higher resolution. Returning None.`

Yes, I am deep in the ramifications of changing how `downscaling_method` works...

elehmer:
I think the change to Radio buttons is successful now! The GUI appears to work just like it did before (but of course without the issues where no checkboxes were checked). @vicford Can you give it a test run again?

vicford (approved):
Looks great, thanks @elehmer!

PR #306, opened 2024-01-22
Remove sopt from hvplot.image call

elehmer:
This PR removes the `sopt` parameter that is passed to `hvplot.image` and not recognized. The `sopt` parameter seems to have been used when data is divergent but I am not sure what it's intended purpose was?

### DIFF
--- a/climakitae/core/data_view.py
+++ b/climakitae/core/data_view.py
@@ -13,7 +13,13 @@ def compute_vmin_vmax(da_min, da_max):
     """Compute min, max, and center for plotting"""
     vmin = np.nanpercentile(da_min, 1)
     vmax = np.nanpercentile(da_max, 99)
-    return vmin, vmax
+    # define center for diverging symmetric data
+    if (vmin < 0) and (vmax > 0):
+        # dabs = abs(vmax) - abs(vmin)
+        sopt = True
+    else:
+        sopt = None
+    return vmin, vmax, sopt
 
 
 def view(data, lat_lon=True, width=None, height=None, cmap=None):
@@ -74,13 +80,14 @@ def view(data, lat_lon=True, width=None, height=None, cmap=None):
         # Such that the colorbar is standardized
         vmin = None
         vmax = None
+        sopt = None
         if "simulation" in data.dims:
             # But, only do this if the data is already read into memory
             # Or else the computation of min and max will take forever
             if data.chunks is None or str(data.chunks) == "Frozen({})":
                 min_data = data.min(dim="simulation")
                 max_data = data.max(dim="simulation")
-                vmin, vmax = compute_vmin_vmax(min_data, max_data)
+                vmin, vmax, sopt = compute_vmin_vmax(min_data, max_data)
 
         # Set default cmap if no user input
         if cmap is None:
@@ -147,6 +154,7 @@ def view(data, lat_lon=True, width=None, height=None, cmap=None):
                     width=width,
                     height=height,
                     clim=(vmin, vmax),
+                    symmetric=sopt,
                 )
             else:
                 # Make a scatter plot if it's just one grid cell
@@ -164,6 +172,7 @@ def view(data, lat_lon=True, width=None, height=None, cmap=None):
                     height=height,
                     s=150,  # Size of marker
                     clim=(vmin, vmax),
+                    symmetric=sopt,
                 )
         except:
             # Print message instead of raising error
--- a/climakitae/core/data_view.py
+++ b/climakitae/core/data_view.py
@@ -13,13 +13,7 @@ def compute_vmin_vmax(da_min, da_max):
     """Compute min, max, and center for plotting"""
     vmin = np.nanpercentile(da_min, 1)
     vmax = np.nanpercentile(da_max, 99)
-    # define center for diverging symmetric data
-    if (vmin < 0) and (vmax > 0):
-        # dabs = abs(vmax) - abs(vmin)
-        sopt = True
-    else:
-        sopt = None
-    return vmin, vmax, sopt
+    return vmin, vmax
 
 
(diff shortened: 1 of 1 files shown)


vicford:
> The `sopt` parameter seems to have been used when data is divergent but I am not sure what it's intended purpose was?

In warming levels/AMY some variables have both positive and negative values, but the colormapping would be heavily skewed to where we need the 0 point to be the inflection point in the colormap

elehmer:
> > The `sopt` parameter seems to have been used when data is divergent but I am not sure what it's intended purpose was?
> 
> In warming levels/AMY some variables have both positive and negative values, but the colormapping would be heavily skewed to where we need the 0 point to be the inflection point in the colormap

Having the divergent colormap centered on `0` would be difficult as the colorramp would be stretched on one side right? I'm not sure there are parameters to do that in hvplot?

vicford:
> Having the divergent colormap centered on `0` would be difficult as the colorramp would be stretched on one side right? I'm not sure there are parameters to do that in hvplot?

Our approach to fix that "stretch" was to use a diverging colormap, and to ensure that whatever the absolute max value was across the data was mirrored on the opposite scale. 
e.g., if the data ranges from -1 to 6, we'd want the scale to be -6 to 6 so that "0" is in the middle of the colormap. Without the symmetry piece was that it would put the inflection point in the colormap at like 3 which is unintuitive for visual interpretation. 
Ultimately we'd need to force the min/max range to ensure that colormapping sets that inflection point accurately so all negative changes are e.g. blue; and all positive changes are e.g. orange *

*noting that depending on the variable the "color" of positive switches --> 
temperature has orange/red as positive and blue as negative
precipitation has blue as positive (more moisture) and orange/reds as negative (less moisture) by standard convention

elehmer:
> > Having the divergent colormap centered on `0` would be difficult as the colorramp would be stretched on one side right? I'm not sure there are parameters to do that in hvplot?
> 
> Our approach to fix that "stretch" was to use a diverging colormap, and to ensure that whatever the absolute max value was across the data was mirrored on the opposite scale. e.g., if the data ranges from -1 to 6, we'd want the scale to be -6 to 6 so that "0" is in the middle of the colormap. Without the symmetry piece was that it would put the inflection point in the colormap at like 3 which is unintuitive for visual interpretation. Ultimately we'd need to force the min/max range to ensure that colormapping sets that inflection point accurately so all negative changes are e.g. blue; and all positive changes are e.g. orange *
> 
> *noting that depending on the variable the "color" of positive switches --> temperature has orange/red as positive and blue as negative precipitation has blue as positive (more moisture) and orange/reds as negative (less moisture) by standard convention

I don't think this is currently working as I see non-symmetrical plots in `ck.view` for Air Temp in C. Is that what you are seeing?

I think maybe we can use the `hvplot.redim` to reset the range for plotting so that `0` is the middle. I'm not sure we have any of the logic installed either to know when to apply this shift in the range?

elehmer:
> > > Having the divergent colormap centered on `0` would be difficult as the colorramp would be stretched on one side right? I'm not sure there are parameters to do that in hvplot?
> > 
> > 
> > Our approach to fix that "stretch" was to use a diverging colormap, and to ensure that whatever the absolute max value was across the data was mirrored on the opposite scale. e.g., if the data ranges from -1 to 6, we'd want the scale to be -6 to 6 so that "0" is in the middle of the colormap. Without the symmetry piece was that it would put the inflection point in the colormap at like 3 which is unintuitive for visual interpretation. Ultimately we'd need to force the min/max range to ensure that colormapping sets that inflection point accurately so all negative changes are e.g. blue; and all positive changes are e.g. orange *
> > *noting that depending on the variable the "color" of positive switches --> temperature has orange/red as positive and blue as negative precipitation has blue as positive (more moisture) and orange/reds as negative (less moisture) by standard convention
> 
> I don't think this is currently working as I see non-symmetrical plots in `ck.view` for Air Temp in C. Is that what you are seeing?
> 
> I think maybe we can use the `hvplot.redim` to reset the range for plotting so that `0` is the middle. I'm not sure we have any of the logic installed either to know when to apply this shift in the range?

Here is an example of what I'm seeing:

image

vicford:
> > I don't think this is currently working as I see non-symmetrical plots in `ck.view` for Air Temp in C. Is that what you are seeing?
> > I think maybe we can use the `hvplot.redim` to reset the range for plotting so that `0` is the middle. I'm not sure we have any of the logic installed either to know when to apply this shift in the range?
> 
> Here is an example of what I'm seeing:

This is more of an issue for the warming levels postage stamps and in AMY, rather than ck.view but I think the plotting functionality is still based on the same code? However, in that example figure, since there are negative values it should switch to ae_diverging colormap rather than ae_orange (default for air temperature) **and** have the colormap centered at zero so there would have to be some assessment of the min/max range and set the plotting range to reflect that.

elehmer:
> > > I don't think this is currently working as I see non-symmetrical plots in `ck.view` for Air Temp in C. Is that what you are seeing?
> > > I think maybe we can use the `hvplot.redim` to reset the range for plotting so that `0` is the middle. I'm not sure we have any of the logic installed either to know when to apply this shift in the range?
> > 
> > 
> > Here is an example of what I'm seeing:
> 
> This is more of an issue for the warming levels postage stamps and in AMY, rather than ck.view but I think the plotting functionality is still based on the same code? However, in that example figure, since there are negative values it should switch to ae_diverging colormap rather than ae_orange (default for air temperature) **and** have the colormap centered at zero so there would have to be some assessment of the min/max range and set the plotting range to reflect that.

I fixed the symmetric axis issue. As far as using the right colormap the air temp dataset uses the orange colormap by default which works well for F units. If you switch to C you can specify the diverging colormap to `ck.view` for a better rendering.

PR #308, opened 2024-01-23
Fosberg Derived Index fix where statements reclassing data incorrectly

elehmer:
This PR changes the way the Fosberg Fire Index is generated. Removes 2 where statements that tried to reclass values < 0 and > 100 to 0 and 100. Unfortunately, the `xr.where` statements reclassified data outside of the spatially clipped region. PR uses `np.clip` to clip values outside of [0-100] to the boundary values.

### DIFF
--- a/climakitae/tools/indices.py
+++ b/climakitae/tools/indices.py
@@ -2,6 +2,7 @@
 
 import xarray as xr
 import numpy as np
+import rioxarray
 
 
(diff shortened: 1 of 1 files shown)


vicford (approved):
Thanks @elehmer

elehmer:
Everyone happy with the functioning of this now?

PR #310, opened 2024-01-25
Making changes to GWL Generating Tables file to accomodate for agnostic tools

claalmve:
These are changing in addition to the `agnostic` tools PR, to allow for the generation of GWL tables needed for `agnostic_tools` notebook in `cae-notebooks`.

### DIFF
--- a/climakitae/util/generate_gwl_tables.py
+++ b/climakitae/util/generate_gwl_tables.py
@@ -123,18 +123,23 @@ def main():
                         )  # .to_pandas())
         return data_one_model
 
-    def get_gwl(smoothed, degrees):
+    def get_gwl(smoothed, degree):
         """Takes a smoothed timeseries of global mean temperature for multiple
         scenarios, and returns a small table of the timestamp that a given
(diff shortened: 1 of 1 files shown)


climakitae/util/generate_gwl_tables.py
```diff
@@ -123,18 +123,23 @@ def build_timeseries(variable, model, ens_mem, scenarios):
...
-        for scenario in smoothed:
-            if smoothed[scenario].sub(degrees).abs().min() > 0.01:
-                gwl[scenario] = np.NaN
+
+        def get_wl_timestamp(scenario, degree):
+            """
+            Given a scenario of wl's and timestamps, find the timestamp that first crosses the degree passed in.
+            Return np.NaN if none of the timestamps pass this degree level.
+            """
+            if any(scenario >= degree):
```
claalmve, line 136: Note to self: Doesn't this need to account for Naomi's previous comment? "Make sure it's not just choosing one of the final timestamps because it's the highest warming despite being no where close to target...?"

PR #322, opened 2024-02-21
Making initial structure and functionality for batch mode

claalmve:
Making the initial structure and functionality for batch mode.


climakitae/util/utils.py
```diff
@@ -685,3 +686,52 @@ def convert_to_local_time(data, selections):  # , lat, lon) -> xr.Dataset:
...
+
+def batch_select(points):
+    """
+    Conducts batch mode analysis on a series of points for a given metric.
+    """
+    # Retrieve data for the entire service area
+    selections = Select()
+    selections.data_type = "Gridded"
+    selections.area_subset = "none"
+    selections.cached_area = ["entire domain"]
```
claalmve, line 699: This is the maximum spanning subset that a user can select from. Now, this is not ideal from a time/space complexity standpoint, but it will work.

32 stations = ~44 min. when calculating heat index (which is typically 8 min. for 1 station).

Future iterations of batch mode can address this with different architecture.

elehmer:
This could be moved into a file of its own like `tools/batch.py`

climakitae/util/utils.py
claalmve: **Note:** Reformatting code to create 32 selection objects, retrieving data, and combining the arrays together to load causes kernel to crash. This current PR may be the best approach so far, even if its runtime can be excessive.

climakitae/tools/batch.py
Tianchi-Liu, line 1: Having the lat lon in the simulation dimension could be potentially confusing. Maybe consider storing the coordinate information in its own dimension.

climakitae/tools/batch.py
```diff
@@ -0,0 +1,59 @@
...
+        data_pts.append(closest_cell)
+
+    # Combine data points into a single xr.Dataset
+    cells_of_interest = xr.concat(data_pts, dim="simulation").chunk(
+        (1, len(closest_cell.time))
+    )
+
+    # Load in the cells of interest into memory, if desired.
+    if load_data:
+        cells_of_interest = cells_of_interest.compute()
```
Tianchi-Liu, line 50: ck.load is a user-friendly wrapper around compute

`data = ck.load(data)` loads data into memory

climakitae/util/utils.py
Tianchi-Liu, line 1: Nit: These seem to belong to a time zone PR.

Tianchi-Liu (approved):
This is working great for heat index data. I left some comments minor compared to runtime. Those I think are nice to have in future versions.

climakitae/tools/batch.py
elehmer: > ck.load is a user-friendly wrapper around compute
> 
> `data = ck.load(data)` loads data into memory

you'd have to load `load` from `data_load` to access that func. and it might cause import issues but worth a shot

climakitae/tools/batch.py
claalmve: Yes, this is true. I believe I tried storing the lat/lon information in coordinate form and then concatenated the simulations together, but I think xarray didn't recognize these as separate, distinct DataArrays, so it would just merge them together instead of keeping them as separate simulations.

climakitae/util/utils.py
claalmve: Good catch, accidentally had them added here.

climakitae/tools/batch.py
claalmve: I will try to fix this on the next version of batch mode. For now, just merging this version in so we make progress towards the next version.

PR #326, opened 2024-02-29
Bug fix on TMY: elevation in export

vicford:
# Description of PR

Previously, TMY file export was breaking on using an API requests call to the EPQS DEM to obtain elevation. As far as I can tell, this is on the EPQS side of things and not on our code base. To address this, I added the corresponding elevation values from `wecc-station` csv to the `hadisd_stations` csv file that is used in the climakitae backend, and modified the TMY code to obtain the corresponding elevation value when requested. At present, when a custom location is passed, elevation is set to 0. This will be a future fix. 

N/A

## Type of change

# How Has This Been Tested?

To test, please run the TMY notebook and ensure that both .tmy and .epw files are produced

# Checklist:

- [ ] I have added tests that prove my fix is effective or that my feature works -- N/A
- [ ] New and existing unit tests pass locally with my changes -- N/A

### DIFF
--- a/climakitae/core/data_export.py
+++ b/climakitae/core/data_export.py
@@ -966,11 +966,11 @@ def write_tmy_file(
             state = station_df.loc[station_df["station id"] == station_code][
                 "state"
             ].values[0]
-            station_code = str(station_code)[:6]
-            timezone = _utc_offset_timezone(lon=stn_lon, lat=stn_lat)
             elevation = station_df.loc[station_df["station id"] == station_code][
                 "elevation"
             ].values[0]
+            station_code = str(station_code)[:6]
+            timezone = _utc_offset_timezone(lon=stn_lon, lat=stn_lat)
 
     # typical meteorological year format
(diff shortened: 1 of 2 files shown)


vicford:
> I understand that fix is on the way, but I wonder if it makes sense for TMY to have -9999 elevations when custom locations are passed. Or another value that cannot be mistaken as actual elevation. Just a thought!

This is a great point @Tianchi-Liu - the official EPW documentation states the range of acceptable elevation is -1000 to 9999.9, with a default value of 0.0. They don't have a rule for missing value for elevation, and I suspect that if a value is missing in the header file, the analysis program fails to run. I think for the time being it is okay to use the default value as a stop gap, and perhaps incorporate an optional argument with markdown explaining that if a custom location is passed the user must provide the elevation value or it will default to 0

claalmve:
Ran TMY export and generation looks good to me with the change in deriving elevation. Nice job finding a fix so quickly!

PR #329, opened 2024-03-06
Make updates to data_export

elehmer:
# Description of PR

Making changes to data_export to remove redundant code, clean up, and give more buffer to user partition
 
- Added compression for local exports. This doubles as a safety buffer as we estimate on non-compressed size.
- Fixed hardcoded paths.
- Added option to explicitly specify which export method to use.

JupyterHub user partitions have filled up and frozen users out and the large files that did it were netCDF exports. This PR should resolve this in most cases.

## Type of change

- [ x ] New feature (non-breaking change which adds functionality)

# How Has This Been Tested?

- [ x ] Ran multiple saves on HUB.

### DIFF
--- a/climakitae/core/data_export.py
+++ b/climakitae/core/data_export.py
@@ -216,7 +216,7 @@ def _export_to_netcdf(data, save_name, mode):
         desired output file name, including the file extension
     mode : string
         localtion logic for storing export file.
-        
+
     Returns
     -------
     None
--- a/climakitae/core/data_export.py
+++ b/climakitae/core/data_export.py
@@ -161,7 +161,7 @@ def _fillvalue_compression_encoding(data):
     comp = dict(zlib=True, complevel=6)
     encoding = {var: comp for var in data.data_vars}
     print(encoding)
-    return encoding.update(filldict)
+    return filldict | encoding
 
 
 def _create_presigned_url(bucket_name, object_name, expiration=60 * 60 * 24 * 7):
--- a/climakitae/core/data_export.py
+++ b/climakitae/core/data_export.py
@@ -224,8 +224,14 @@ def _export_to_netcdf(data, save_name):
         _warn_large_export(est_file_size)
         _update_attributes(data)
         _update_encoding(data)
-        comp = dict(_FillValue=None)
-        encoding = {coord: comp for coord in data.coords}
+        fill = dict(_FillValue=None)
+        filldict = {coord: fill for coord in data.coords}
+        comp = dict(zlib=True, complevel=6)
+        if isinstance(data, xr.core.dataarray.Dataset):
+            encoding = {var: comp for var in data.data_vars}
+        else:
+            encoding = {data.name: comp}
+        encoding.update(filldict)
         data.to_netcdf(path, encoding=encoding)
         print(
             (
--- a/climakitae/core/data_export.py
+++ b/climakitae/core/data_export.py
@@ -631,7 +631,10 @@ def _export_to_csv(data, save_name):
 
 
 def export(data, filename="dataexport", format="NetCDF", mode="auto"):
-    """Save data as a file in the current working directory.
+    """Save xarray data as either a NetCDF or CSV in the current working directory,
+    or stream the export file to an AWS S3 scratch bucket and give download URL. Default
+    behavior is for the code to automatically determine the output destination based on whether
+    file is small enough to fit in HUB user partition, this can be overridden using the mode parameter.
 
     Parameters
     ----------
--- a/climakitae/core/data_export.py
+++ b/climakitae/core/data_export.py
@@ -532,7 +532,7 @@ def _dataset_to_dataframe(dataset):
     return df
 
 
-def _export_to_csv(data, output_path):
+def _export_to_csv(data, save_name):
     """
     Export user-selected data to CSV format.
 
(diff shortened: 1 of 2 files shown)


mdkoenig:
[Look into Export partition issues](https://trello.com/c/ccrrx4Q9/558-look-into-export-partition-issues)

elehmer:
> > > While I’m still doing tests, I have a question about the option to specify the save location. I am not sure if the users need this option, or how it helps with the user partitions issue. Could you clarify the motivation for adding it?
> > 
> > 
> > I added it because I was having an issue where I couldn't download large files from the HUB, but that has resolved itself. It is invisible to the notebooks (no changes required). I think it is a nice option to have if you know the file is too big before hand or it could be useful of us when doing scripting in the future.
> 
> Got it. I can definitely see it becoming useful for developers. I think it would also be nice if the users have the option to save to s3, in case they run into issues downloading from the hub. Users, however, could use more guidance. The case when the local option cannot be honored also needs to be handled. This might be a larger discussion. For this PR, I’d suggest keeping the option in a hidden function for developers until user-facing features are fully developed.

Ah! I see my error in logic! Thanks! There is a scenario where they could specify "local" and fill up the drive! Oops!

elehmer:
So, there seems to me to be a practical limit to what can be exported to CSV. At somewhere between 0.75-1.00GB `DataArray` size (at least for WRF) the conversion to pandas `DataFrame` exceeds the available memory on the HUB and kernel crashes. So I have added logic to handle that. Estimating CSV file size before generating it is very hard as its relative size compared to the memory size of the array varies. Very small memory arrays end up taking more space (even compressed) on export. Larger arrays end up being roughly 2x the size of the compressed CSV (although the uncompress CSV is 10x the array size!). So the 0.85GB limit I have imposed seems like a good idea. I would like some testing around this to optimize this threshold.

elehmer:
> > > > While I’m still doing tests, I have a question about the option to specify the save location. I am not sure if the users need this option, or how it helps with the user partitions issue. Could you clarify the motivation for adding it?
> > > 
> > > 
> > > I added it because I was having an issue where I couldn't download large files from the HUB, but that has resolved itself. It is invisible to the notebooks (no changes required). I think it is a nice option to have if you know the file is too big before hand or it could be useful of us when doing scripting in the future.
> > 
> > 
> > Got it. I can definitely see it becoming useful for developers. I think it would also be nice if the users have the option to save to s3, in case they run into issues downloading from the hub. Users, however, could use more guidance. The case when the local option cannot be honored also needs to be handled. This might be a larger discussion. For this PR, I’d suggest keeping the option in a hidden function for developers until user-facing features are fully developed.
> 
> Ah! I see my error in logic! Thanks! There is a scenario where they could specify "local" and fill up the drive! Oops!

5a408b1 Should fix this

elehmer:
[Ability to estimate file sizes (CSV)](https://trello.com/c/5J8YQ2HI/569-ability-to-estimate-file-sizes-csv)

climakitae/core/data_export.py
```diff
@@ -183,7 +213,7 @@ def _create_presigned_url(bucket_name, object_name, expiration=60 * 60 * 24 * 7)
     return url
 
 
-def _export_to_netcdf(data, save_name):
+def _export_to_netcdf(data, save_name, mode):
```
Tianchi-Liu, line 216: The docstring about save location logic is outdated.

climakitae/core/data_export.py
```diff
@@ -199,17 +229,50 @@ def _export_to_netcdf(data, save_name):
         data to export to NetCDF format
     save_name : string
         desired output file name, including the file extension
+    mode : string
+        localtion logic for storing export file.
```
Tianchi-Liu, line 233:
```suggestion
        location logic for storing export file.
```

climakitae/core/data_export.py
```diff
@@ -199,17 +229,50 @@ def _export_to_netcdf(data, save_name):
...
+
+    _warn_large_export(est_file_size)
+    _update_attributes(_data)
+    _update_encoding(_data)
+
+    file_location = "local"
+
+    if mode == "local":
+        if disk_space <= est_file_size:
+            raise Exception("Data too large to save locally. Use the mode=s3 option.")
```
Tianchi-Liu, line 261: “locally” might be confused with the “local machine” phrase printed after a save succeeds. “s3” can be confusing too, especially to non-technical users.

climakitae/core/data_export.py
```diff
@@ -539,6 +642,8 @@ def export(data, filename="dataexport", format="NetCDF"):
         of "my_filename.nc"). The default is "dataexport".
     format : str, optional
         File format ("NetCDF" or "CSV"). The default is "NetCDF".
+    mode : str, optional
+        Save location logic for NetCDF file ("auto", "local", "s3"). The default is "auto"
```
Tianchi-Liu, line 646: Again, “local” and “s3” can be confusing.

climakitae/core/data_export.py
```diff
@@ -525,9 +622,15 @@ def _export_to_csv(data, output_path):
...
+        (
+            "Saved! You can find your file(s) in the panel to the left"
+            " and download to your local machine from there."
+        )
+    )
 
 
-def export(data, filename="dataexport", format="NetCDF"):
+def export(data, filename="dataexport", format="NetCDF", mode="auto"):
     """Save data as a file in the current working directory.
```
Tianchi-Liu, line 634: or save the file to a S3 bucket and provide a
URL for download

Tianchi-Liu (commented):
#### Review related to NetCDF:

Exports tested all succeed! Below are a couple thoughts on messaging and documentation. 

The compression greatly reduces the size of large (GB-level) exports. (No action needed in this PR, but as a result, compressed local exports seem to be a promising solution to S3 exports blowing up the memory.)

Nit: DataArrays are converted to Datasets. This change is small, but will affect downstream operations. It’d be nice to have the DataArray data structure back if convenient, or give a heads-up about the change.

climakitae/core/data_export.py
elehmer: Very valid point. I will come up with alternative language. Thx

climakitae/core/data_export.py
elehmer: Added more documentation...

climakitae/core/data_export.py
```diff
@@ -199,17 +229,50 @@ def _export_to_netcdf(data, save_name):
...
     -------
     None
 
     """
-    est_file_size = _estimate_file_size(data, "NetCDF")
-    disk_space = shutil.disk_usage("./")[2] / bytes_per_gigabyte
+    print("Exporting specified data to NetCDF...")
+
+    # Convert xr.DataArray to xr.Dataset so that compression can be utilized
+    _data = data
```
elehmer, line 243: I just realized that this might create a whole other copy of the data and would cause memory to double?

elehmer: Hmm, I will look further into this in another PR

climakitae/core/data_export.py
elehmer: I've decided to keep the language the same and just explain it more

PR #330, opened 2024-03-08
Generalizing agnostic tools to include WRF

claalmve:
# Description of PR

Generalizing agnostic tools to handle WRF data as well.

We want to visualize WRF data, and have the option to select 1 or more specific simulations, so we need to generalize existing code (agnostic tools) in order to do so.

None

None

## Type of change

# How Has This Been Tested?

Running `agnostic_tools.ipynb` with code changes and confirming that WRF data is indeed extracted.

# Checklist:

### DIFF
--- a/climakitae/explore/agnostic.py
+++ b/climakitae/explore/agnostic.py
@@ -643,13 +643,17 @@ def plot_WRF(sim_vals, metric):
 def plot_double_WRF(var1, var2):
     """Plots aggregations of WRF models on a scatterplot of two quantitative variables. Labels points with specific WRF model names."""
     # Make sure that the two variables are the same length and have the same simulation names
-    if (len(var1) != len(var2)) & (set(var1.simulation.values) != set(var2.simulation.values)):
-        raise IndexError('The two variables must have the same length of simulations and have the same simulation names.')
-    
-    var1 = var1.sortby('simulation')
-    var2 = var2.sortby('simulation')
-    fig, ax = plt.subplots(figsize=(7,5))
-    
+    if (len(var1) != len(var2)) & (
+        set(var1.simulation.values) != set(var2.simulation.values)
+    ):
+        raise IndexError(
+            "The two variables must have the same length of simulations and have the same simulation names."
+        )
+
+    var1 = var1.sortby("simulation")
+    var2 = var2.sortby("simulation")
+    fig, ax = plt.subplots(figsize=(7, 5))
+
     # Get sim names
     sims = [name.split(",")[0] for name in list(var1.simulation.values)]
     sims = [name[4:] for name in sims]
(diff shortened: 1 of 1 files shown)


climakitae/explore/agnostic.py
```diff
@@ -466,14 +487,21 @@ def get_lat_lon_loca(lat, lon, metric, years, months=list(np.arange(1, 13))):
...
     # Create selections object
-    selections = _create_lat_lon_select(lat, lon, metric, years)
+    selections = _create_lat_lon_select(lat, lon, metric, downscaling_method, years)
     # Runs calculations and derives statistics from LOCA data pulled via selections object
     return _compute_selections_and_stats(selections, metric, years, months)
 
 
-def get_area_subset_loca(
-    area_subset, cached_area, metric, years, months=list(np.arange(1, 13))
+def agg_area_subset_sims(
```
Tianchi-Liu, line 498: The function name should be updated also in agnostic_tools.ipynb soon.

climakitae/explore/agnostic.py
```diff
@@ -437,7 +456,9 @@ def _compute_selections_and_stats(selections, metric, years, months):
     return single_stats, multiple_stats, results
 
 
-def get_lat_lon_loca(lat, lon, metric, years, months=list(np.arange(1, 13))):
+def agg_lat_lon_sims(
```
Tianchi-Liu, line 459: The function name should be updated also in agnostic_tools.ipynb soon.

climakitae/explore/agnostic.py
```diff
@@ -362,10 +383,8 @@ def _compute_results(selections, metric, years, months):
 
     # Retrieving closest grid-cell's data for lat/lon area subsetting
     if selections.area_subset == "lat/lon":
```
Tianchi-Liu, line 385: Users of `agg_area_subset_sims` could potentially subset area by “lat/lon”, in which case I think `agg_area_subset_sims `intends to do computation on all data?

Tianchi-Liu (commented):
Two major concerns reading through the changes:

climakitae/explore/agnostic.py
claalmve: Do you mean in the case where all the user wants to select on a bounding box of lat/lons?

Tianchi-Liu: Yes, for such users, if any, is the calculation intended to incorporate all grid cells in the bounding box, instead of the one grid cell in the middle?

claalmve: This is a good point, I think this may be a question for development standup. Currently, I believe the scope of the function is to still just handle individual gridpoints, but as users will want to include assets that span multiple gridcells, I can imagine that we would need to modify the function to handle these cases too.

PR #331, opened 2024-03-14
Re-add 2 sims that were removed before

elehmer:
# Description of PR

This PR readds 2 simulations that were being filtered out of selections.


Tianchi-Liu (approved):
All LOCA-dependent notebooks are running, except for agnostic_tools.ipynb which is under development in PR 332. I’ll test it there.

PR #344, opened 2024-05-01
Making slight modifications to `agnostic.py` to show all 8 WRF models

claalmve:
# Description of PR
Currently, `agnostic.py` pulls 3km monthly data to create aggregations for LOCA/WRF simulations (V1). However, this may be a bit misleading since we have 8 WRF simulations to pull from rather than 4 WRF simulations shown in later visualizations in `agnostic_tools.ipynb`. So, this PR fixes that to allow for more timescales of data to be used for WRF data (hourly, daily, monthly), but not for LOCA, since the LOCA simulations in total are more data intensive.

Additionally, this PR upgrades the climate response plots at the end of the notebook by converting them from matplotlib plots to hvplots. This allows the users to toggle GCMs/simulations as desired, and save the figure if they would like.

This PR also makes a slight change to the `plot_WRF` function, allowing 4 WRF simulations or 8 WRF simulations to be visualized properly without overlapping x-labels.

1. Showing all 8 WRF simulations depending on timescale.
2. Upgrading climate response plot from matplotlib to hvplot
3. Plotting 4 vs 8 WRF simulations correctly respectively.

We want to show all 8 WRF simulations, we want to toggle simulations/GCMs, we want more visually appealing plots.

## Type of change

# Checklist:


climakitae/explore/agnostic.py
```diff
@@ -277,20 +277,21 @@ def create_conversion_function(lookup_tables):
 ##### TASK 2 #####
 
 
-def _get_var_info(variable, downscaling_method):
+def _get_var_info(variable, downscaling_method, wrf_timescale="monthly"):
     """Gets the variable info for the specific variable name and downscaling method"""
     var_desc_df = read_csv_file(variable_descriptions_csv_path)
+    timescale = wrf_timescale if downscaling_method == "Dynamical" else "monthly"
```
vicford, line 283: I think this line is ultimately redundant, especially since the default setting is intentionally "monthly". We've included options on how to set this in the notebook

climakitae/explore/agnostic.py
```diff
@@ -277,20 +277,21 @@ def create_conversion_function(lookup_tables):
...
 
-def _get_var_info(variable, downscaling_method):
+def _get_var_info(variable, downscaling_method, wrf_timescale="monthly"):
     """Gets the variable info for the specific variable name and downscaling method"""
     var_desc_df = read_csv_file(variable_descriptions_csv_path)
+    timescale = wrf_timescale if downscaling_method == "Dynamical" else "monthly"
     return var_desc_df[
         (var_desc_df["display_name"] == variable)
-        & (var_desc_df["timescale"].str.contains("monthly"))
+        & (var_desc_df["timescale"].str.contains(timescale))
```
vicford, line 286: Might need to add some error tracing here in terms of "other" options passed... like is daily an option?

vicford (changes requested):
Okay, so this works for monthly 100% fine, but on testing for hourly setting in WRF it retrieves data fine but the last plotting cell fails because there are a different number of simulations retrieved. 
- Either we force only selection of monthly (eh)
- Add in "guardrails" and only return simulations that overlap in both variables in the backend
   - This is curious to me because there should be 8 simulations for hourly precipitation and there's not being returned here 
Screenshot 2024-05-01 at 5 25 02 PM

climakitae/explore/agnostic.py
claalmve: I think this line just prevents LOCA2 simulations from being run on more granular timescales (like daily or hourly), in case the user were to input daily/hourly for `wrf_timescale` but also use a `downscaling_method` of Statistical.

climakitae/explore/agnostic.py
claalmve: `daily` has been tested to work, and I will add a validation check to make sure no erroneous strings are being passed in.

claalmve:
> Okay, so this works for monthly 100% fine, but on testing for hourly setting in WRF it retrieves data fine but the last plotting cell fails because there are a different number of simulations retrieved.
> 
> * Either we force only selection of monthly (eh)
> * Add in "guardrails" and only return simulations that overlap in both variables in the backend
>   
>   * This is curious to me because there should be 8 simulations for hourly precipitation and there's not being returned here
>     Screenshot 2024-05-01 at 5 25 02 PM

This is a good point! I did some debugging and realize the `wrf_timescale` wasn't being passed into `results_gridcell2`, allowing the timescales to be different for each results aggregation. The timescale is now passed in so it should be consistent!

vicford (changes requested):
Getting this on the `hourly` selection for Dynamical data

`single_stats_area, multiple_stats_area, results_area = agg_area_subset_sims(area_subset, selected_area, downscaling_method, variable, agg_func, units, years, months, wrf_timescale)`

Screenshot 2024-05-14 at 11 07 39 AM

vicford (approved):
User error on comment above :) 
Changed some of the default language in the markdown itself

PR #387, opened 2024-06-25
Updating GWL Tables to include WL 2.5

claalmve:
# Description of PR

Adding support for WL 2.5 as needed per CPUC ruling.

## Type of change

# Checklist:


climakitae/util/logger.py
claalmve, line 1: This was renamed because upon running `python generate_gwl_tables.py`, it threw an error because the logging.py file also imports a library called `logging` (causing a circular import). So, our `logging.py` file wsa renamed to `logger.py` to accomodate for that.

climakitae/explore/vulnerability.py
```diff
@@ -244,10 +244,13 @@ def _check_params_get_name(
     if approach == "warming_level" and warming_level not in [
         "1.5",
         "2.0",
+        "2.5",
         "3.0",
         "4.0",
     ]:
-        raise ValueError("Passed in warming level must be '1.5', '2.0', '3.0', '4.0'")
+        raise ValueError(
+            "Passed in warming level must be '1.5', '2.0', '2.5', '3.0', '4.0'"
```
vicford, line 252: Noting that the filename exported for vulnerability data at 2.5 warming level is "... _2degreeWL_..." so we need to modify the filename to indicate the 2.5 piece. Wondering here now if we do "2pt5WL" for the 1.5 and 2.5. What do you think @claalmve ?

vicford (changes requested):
Small change on the vulnerability naming piece here @claalmve for export (could be a separate PR). 

This seems to run fine in both the vulnerability assessment notebook and warming levels (in that nothing breaks) but am noticing that 2.5°C is not an option in the WL visualize panel and 2.5°C is not one of the "main" levels that is calculated in the compute step
- This is pending feedback from Owen & Mark on desired behavior (and could come later in a separate PR)

climakitae/explore/vulnerability.py
claalmve: Makes sense to me!

vicford (approved):

PR #393, opened 2024-06-27
Update Scenario Param trigger fix

elehmer:
# Description of PR

This PR updates the `_update_scenario` param trigger to properly set the historical scenarios after changing the affected widgets.

previously we tested if one list was in another (currently hist. scen. options vs. possible new options) using `in`. This apparently does not work as it was returning false when it was true. I therefore wrote a function that converts the lists into numpy arrays and compares them.

selections are being changed after parameters are being set and bring back incorrect data. Most obviously a problem when not using the GUI as you can't see that historical selections have changed.

## Type of change

# How Has This Been Tested?

- [ x ] Tested by creating the code from scratch on HUB and by running branch on HUB and using `ck.Select`
- [ ] Test B

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings

### DIFF
--- a/climakitae/core/data_interface.py
+++ b/climakitae/core/data_interface.py
@@ -996,21 +996,19 @@ class DataParameters(param.Parameterized):
         ]
         self.param["scenario_historical"].objects = scenario_historical_options
 
-        # test to check if list "a" is contained within list "b" as python doesn't have that out of the box
-        def _check_list_contained(a, b):
-
-            # convert list a to numpy array
-            a_arr = np.array(a)
-            # convert list b to numpy array
-            b_arr = np.array(b)
-
-            for i in range(len(b_arr)):
-                if np.array_equal(a_arr, b_arr[i : i + len(a_arr)]):
-                    return True
-            return False
+        # check if input historical scenarios match new available scenarios
+        # if no reanalysis scenario then return False
+        def _check_inputs(a, b):
+            chk = False
+            if len(b) < 2:
+                return chk
+            for i in a:
+                if i in a:
+                    chk = True
+            return chk
 
         # check if new selection has the historical scenario options and if not select the first new option
-        if _check_list_contained(_scenario_historical, scenario_historical_options):
+        if _check_inputs(_scenario_historical, scenario_historical_options):
             self.scenario_historical = _scenario_historical
         else:
             self.scenario_historical = [scenario_historical_options[0]]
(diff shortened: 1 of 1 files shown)


nicolejkeeney (approved):
This looks good to me, can you add some code comments in though to explain why the change was made? Maybe just a line or two from your description in the PR? The method is somewhat unintuitive I think (but makes sense with your description).

elehmer:
> This looks good to me, can you add some code comments in though to explain why the change was made? Maybe just a line or two from your description in the PR? The method is somewhat unintuitive I think (but makes sense with your description).

Any better?

climakitae/core/data_interface.py
```diff
@@ -991,7 +995,24 @@ def _update_scenarios(self):
             if scen in historical_scenarios
         ]
         self.param["scenario_historical"].objects = scenario_historical_options
-        if self.scenario_historical not in scenario_historical_options:
+
+        # test to check if one list is contained within another as python doesn't have that out of the box
```
claalmve, line 999: Maybe specify `check if A is contained in B...` for clarity of checking which list is within which.

climakitae/core/data_interface.py
```diff
@@ -991,7 +995,24 @@ def _update_scenarios(self):
...
+            # convert list b to numpy array
+            b_arr = np.array(b)
+
+            for i in range(len(b_arr)):
+                if np.array_equal(a_arr, b_arr[i : i + len(a_arr)]):
+                    return True
+            return False
+
+        # check if new selection has the historical scenario options and if not select the first new option
+        if _check_list_contained(_scenario_historical, scenario_historical_options):
```
claalmve, line 1013: Is it possible to use sets here for comparing these two lists since both of these can only be a max length of 2 (and since we know what all the possible values)? 

Like `if set(_scenario_historical).issubset(scenario_historical_options):` ? Or is this logic too simplified?

elehmer: Yeah let me figure out how to do this without `numpy` silly to import it just for this

elehmer: I think I have fixed the function in [5401d61](https://github.com/cal-adapt/climakitae/pull/393/commits/5401d61950de5db57994917c9415197a7ed978f4)

claalmve:
Looks good to me! Thanks for implementing the changes @elehmer !

claalmve (approved):

Issue #299, opened 2023-12-04
unintuitive colorbar defaults


thenaomig:
This is showing up in warming_levels.ipynb, run with all of the defaults. The colorbar is a diverging one even though all of the data is positive, which results in a lot of blue pixels (implying colder temperatures) when every pixel is in-fact warmer.

vicford:
This is based on the default behavior in `_get_cmap`. Will modify this function to determine the min/max range first to establish which colormap + consideration of moisture variables which should have "blues" represent more moisture as following standard meteorological convention.

PR #407, opened 2024-08-02
Improve visualizations in the first panel of wl.visualize

nicolejkeeney:
# Improve visualizations in the first panel ("maps of individual simulations") in wl.visualize()

I improved the visualizations for the first panel in ```wl.visualize()```. I improved the colormap, addressing this old [issue](https://github.com/cal-adapt/climakitae/issues/299) Naomi raised. I also added a colorbar and reasonable plot titles. For data with more than 4 simulations, a single plot is shown with a dropdown widget for toggling between simulations. For data with 4 or fewer simulations, postage stamp maps are created. 

I also added the ability for a scatter plot to be generated for data that is only a single grid cell. It's not that informative, but at least it shows you something, and doesn't throw an error :) 

**To test**: Run the warming_levels.ipynb notebook up until ```wl.visualize()``` (section 1c in the notebook). Try a few different data options. 

**NOTE**: I wasn't able to get to the second tab-- the one with the stats-- before I headed out of town. I'll fix that one up when I return :D 

## Type of change
Subtle visualization improvements 
None of the below applicable

# How Has This Been Tested?

n/a
- [ ] Test A
- [ ] Test B

# Checklist:

- [] My changes generate no new warnings
- [n/a] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/climakitae/explore/warming.py
+++ b/climakitae/explore/warming.py
@@ -11,6 +11,9 @@ import param
 import panel as pn
 import dask
 import calendar
+import warnings
+
+warnings.simplefilter(action="ignore", category=FutureWarning)
 
 from climakitae.core.data_view import compute_vmin_vmax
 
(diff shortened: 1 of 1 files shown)


vicford (changes requested):
Amazing @nicolejkeeney -- thank you for tackling this on. 

A few nitpicky things: 
- If we can re-add the variable and unit label to the colorbar (like what is on the second tab), to reaffirm what variable is being displayed and units
- At 4°C WL (and some in 3°C) (in WRF) only 1 model actually has data, but the rest of the postage stamps are correctly empty, but it's a bit funny that all model options show up. Can we modify this so only postage stamps for models that do have valid data are provided? 
- Looks like the ae_diverging colormap isn't quite correctly applied with the vmin characterization; see screenshot -- this is with mean wind speed at 10m
Screenshot 2024-08-02 at 12 26 39 PM

claalmve:
I love these changes to the warming level plots! This is a much more intuitive interpretation of the postage plots with the same scale on the colorbar.

For some reason, I'm getting this error when running `warming_levels` on a smaller lat/lon subset:
image

Lat/lon params changed:
```
wl.wl_params.latitude=(34.0,34.03)
wl.wl_params.longitude=(-117.5,-117.45)
```

climakitae/explore/warming.py
```diff
@@ -697,101 +722,163 @@ def warming_levels_select(self):
...
-                    # Creating histogram since all simulations are too many to put on a bar plot
-                    all_plots = all_plot_data.hvplot.hist()
+        # If there are less than or equal to four simulations, make postage stamps
+        if len(data_to_plot.simulation.values) <= 4:
 
-            try:
-                all_plots.opts(
-                    title=wl_viz.wl_params.variable
+            # if there's only one data point, make a scatter plot
+            if len(data_to_plot.x.values) == 1 and len(data_to_plot.y.values) == 1:
```
claalmve, line 767: This breaks in `warming_levels_approach` when `data_to_plot` comes from a Statistical downscaling method.

nicolejkeeney: Great thanks!

nicolejkeeney:
> I love these changes to the warming level plots! This is a much more intuitive interpretation of the postage plots with the same scale on the colorbar.
> 
> For some reason, I'm getting this error when running `warming_levels` on a smaller lat/lon subset: image
> 
> Lat/lon params changed:
> 
> ```
> wl.wl_params.latitude=(34.0,34.03)
> wl.wl_params.longitude=(-117.5,-117.45)
> ```

I think this isn't related to the panel at all, the workflow raises an error in the steps above the panel when trying to load the data because there is no data in that subset. 

We should probably have ```wl.calculate()``` throw an informative error if this is the case. It prints the text below, then raises a Value Error. 
Screenshot 2024-08-22 at 9 02 24 AM

PR #418, opened 2024-09-03
Unit conversion happening twice

nicolejkeeney:
# Description of PR

Looks like unit conversion was happening two times in the data retrieval code. An error was being raised for monthly precip data retrieval in the first conversion because the frequency attribute had not yet been added by that time in the workflow, raising an error for unit conversion. 

I'm not sure if this may raise other issues down the line. I tested a few conversions and it seems fine but maybe there are things I'm not thinking of. Time will tell...

## Type of change

# How Has This Been Tested?

- [ ] Test A
- [ ] Test B

# Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings

### DIFF
--- a/climakitae/core/data_load.py
+++ b/climakitae/core/data_load.py
@@ -596,8 +596,8 @@ def _merge_all(selections, data_dict):
     all_ssps = all_ssps[var_id]
 
     # Convert units:
-    all_ssps = _override_unit_defaults(all_ssps, var_id)
-    all_ssps = convert_units(da=all_ssps, selected_units=selections.units)
+    # all_ssps = _override_unit_defaults(all_ssps, var_id)
+    # all_ssps = convert_units(da=all_ssps, selected_units=selections.units)
 
     return all_ssps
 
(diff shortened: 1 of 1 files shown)


nicolejkeeney:
[Assessing precip units issue](https://trello.com/c/DLbnk75l/708-assessing-precip-units-issue)

vicford (approved):
Now working in `historical_climate_data_comparisons` notebook!!

claalmve:
I looked into the call stacks of this PR and `cksplit2a` when retrieving precip data, and this fix makes sense with me! It looks like the extra `convert_units` is happening on each individual dataset before they're merged together, and frequency is only added after the merge, so this fix makes sense to fix the issue at hand. What I'm not too sure about is why this only happens with precip retrieval and not other variables. Otherwise, LGTM!

claalmve (approved):

nicolejkeeney:
> I looked into the call stacks of this PR and `cksplit2a` when retrieving precip data, and this fix makes sense with me! It looks like the extra `convert_units` is happening on each individual dataset before they're merged together, and frequency is only added after the merge, so this fix makes sense to fix the issue at hand. What I'm not too sure about is why this only happens with precip retrieval and not other variables. Otherwise, LGTM!

Thanks for checking the logic!! I'll merge it now

PR #420, opened 2024-09-04
Update warming.py

claalmve:
I accidentally left some code that's not being used in `warming.py`.


vicford (approved):
not being used anywhere

PR #423, opened 2024-09-11
Quick fix bug with UnboundLocalError

claalmve:
# Description of PR

Didn't define `one_da` for Air Temp case for 1 in X. Fixing that in this PR.


vicford (approved):
Great catch. Was able to reproduce issue in main.

PR #443, opened 2024-10-03
Revert "SCE Pilot- Upgrading CAVA pilot to focus on batch mode to reduce code complexity"

claalmve:
Reverts cal-adapt/climakitae#442 because LOCA batch mode doesn't still doesn't work


vicford (approved):
Thanks @claalmve. Tested on both WRF and LOCA, single location and 2+ locations and not seeing the issue anymore

PR #455, opened 2024-10-31
Fixing WL simulation filtering

claalmve:
# Description of PR

### Summary of changes and related issue
Currently, the filtering function in WL data approach doesn't work because it pulls both UCLA/UCSD grids for WRF data. This caused @nicolejkeeney and I to look further into the function, and we realized the logic of this code already existed in `_get_cat_subset`, so we decided to re-use that functionality and add a `historical` data removal filter.

The main changes include using `_get_cat_subset` in the `drop_invalid_sims` function, and moving `get_cat_subset` to utils since it's used more than just inside `data_load.py`.

We need to combine data across SSPs for a WL approach but we don't want to create dummy simulations that don't actually exist (i.e. for 9 km WRF data, only 1 2-4.5 and 5-8.5 models exist, while 8 3-7 models exist, which would create 24 simulations. We only want 10 simulations, so this extra filtering has to happen after simulations are combined across SSPs too).

### Type of change

<br>

#### Practical
- [ ] 80% unit test coverage
- [ ] Documentation
  - [ ] All functions/adjusted functions documented in the [readthedocs](https://climakitae.readthedocs.io/en/latest/).
  - [ ] Documentation is pushed
- [x] Complex code commented
- [x] Context of function is clearly provided
  - [ ] Intent of function is provided
  - [ ] How to test, so that it is not siloed on scientists and anyone can review
  - [ ] Appropriate manual testing was completed
- [x] Linting completed and resolved

#### Conceptual
- [x] Doesn't replicate existing functionality
- [x] Aligns with general coding standard of existing functions
- [x] Matches desired functinonality from users/scientists


nicolejkeeney (changes requested):
Looks good to me, thanks Calvin. Seems a lot cleaner now. Left a small comment. Feel free to merge after your fix that

PR #460, opened 2024-11-06
Adding WL 1.2 to existing GWL times files

claalmve:
# Description of PR

### Summary of changes and related issue
[What's changed in this PR?]

[Why did you change this and what applicable context is needed to understand why this change is needed?]

### Type of change

<br>

#### Practical
- [ ] 80% unit test coverage
- [ ] Documentation
  - [ ] All functions/adjusted functions documented in the [readthedocs](https://climakitae.readthedocs.io/en/latest/).
  - [ ] Documentation is pushed
- [ ] Complex code commented
- [ ] Context of function is clearly provided
  - [ ] Intent of function is provided
  - [ ] How to test, so that it is not siloed on scientists and anyone can review
  - [ ] Appropriate manual testing was completed
- [ ] Linting completed and resolved

#### Conceptual
- [ ] Doesn't replicate existing functionality
- [ ] Aligns with general coding standard of existing functions
- [ ] Matches desired functinonality from users/scientists


nicolejkeeney (commented):
Checked retrieving WL data for these new levels with both the DataParameters() GUI and the `get_data()` function, and this works fine! Thanks!

vicford (changes requested):
Works just fine in `get_data` functionality! 

As is required (@mdkoenig made me do it):
- Documentation update: N/A
- Unit tests update: Are there unit tests for warming levels that we can test for these new warming levels? If so, I believe `tests/create_test_dataset_warming_level_hourly.py` needs to be updated and whatever specific tests as well.

claalmve:
In this PR is an example usage of a constant `WARMING_LEVELS` in constants.py, that is then imported in `data_interface.py`. If this looks good, I can change all WL references to this `WARMING_LEVELS` reference in this PR.

vicford (approved):
Was able to retrieve 0.8 and 1.2 in DataParameters GUI and `get_data`; and tests successfully ran on the Hub. I like the constants script, although I imagine we may have very few to put in there beyond `WARMING_LEVELS`, happy to defer to other dev opinions on that.

climakitae/core/constants.py
```diff
@@ -0,0 +1,4 @@
+# constants.py
+"""This module defines constants across the codebase"""
+
+WARMING_LEVELS = [0.8, 1.2, 1.5, 2.0, 2.5, 3.0, 4.0]
```
nicolejkeeney, line 4: I really like the idea of a `constants.py` file, however, I think it's implementation at this stage is incomplete. Doing a search across climakitae, you can see that the string "1.5, 2.0", for example, is found in 11 files: https://github.com/search?q=repo%3Acal-adapt%2Fclimakitae%201.5%2C%202.0&type=code

To properly make use of a `constants.py` file, we should be replacing anywhere the warming levels list is used with these constants, like you've done in `data_interface.py` here. Which, you could do in this PR if you want, or you could remove constants.py here and do that entire change in a separate PR (which I am more in favor of, personally). 

I'm not sure if we need to be actively updating some of the files from that link with the newest things. For example, I'm not sure if we need to be updating `vulnerability.py`-- perhaps check in with Mark about that? Same with `uncertainty.py`. I don't know what [this file]([climakitae/data/gwl_1981-2010ref_EC-Earth3_ssp370.csv](https://github.com/cal-adapt/climakitae/blob/afc2dc2618240a85d31cab51c80dd9b7b7b0daf5/climakitae/data/gwl_1981-2010ref_EC-Earth3_ssp370.csv)) is but it needs (?) to be updated as well.

**TLDR**: If we don't want to update anything but `data_interface.py`, I don't think it makes sense to have a `constants.py` file, and I think it overcomplicates the code. However, if we want to update all the other files that use warming levels, the `constants.py` file should be read in and used all those files as well (i.e. replace all the strings).

tests/util/test_util.py
```diff
@@ -0,0 +1,75 @@
...
+@pytest.mark.parametrize("name, df", gwl_dfs)
+def test_unique_combinations(name, df):
+    """
+    Test that the combination of 'GCM', 'run', and 'scenario' columns is unique.
+
+    - Ensures there are no duplicate rows based on these columns.
+    """
+    assert (
+        df.duplicated(subset=["GCM", "run", "scenario"]).sum() == 0
+    ), f"Duplicate combinations of 'GCM', 'run', and 'scenario' found in DataFrame '{name}'."
```
nicolejkeeney, line 39: I like this check!

nicolejkeeney (changes requested):
I left a long (sorry!) comment about `constants.py`. I really like your unit tests!

nicolejkeeney:
> Was able to retrieve 0.8 and 1.2 in DataParameters GUI and `get_data`; and tests successfully ran on the Hub. I like the constants script, although I imagine we may have very few to put in there beyond `WARMING_LEVELS`, happy to defer to other dev opinions on that.

I'd like to see the SSP strings in the `constants.py` file if we chose to use this method of organizing constants (which I think is a good idea if we flesh it out more).

tests/util/test_util.py
```diff
@@ -0,0 +1,75 @@
+import pytest
+import pandas as pd
+from climakitae.util.utils import read_csv_file
+from climakitae.core.constants import WARMING_LEVELS
+from climakitae.core.paths import gwl_1850_1900_file, gwl_1981_2010_file
```
nicolejkeeney, line 5: Can you also add a brief string description at the very top of the file? Like in this test file: https://github.com/cal-adapt/climakitae/blob/afc2dc2618240a85d31cab51c80dd9b7b7b0daf5/tests/data_interface/test_variable_descriptions_csv.py#L1

mdkoenig:
Very demure, very DRY! <3

climakitae/core/constants.py
claalmve: Yeah, I just replaced it in one place for now to show how it could be used and see people's thoughts before replacing it all the other places too. 

If we do go forward with constants.py, I think that we should replace all references of warming levels with this constant, and within specific files that may not use all warming levels (like `vulnerability.py` or `uncertainty.py`), we can modify the constant in-place within that file so that it's clearly demonstrated that _specific_ warming levels were extracted from all the potential warming levels in our codebase.

claalmve: I can also separate a `constants.py` into a different PR and get this one across the finish line without that first.

PR #462, opened 2024-11-13
Enable non-GUI view and retrieval of derived variables/indices

nicolejkeeney:
# Description of PR
Enable users to view and retrieve derived variables/indices without using the DataInterface GUI. This wasn't previously an option. 

### Summary of changes and related issue
- Derived variables/indices are now available for viewing using the existing function ```get_data_options()```
- Derived variables/indices are now available for retrieval using the existing function ```get_data()```
- Added a new column "dependencies" to the variable_descriptions.csv file, which is used to determine variable dependencies for the derived variables/indices. This helped me determine the correct subset/combo of options for each variable to return to the user. 
- Wrote some unit tests (hell yeah!!) 

### Relevant motivation and context
When I wrote the two functions ```get_data_options()``` and ```get_data()```, I didn't realize they didn't allow you to view/retrieve derived variables; you can only access the catalog variables. Calvin mentioned he needed to get the derived variables using the ```get_data()``` function, so I took the opportunity to improve the function. 

**NOTE**: This required a shocking amount of code... the logic in climakitae for accessing the correct subset/combination of options for each variable is really complex. 

These are the variable ("display") names of all the derived variables: 
- Relative humidity [hourly only-- daily/monthly is native in the catalog]
- Wind speed at 10m
- Wind direction at 10m
- Dew point temperature
- Specific humidity at 2m
- Dew point temperature
- Fosberg fire weather index
- Effective Temperature
- NOAA Heat Index

### How to test 
1) Run through the notebook ```climakitae_direct_data_download.ipynb``` to ensure that it still works correctly, since I modified the functions used in that notebook 
2) Try retrieving some derived variables/indices. For example: 

```
# See all the data combinations (i.e. different resolution, scenario, etc.) for the Fosberg fire weather index
get_data_options(variable = "Fosberg fire weather index")

# Retrieve data for the Fosberg fire weather index 
get_data(
    variable = "Fosberg fire weather index", 
    scenario = "Historical Climate", 
    timescale = "hourly", 
    resolution = "9 km",
    downscaling_method = "Dynamical", 
    time_slice = (1990,1991)
)

# See all the data combinations (i.e. different resolution, scenario, etc.) for Dew point temperature
get_data_options(variable = "Dew point temperature")

# Retrieve data for Dew point temperature 
get_data(
    variable = "Dew point temperature", 
    scenario = ["Historical Climate","SSP 2-4.5 -- Middle of the Road"], 
    timescale = "monthly", 
    resolution = "45 km",
[...]

PR #467, opened 2024-11-23
Retrieve derived indices using a warming levels approach

nicolejkeeney:
# Description of PR
Retrieve derived indices using a warming levels approach. And, some unit tests. 

I previously thought I had implemented this feature in PR #462 , but I forgot to enable the ability to retrieve the data using both a time-based and a warming level approach. 

### How to test 
Try to retrieve the Derived Indices (under the "Derived Indices" button in the GUI) using a Warming Level approach. Try with both the DataInterface GUI(i.e. `selections.show()`)  and with `get_data()`. 

The following are the derived variable options: 
- Fosberg fire weather index
- Effective Temperature
- NOAA Heat Index

### Type of change

#### Practical
- [x] 80% unit test coverage
- [x] Documentation
  - [x] All functions/adjusted functions documented in the [readthedocs](https://climakitae.readthedocs.io/en/latest/).
  - [x] Documentation is pushed
- [x] Complex code commented
- [x] Context of function is clearly provided
  - [x] Intent of function is provided
  - [x] How to test, so that it is not siloed on scientists and anyone can review
  - [x] Appropriate manual testing was completed
- [x] Linting completed and resolved

#### Conceptual
- [x] Doesn't replicate existing functionality
- [x] Aligns with general coding standard of existing functions
- [x] Matches desired functionality from users/scientists

### DIFF
--- a/climakitae/core/data_interface.py
+++ b/climakitae/core/data_interface.py
@@ -1672,7 +1672,11 @@ def get_data_options(
     # Raise error for bad input from user
     for user_input in [variable, downscaling_method, resolution, timescale]:
         if (user_input is not None) and (type(user_input) != str):
-            print("Function arguments require a single string value for your inputs")
+            print(
+                _format_error_print_message(
+                    "Function arguments require a single string value for your inputs"
+                )
+            )
             return None
 
     def _list(x):
@@ -1701,7 +1705,7 @@ def get_data_options(
         & (cat_df["scenario"].isin(d["scenario"]))
     ].reset_index(drop=True)
     if len(cat_subset) == 0:
-        print("No data found for your input values")
+        print("ERROR: No data found for your input values")
         return None
 
     if tidy:
@@ -1797,6 +1801,11 @@ def get_subsetting_options(area_subset="all"):
     return geoms_df
 
 
+def _format_error_print_message(error_message):
+    """Format error message using the same format"""
+    return "ERROR: {0} \nReturning None".format(error_message)
+
+
 def get_data(
     variable,
     downscaling_method,
@@ -1880,11 +1889,6 @@ def get_data(
     """
 
     # Internal functions
-
-    def _format_error_print_message(error_message):
-        """Format error message using the same format"""
-        return "ERROR: {0} \nReturning None".format(error_message)
-
     def _error_handling_warming_level_inputs(wl, argument_name):
         """Error handling for arguments: warming_level and warming_level_month
         Both require a list of either floats or ints
@@ -1988,7 +1992,9 @@ def get_data(
 
     # Get intake catalog and variable descriptions from DataInterface object
     data_interface = DataInterface()
-    var_df = data_interface.variable_descriptions
+    var_df = data_interface.variable_descriptions.rename(
+        columns={"variable": "display_name"}
+    )  # Rename column so that it can be merged with cat_df
     catalog = data_interface.data_catalog
     cat_df = _get_user_friendly_catalog(
         intake_catalog=catalog, variable_descriptions=var_df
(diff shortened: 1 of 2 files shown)


vicford (changes requested):
`get_data` version:
- All variables were retrievable
- Noticed for effective temperature, if the timestep was hourly not daily it gives a non-specific error (see pic below)
Screenshot 2024-11-25 at 1 00 28 PM

GUI version:
- For derived index the warming level option disappears and defaults back to time

nicolejkeeney:
> `get_data` version:
> 
> * All variables were retrievable
> * Noticed for effective temperature, if the timestep was hourly not daily it gives a non-specific error (see pic below)

Effective temperature is only available for daily data, hence the error. The two other indices are only available for hourly. 
Screenshot 2024-11-25 at 7 02 56 PM

That is a bad error message though... I'll think on how to return a more informative error message. Good find

nicolejkeeney:
> GUI version:
> 
> * For derived index the warming level option disappears and defaults back to time

Can you elaborate more on when this happens?

vicford:
> > GUI version:
> > 
> > * For derived index the warming level option disappears and defaults back to time
> Can you elaborate more on when this happens? Not sure I follow.

Okay when I went to take a quick video of what was happening, I couldn't replicate so was probably user error on my part!

vicford (approved):
Just an update to that bad error message would be good!

nicolejkeeney:
> Just an update to that bad error message would be good!

You actually found a bigger bug that was affecting more than just the derived indices. Basically anytime you put a bad input for `timescale`, that confusing error is raised (this can happen for other variables as well, since some variables are only available for hourly or only daily/monthly). 

Anyways good find. I just fixed it, I think. And added a relevant unit test to check that the function raises an appropriate error-- or, in this case, a print message.

tests/data_interface/test_get_data.py
```diff
@@ -0,0 +1,125 @@
...
+                    cached_area="CA",
+                    approach="Time",
+                )
+            except:
+                pytest.fail(
+                    "Data for hourly Relative Humidity with a Time-based approach could not be retrieved"
+                )
+
+
+def TestAppropriateStringErrorReturnedIfBadInputGetData():
```
claalmve, line 75: Is this supposed to be `class`?

tests/data_interface/test_get_data.py
```diff
@@ -0,0 +1,125 @@
...
+    def test_error_raised_bad_timescale_input(self):
+        """Test that an appropriate message is printed if bad input for timescale
+        Previously, this input would raise a confusing error!
+        Fixed in November 2024 to return a useful print message and None instead
+        """
+
+        # Error message we expect to be printed by the function
+        expected_print_message = "ERROR: No data found for your input values\n"
+
+        # NOTE: function PRINTS this message-- it does not return it as en error
```
claalmve, line 110: nit: typo- "en error"

tests/data_interface/test_get_data.py
```diff
@@ -0,0 +1,125 @@
...
+def TestAppropriateStringErrorReturnedIfBadInputGetData():
+    """Test that an appropriate error message is printed to the user describing the issue and how to resolve it."""
+
+    def test_error_raised_string_input_warming_level(self):
+        """Warming level should be a float input! Make sure the function prints the appropriate error message"""
+
+        # Error message we expect to be printed by the function
+        expected_print_message = "ERROR: Function argument warming_level requires a float/int or list of floats/ints input. Your input: <class 'str'> \nReturning None\n"
+
+        # NOTE: function PRINTS this message-- it does not return it as en error
```
claalmve, line 84: nit: typo- "en error"

claalmve:
One consideration for the test side of this PR is that the current `get_data` tests take a couple mins to runs. This of course is not that long of a time, but I think we'd ideally want to reduce the amount of time each test takes as much as possible, as the number of tests will only accumulate over time. Would it be possible to either:

1. limit the spatial/temporal scope of the pytest tests or 
2. create dummy/pre-cached datasets that `get_data` grabs from instead of live retrieving datasets every time the test is run (I bring up 2 as a potential idea to instead create live data tests that wrap around the `retrieve` function rather than `get_data`, since `retrieve` is the common point of retrieving data between `DataParameters` and `get_data`).

climakitae/core/data_interface.py
```diff
@@ -1693,7 +1705,7 @@ def _list(x):
         & (cat_df["scenario"].isin(d["scenario"]))
     ].reset_index(drop=True)
     if len(cat_subset) == 0:
-        print("No data found for your input values")
+        print("ERROR: No data found for your input values")
```
vicford, line 1708: Minor: Can we add "Please modify your data request"

vicford (approved):

climakitae/core/data_interface.py
```diff
@@ -1664,7 +1672,11 @@ def get_data_options(
     # Raise error for bad input from user
     for user_input in [variable, downscaling_method, resolution, timescale]:
         if (user_input is not None) and (type(user_input) != str):
-            print("Function arguments require a single string value for your inputs")
+            print(
+                _format_error_print_message(
+                    "Function arguments require a single string value for your inputs"
```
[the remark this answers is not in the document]
nicolejkeeney: hmm not sure i follow? it seems clear to me, can you suggest an alternative that is more clear?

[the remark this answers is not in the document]
vicford: From the perspective of a less tech-savvy user, my first thought was "what is a single string". i.e., "single" means what?

nicolejkeeney:
> This PR works great for me! Retrieval through `get_data` and `DataParameters` both worked well. I love the addition of the derived indices and making sure `get_data` becomes more and more robust for us all to use.
> 
> One consideration for the test side of this PR is that the current `get_data` tests take a couple mins to run (increasing the total for the test suite to 4min). This of course is not that long of a time, but I think we'd ideally want to reduce the amount of time each test takes as much as possible, as the number of tests will only accumulate over time. Would it be possible to either:
> 
> 1. limit the spatial/temporal scope of the pytest tests or
> 2. create dummy/pre-cached datasets that `get_data` grabs from instead of live retrieving datasets every time the test is run (I bring up 2 as a potential idea because another place to potentially test live retrieval at is at the `retrieve` function directly rather than `get_data`, since `retrieve` is the common point of retrieving data between `DataParameters` and `get_data`).

Good point, I went with option 1 and reduced the temporal scope of the tests. Not sure we want to use dummy data here since the point of the function is to test the data retrieval step. Good idea to keep an eye on the total time elapsed for the tests so it doesn't get out of control at some point; I hadn't thought of that before.

PR #509, opened 2025-02-20
Export: Add Zarr local/s3 support, Remove NetCDF s3 support

elehmer:

This PR adds Zarr export option on the HUB user partition and to s3 scratch bucket. It also removes NetCDF export to s3 as that is limited to <2GB.

It also adds `ck.remove_zarr` function to help users delete Zarr exports on the HUB (Zarr stores are a tree of directories and not easily deleted from the HUB GUI).

Add Zarr support and fix broken export functions.

## How to test 

Export data to Zarr, NetCDF, CSV and reload on the HUB.

## Type of change
- [ x ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ x ] This change requires a documentation update
- [ ] None of the above  

## To-Do
- [ ] Unit tests
- [ ] Documentation
  - [ ] Complex code commented
- [ ] Black formatting has been utilized

### DIFF
--- a/climakitae/__init__.py
+++ b/climakitae/__init__.py
@@ -1,5 +1,5 @@
 from climakitae.core.data_load import load
-from climakitae.core.data_export import export
+from climakitae.core.data_export import export, remove_zarr
 
 try:
     from importlib.metadata import version as _version
@@ -18,6 +18,7 @@ __all__ = (
     # Methods
     "load",
     "export",
+    "remove_zarr",
     # Constants
     "__version__",
 )
--- a/climakitae/core/data_export.py
+++ b/climakitae/core/data_export.py
@@ -200,6 +200,16 @@ def _compression_encoding(data):
     return compdict
 
 
+def _convert_da_to_ds(data):
+    if isinstance(data, xr.core.dataarray.DataArray):
+        if not data.name:
+            # name it in order to call to_dataset on it
+            data.name = "data"
+        return data.to_dataset()
+    elif isinstance(data, xr.core.dataset.DataSet):
+        return data
+
+
 def _export_to_netcdf(data, save_name):
     """
     Export user-selected data to NetCDF format.
@@ -223,12 +233,7 @@ def _export_to_netcdf(data, save_name):
     print("Exporting specified data to NetCDF...")
 
     # Convert xr.DataArray to xr.Dataset so that compression can be utilized
-    _data = data
-    if isinstance(_data, xr.core.dataarray.DataArray):
-        if not _data.name:
-            # name it in order to call to_dataset on it
-            _data.name = "data"
-        _data = _data.to_dataset()
+    _data = _convert_da_to_ds(data)
 
     est_file_size = _estimate_file_size(_data, "NetCDF")
     disk_space = shutil.disk_usage(os.path.expanduser("~"))[2] / bytes_per_gigabyte
@@ -286,12 +291,7 @@ def _export_to_zarr(data, save_name, mode):
     print("Exporting specified data to Zarr...")
 
     # Convert xr.DataArray to xr.Dataset so that compression can be utilized
-    _data = data
-    if isinstance(_data, xr.core.dataarray.DataArray):
-        if not _data.name:
-            # name it in order to call to_dataset on it
-            _data.name = "data"
-        _data = _data.to_dataset()
+    _data = _convert_da_to_ds(data)
 
     est_file_size = _estimate_file_size(_data, "Zarr")
     disk_space = shutil.disk_usage(os.path.expanduser("~"))[2] / bytes_per_gigabyte
--- a/climakitae/core/data_export.py
+++ b/climakitae/core/data_export.py
@@ -297,9 +297,8 @@ def _export_to_netcdf(data, save_name, mode):
                     "or specify a new file name here."
                 )
             )
-        #encoding = _fillvalue_encoding(_data) | _compression_encoding(_data)
-        encoding = _compression_encoding(_data)
-        _data.to_netcdf(path, encoding=encoding)
+        encoding = _fillvalue_encoding(_data) | _compression_encoding(_data)
+        _data.to_netcdf(path, format="NETCDF4" engine="netcdf4", encoding=encoding)
         print(
             (
                 "Saved! You can find your file in the panel to the left"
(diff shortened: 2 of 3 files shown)


vicford (approved):
Nice! Tested on csv, .nc, and a local zarr; and the `remove_zarr` function worked well. We should add the "remove_zarr" as some markdown into at least `getting_started` in its revision.

elehmer:
> Nice! Tested on csv, .nc, and a local zarr; and the `remove_zarr` function worked well. We should add the "remove_zarr" as some markdown into at least `getting_started` in its revision.

Very good idea. I will make a branch with it added. And I will add a reference in the `readthedocs`

climakitae/__init__.py
```diff
@@ -18,6 +18,7 @@
     # Methods
     "load",
     "export",
+    "remove_zarr",
```
claalmve, line 21: How come you add `remove_zarr` to the init? Is it so that you can get the `ck.remove_zarr` function call?

climakitae/core/data_export.py
```diff
@@ -67,226 +99,224 @@ def _warn_large_export(file_size, file_size_threshold=5):
...
-        def _list_n_none_to_string(dic):
-            """Convert list and None to string.
+    _unencode_missing_value(data)
+    for coord in data.coords:
+        _unencode_missing_value(data[coord])
 
-            Parameters
-            ----------
-            dic: dict
+    for data_var in data.data_vars:
```
claalmve, line 143: Much cleaner! Love this

climakitae/core/data_export.py
```diff
@@ -26,6 +27,37 @@
...
+    try:
+        shutil.rmtree(dir_path)
+        print(f"Zarr dataset '{dir_path}' deleted successfully.")
+    except FileNotFoundError:
+        print(f"Zarr dataset '{dir_path}' not found.")
+    except OSError as e:
+        print(f"Error deleting Zarr dataset '{dir_path}': {e}")
+
+
+def _add_metadata(data):
```
claalmve, line 40: Is `_add_metadata` replacing `_update_attributes`?

climakitae/core/data_export.py
```diff
@@ -67,226 +99,224 @@ def _warn_large_export(file_size, file_size_threshold=5):
...
-        for data_var in data.data_vars:
-            _unencode_missing_value(data[data_var])
+    Returns
+    -------
+    None
+    """
+    print("Exporting specified data to NetCDF...")
+
+    # Convert xr.DataArray to xr.Dataset so that compression can be utilized
+    _data = data
```
claalmve, line 204: Since lines 203 - 218 are almost the same as lines 267 - 281 (aside from a 'NetCDF' vs 'Zarr' string), could we combine both logics into one helper function? Just for conciseness and to make it easier to update this logic in the future

claalmve (changes requested):
Looks good to me! I left a small comment about combining two code pieces together to one helper function to increase code readability. I love how sectioned the different export methods are now, it is very intuitive to follow. 

I left my review as Request Changes because of a weird error that could potentially come up when interrupting the export function in the middle of its execution (sent more details over Slack).

climakitae/__init__.py
elehmer: Yes, that is where package level functions are referenced.

climakitae/core/data_export.py
elehmer: `_update_attributes` was only needed for exporting `netCDF` to S3, so it is gone. `_add_metadata` is the `DataSet` level metadata that was being written to the original dataset (and therefore was still there after export). I have moved that to this function that is run on the copy of the data so that doesn't happen any more.

climakitae/core/data_export.py
elehmer: Yes, that would be a good thing to adopt. I will work on consolidating the code. Thx

claalmve:
Sorry, one more comment. I noticed when I tried to export a NetCDF to s3, it automatically would start exporting locally (which is correct behavior). However, as a user, I don't get any bigger notification that tells me that this file is now actually exporting locally rather than to s3. Would it be possible to add some bigger text or warning that signals to the user that the NetCDF is _actually_ exporting locally instead of to s3 like they pass in?

image

elehmer:
> Sorry, one more comment. I noticed when I tried to export a NetCDF to s3, it automatically would start exporting locally (which is correct behavior). However, as a user, I don't get any bigger notification that tells me that this file is now actually exporting locally rather than to s3. Would it be possible to add some bigger text or warning that signals to the user that the NetCDF is _actually_ exporting locally instead of to s3 like they pass in?
> 
> image

Yes, I will add something to inform the user it isn't going to S3

elehmer:
> > Nice! Tested on csv, .nc, and a local zarr; and the `remove_zarr` function worked well. We should add the "remove_zarr" as some markdown into at least `getting_started` in its revision.
> 
> Very good idea. I will make a branch with it added. And I will add a reference in the `readthedocs`

Made some changes in: https://github.com/cal-adapt/cae-notebooks/pull/148

climakitae/core/data_export.py
elehmer: I have made a new func `_convert_da_to_ds` to address this in: [4925c91](https://github.com/cal-adapt/climakitae/pull/509/commits/4925c91c4856eac9fe5ac0e3eb978d7e3bb4f057)

elehmer:
> > Sorry, one more comment. I noticed when I tried to export a NetCDF to s3, it automatically would start exporting locally (which is correct behavior). However, as a user, I don't get any bigger notification that tells me that this file is now actually exporting locally rather than to s3. Would it be possible to add some bigger text or warning that signals to the user that the NetCDF is _actually_ exporting locally instead of to s3 like they pass in?
> > image
> 
> Yes, I will add something to inform the user it isn't going to S3

Fixed in: https://github.com/cal-adapt/climakitae/pull/509/commits/4534739f371f5b02eb76f24ef00f737eec2931e6 by exiting if user choses mode='s3' and not format='Zarr'

elehmer:
To address the issue where Zarr export to S3 is interrupted or you try to write over an existing Zarr in S3, instead of trying to delete the keys (Zarr) in S3 an error is returned saying that the Zarr already exists in S3. This issue is because we don't wait for the keys to be deleted before rewriting the Zarr to S3. We could wait, but I think it is better to not overwrite any existing dataset.

elehmer:
@vicford @claalmve Issues should be addressed now. Let me know what you think. Thanks for the help!

vicford (approved):

claalmve:
Looks good to me! Thanks for making the changes @elehmer !

PR #518, opened 2025-03-12
Time-optimization changes to CAVA data function

claalmve:
This PR focuses on the 1-in-X calculation for the CAVA data function. It addresses SCE's concerns with a time bottleneck by parallelizing the compute of 1-in-X calculations across all points at the same time, and also allows for multiple X parameters to be passed into the 1-in-X parameter.

This reduces the overall runtime of the 1-in-X calculation from N points X M simulations X P 1-in-X values to now just M simulations run time. The next PR for time optimization will focus on reducing the M simulations runtime to just 1 simulation run time.

## How to test 
Run `vulnerability_assessment.ipynb` with multiple X's in `one_in_x` param, like

`one_in_x=np.array([20, 100])`

and check that the multiple X calculations are working properly, and that the `one_in_x` distributions are not being fit on the points one by one when `batch_mode=True`.

Use the latest notebook in https://github.com/cal-adapt/cae-notebooks/pull/150

## Type of change
- [ ] None of the above  

## To-Do
- [ ] Unit tests
- [x] Documentation
  - [ ] Complex code commented

### DIFF
--- a/climakitae/explore/vulnerability.py
+++ b/climakitae/explore/vulnerability.py
@@ -831,7 +831,7 @@ def cava_data(
                 for pt_idx in range(len(metric_data)):
                     _export_no_e(
                         metric_data[pt_idx],
-                        filename=f"{clean_params['calc_name']}_{str(round(metric_data[pt_idx].lat.item(), 3)).replace('.', '')}N_{round(str(metric_data[pt_idx].lon.item(), 3)).replace('.', '')}W",
+                        filename=f"{clean_params['calc_name']}_{str(round(metric_data[pt_idx].lat.item(), 3)).replace('.', '')}N_{str(round(metric_data[pt_idx].lon.item(), 3)).replace('.', '')}W",
                         format=file_format,
                     )  # Will need to include naming convention for calculated file too.
 
--- a/climakitae/explore/vulnerability.py
+++ b/climakitae/explore/vulnerability.py
@@ -225,8 +225,10 @@ def _metric_agg(
                 sample_size=len(ams),
             )
 
-            # Change `arg_value` dimension coordinates to `1-in-X` parameter
-            calc_val.assign_coords(arg_value=one_in_x)
+            # Change `arg_value` dimension coordinates to `1-in-X` parameter and rename to `one_in_x`
+            calc_val = calc_val.assign_coords(arg_value=one_in_x).rename(
+                {"arg_value": "one_in_x"}
+            )
 
         elif percentile or percentile == 0:
 
--- a/climakitae/explore/vulnerability.py
+++ b/climakitae/explore/vulnerability.py
@@ -186,7 +186,7 @@ def _metric_agg(
                     duration=duration,
                     groupby=groupby,
                     check_ess=False,
-                ).squeeze()  # `get_return_value` is expecting a 1-D array
+                ).squeeze()
 
                 # Calculating 1-in-X return value
                 return_values = get_return_value(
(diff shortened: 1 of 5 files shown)


vicford (changes requested):
Wow that calculation was wicked fast! Cool!

A few things:
- My **strong** preference is to not have the `np.array([10, 100])` piece in the front end. I'd rather it be a list instead (like what we do similar for the selecting a single location), and in the back end it is passed as `np.array()`
   - i.e., `one_in_x = [10, 100]` not `one_in_x = np.array([10, 100])`
- Noticed that for the goodness of fit it now outputs per simulation but doesn't actually provide the value like it did previously? I imagine that that's because of the multiple return periods now?
- Is there anyways we can modify the output text instead of `Calculating 1-in-[ 10 100] year` to be more user-friendly looking? something like `Calculating 1-in-10 year 1-day max air temperature at 2m and 1-in-100 year 1-day max air temperature at 2m`
   - I'm assuming the `1-day` piece is now being added in because of the calculation in the backend? I didn't think it was like that previously
- The exported file names are off 
   - Preference here is to actually have something like: 
      - `one_in_multix_1_day_temperature` or
      - `one_in_10-100_1_day_temeprature` (this i think is clearer)
Screenshot 2025-03-13 at 12 24 52 PM

- I like how the `one_in_x` is added as a dimension, that's clever. Once this gets merged in I can modify the technical methodology to update on what's going on for the p_values to be specific (i.e., its the fit of the entire distribution). I'll update the markdown in the notebook to reflect too (I think it's better to separate examples here).

climakitae/explore/vulnerability.py
```diff
@@ -291,7 +288,9 @@ class CavaParams(param.Parameterized):
     heat_idx_threshold = param.Number(
         default=None, allow_None=True, doc="Threshold for heat index"
     )
-    one_in_x = param.Integer(default=None, allow_None=True, doc="1-in-x year event")
+    one_in_x = param.Parameter(
+        default=None, allow_None=True, doc="1-in-x year event(s) — number or list"
```
vicford, line 292: doc here reinforces my comment about passing in a list not an np.array of a list

vicford:
also linting is breaking

elehmer:
I was able to run through the entire notebook successfully. Memory peaked at about 24GB RAM but came down - seems that memory would go up and then drop back down.

elehmer (approved):

claalmve:
> Wow that calculation was wicked fast! Cool!
> 
> A few things:
> 
> * My **strong** preference is to not have the `np.array([10, 100])` piece in the front end. I'd rather it be a list instead (like what we do similar for the selecting a single location), and in the back end it is passed as `np.array()`
>   
>   * i.e., `one_in_x = [10, 100]` not `one_in_x = np.array([10, 100])`
> * Noticed that for the goodness of fit it now outputs per simulation but doesn't actually provide the value like it did previously? I imagine that that's because of the multiple return periods now?
> * Is there anyways we can modify the output text instead of `Calculating 1-in-[ 10 100] year` to be more user-friendly looking? something like `Calculating 1-in-10 year 1-day max air temperature at 2m and 1-in-100 year 1-day max air temperature at 2m`
>   
>   * I'm assuming the `1-day` piece is now being added in because of the calculation in the backend? I didn't think it was like that previously
> * The exported file names are off
>   
>   * Preference here is to actually have something like:
>     
>     * `one_in_multix_1_day_temperature` or
>     * `one_in_10-100_1_day_temeprature` (this i think is clearer)
>       Screenshot 2025-03-13 at 12 24 52 PM
> * I like how the `one_in_x` is added as a dimension, that's clever. Once this gets merged in I can modify the technical methodology to update on what's going on for the p_values to be specific (i.e., its the fit of the entire distribution). I'll update the markdown in the notebook to reflect too (I think it's better to separate examples here).

Sorry meant to this before @vicford :
1. Done! `cava_data` can take lists instead of numpy arrays now.
2. Yes, because of multiple return periods now.
3. Output text fixed!
4. Exported file names fixed!
5. Sounds good!

vicford (approved):

PR #540, opened 2025-04-15
Test Coverage for `util.dev_logging`

neilSchroeder:
Added test coverage for `climakitae.util.dev_logging`
Also changed the structure of the `_log` function to add `__wrapped__` and `_is_logged` parameters to be able to catch wrapped functions and remove the wrapper later without having to reload the module.

Test coverage is important. 
The other changes are motivated by code hygiene and philosophy. You shouldn't have to reload your module to remove the wrapped function.

## How to test 
```bash
pip install pytest-cov
pytest tests/util/test_dev_logging.py --cov=climakitae.util.dev_logging --cov-report term-missing
```

## Type of change
- [ ] None of the above  

## To-Do
- [x] Unit tests
- [x] Documentation
  - [x] Complex code commented
  - [ ] Helper functions hidden with `_` before the name: NOTE removed functions from hidden state since they're supposed to be called by a user (even if that user is a dev).

### DIFF
--- a/climakitae/util/dev_logging.py
+++ b/climakitae/util/dev_logging.py
@@ -128,4 +128,3 @@ def _disable_lib_logging(module: types.ModuleType):
             original_func = res.__wrapped__
             setattr(module, name, original_func)
             delattr(res, "_is_logged")
-    return
--- a/climakitae/util/dev_logging.py
+++ b/climakitae/util/dev_logging.py
@@ -25,7 +25,6 @@ Some limitations of this logger include:
 
 """
 
-import importlib
 import time
 import types
 
@@ -74,7 +73,7 @@ def _log(func: types.FunctionType) -> types.FunctionType:
     return _wrapper
 
 
-def _enable_lib_logging(obj: types.ModuleType):
+def enable_lib_logging(obj: types.ModuleType):
     """
     Adds the `log` wrapper to all functions and sub-classes within the given module or class.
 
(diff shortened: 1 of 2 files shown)


climakitae/util/dev_logging.py
```diff
@@ -27,20 +27,33 @@
...
+    Returns
+    -------
+    function
+        The wrapped function with logging capabilities.
     """
 
-    def _wrapper(*args, **kwargs):
-        """Wraps timer and print statement around functions if `lib_log_enabled` is True, otherwise
-        just return the function's result."""
+    def _wrapper(*args, **kwargs) -> types.FunctionType:
```
acordonez, line 52: Out of curiosity: Are there times where we would put type hints on *args and **kwargs parameters? Or is it better to leave them bare?

acordonez (approved):

climakitae/util/dev_logging.py
neilSchroeder: A tricky question. My personal take is no. This is a situation where `*args` and `**kwargs` are so ingrained in the definitions of python that it is understood implicitly that `*args` is a general tuple and `**kwargs` is a general dictionary. There's no way to know exactly what types these arguments are going to contain since this is a wrapper that can be used on _any_ function. Because these are here to indicate that you're passing a function through this wrapper, its better to leave them bare.

climakitae/util/dev_logging.py
```diff
@@ -49,16 +62,20 @@ def _wrapper(*args, **kwargs):
...
         end_time = time.time()
         print(
-            "    " * indentation_level
+            "\t" * indentation_level
             + f"Execution time for {func.__name__}: {end_time - start_time:.4g}"
         )
         return results
 
+    # Set the wrapper's attributes to maintain the original function's metadata
+    _wrapper.__wrapped__ = func
```
claalmve, line 71: I like this adding of attributes! Makes a lot of sense

claalmve (approved):

PR #551, opened 2025-05-02
Test coverage for `core.data_interface`

neilSchroeder:
Testing and small fixes to `core.data_interface`

Fixing bugs, cleaning things up for good hygiene, test coverage to 90%.
Lots of single lines that aren't covered by other things. This will mostly be redone this summer. So this is good enough for now.
## How to test 
```
pip install pytest-cov
pytest tests/data_interface/ --cov=climakitae.core.data_interface --cov-report term-missing
```
be warned that this test takes a long time to run due to `get_data`

## Type of change
- [ ] None of the above  

## To-Do
- [x] Unit tests
- [x] Documentation
  - [x] Complex code commented


acordonez (approved):

climakitae/core/data_interface.py
```diff
@@ -1927,70 +1993,75 @@ def _check_valid_input_station(stations, station_options_all):
...
+            if not isinstance(wl, list):
                 raise ValueError(
-                    "Function argument {0} requires a float/int or list of floats/ints input. Your input: {1}".format(
-                        argument_name, type(wl)
-                    )
+                    f"""Function argument {argument_name} requires a float/int or list 
+                    of floats/ints input. Your input: {type(wl)}"""
                 )
-        if type(wl) == list:
+        if isinstance(wl, list):
```
nicolejkeeney, line 2052: Cool, I just googled why this is a better method for checking the variable type and learned something from your PR. Thanks!

tests/data_interface/test_data_parameters.py
```diff
@@ -0,0 +1,514 @@
...
+
+def mock_scenario_to_experiment_id(scenario, reverse=False):
+    """Mock implementation of scenario_to_experiment_id with additional scenarios"""
+    scenario_dict = {
+        "Historical Reconstruction": "reanalysis",
+        "Historical Climate": "historical",
+        "SSP 2-4.5": "ssp245",
+        "SSP 5-8.5": "ssp585",
+        "SSP 3-7.0": "ssp370",
+        "SSP 1-2.6": "ssp126",  # Add the missing scenario
```
nicolejkeeney, line 25: Are there talks of adding this scenario into code base? If not, I think we should remove it. Perhaps it is in our data catalog though and I’m not aware? 

I think SSP 1-2.6 (the “sustainability” approach) is pretty unlikely and not commonly used, although maybe I am wrong.

tests/data_interface/test_data_parameters.py
```diff
@@ -0,0 +1,514 @@
...
+                        "Dynamical",
+                        "Dynamical",
+                    ],
+                    "timescale": [
+                        "hourly, daily, monthly",
+                        "daily, monthly",
+                        "hourly",
+                        "hourly",
+                    ],
+                    "unit": ["°C", "mm day⁻¹", "°C", "mm day⁻¹"],
```
nicolejkeeney, line 79: I don’t think these are the right strings for these units, and I don’t think degC is the native unit for Air Temp either. Here’s the [csv](https://github.com/cal-adapt/climakitae/blob/main/climakitae/data/variable_descriptions.csv) with the native units for each variable.

tests/data_interface/test_data_parameters.py
```diff
@@ -0,0 +1,514 @@
...
+                    "extended_description": [
+                        "Temperature",
+                        "Rain",
+                        "Heat Index",
+                        "Heavy rain",
+                    ],
+                    "show": [True, True, True, True],
+                }
+            )
+            mock_instance.variable_descriptions = var_desc
```
nicolejkeeney, line 90: Why not just read in the csv file? Shouldn’t add too much overhead to the test time and then you could guarantee that all the info is correct for each variable.

tests/data_interface/test_data_parameters.py
```diff
@@ -0,0 +1,514 @@
...
+            )
+
+            with patch(
+                "climakitae.core.data_interface._get_variable_options_df"
+            ) as mock_get_vars:
+                # Add the 'variable_id' column to the DataFrame
+                mock_var_df = pd.DataFrame(
+                    {
+                        "display_name": ["Air Temperature at 2m"],
+                        "unit": ["°C"],
```
nicolejkeeney, line 167: Same here with this not being a valid unit option (deg C)

nicolejkeeney (commented):
I'm having a hard time understanding the purpose of the mock data for csv / data tables since those files are pretty small. If we update the shapefiles, for example, and the column names changed for some reason in the shapefiles (i.e. "NAME" became "name", for whatever reason), we would need to go in and update all the mock column names here. Versus if we just read in the shapefile once in the tests, it would just need to be updated in one spot. But I left this as a comment rather than a "request changes" since I'm not familiar with this style of testing and perhaps I'm missing something regarding the workflow here.

tests/data_interface/test_data_parameters.py
neilSchroeder: Ope, thank you. Removed.

tests/data_interface/test_data_parameters.py
neilSchroeder: Agreed, in this instance

climakitae/core/data_interface.py
neilSchroeder: Nice! Pro tip: if you install the `pylint` extension in VSCode it'll catch this for you.

nicolejkeeney (approved):
Looks great

PR #554, opened 2025-05-05
Add clip_to_shapefile function

acordonez:
This PR adds `clip_to_shapefile` to util.utils and adds corresponding unit tests. This function uses rioxarray.clip under-the-hood and should work on any shapefile with a valid CRS. Rioxarray already does a lot of data validation and checking so the only validation I'm doing is making sure the input datasets have a CRS set. I add some helpful hints for common errors, which are often due to issues with the CRS or the grid size.

This change gives AE users the ability to clip AE data to a shapefile of their choosing. 

## How to test 
```
pip install pytest
pytest tests/util/test_utils.py
```
Testers can also grab a shapefile and try out the function in AE. Install this branch first.
Here's an example with the [TIGER/Line county shapefile](https://www.census.gov/cgi-bin/geo/shapefiles/index.php?year=2024&layergroup=ZIP+Code+Tabulation+Areas).
```
from climakitae.util.utils import clip_to_shapefile
from climakitae.core.data_interface import get_data
# 3km is small enough to fit in most counties in CA
data = get_data(
    variable = "Precipitation (total)", 
    downscaling_method = "Statistical", 
    resolution = "3 km", 
    timescale = "monthly", 
    scenario = "Historical Climate"
)
# file from https://www.census.gov/cgi-bin/geo/shapefiles/index.php?year=2024&layergroup=ZIP+Code+Tabulation+Areas
shapefile_path="../../shapefiles/tl_2024_us_zcta520.zip"
# can change "95912" to any valid zip code
feature=("ZCTA5CE20",'95912')
# selecting just one time slice for quick demo
tmp=data.sel({"simulation":'LOCA2_IPSL-CM6A-LR_r10i1p1f1'}).isel({"time":0})
clipped = clip_to_shapefile(tmp,shapefile_path,feature=feature)
clipped.plot()

# Pass all-touched keyword (or any other keyword to rio.clip)
#clipped = clip_to_shapefile(tmp,shapefile_path,feature=feature,all_touched=True)
#clipped.plot()
```

## Type of change
- [ ] None of the above  

## To-Do
- [x] Unit tests
- [x] Documentation
  - [x] Complex code commented

### DIFF
--- a/climakitae/util/utils.py
+++ b/climakitae/util/utils.py
@@ -10,6 +10,7 @@ import pandas as pd
 import pyproj
 import rioxarray as rio
 from shapely.geometry import mapping
+from typing import Any
 import xarray as xr
 from timezonefinder import TimezoneFinder
 
@@ -1190,7 +1191,7 @@ def stack_sims_across_locs(ds):
 def clip_to_shapefile(
     data: xr.Dataset | xr.DataArray,
     shapefile_path: str,
-    feature: tuple[str, str | int | float | bool | list[str | int | float | bool]] = (),
+    feature: tuple[str, Any] = (),
     name: str = "user-defined",
     **kwargs,
 ) -> xr.Dataset | xr.DataArray:
--- a/climakitae/util/utils.py
+++ b/climakitae/util/utils.py
@@ -1115,7 +1115,14 @@ def stack_sims_across_locs(ds):
     ]
     return ds
 
-def clip_to_shapefile(data: xr.Dataset | xr.DataArray, shapefile_path: str, feature: tuple[str,str | int | float | bool | list[str | int | float | bool]]=(), name: str = "custom", **kwargs) -> xr.Dataset | xr.DataArray:
+
+def clip_to_shapefile(
+    data: xr.Dataset | xr.DataArray,
+    shapefile_path: str,
+    feature: tuple[str, str | int | float | bool | list[str | int | float | bool]] = (),
+    name: str = "custom",
+    **kwargs,
+) -> xr.Dataset | xr.DataArray:
     """Use a shapefile to select an area subset of AE data.
 
     Parameters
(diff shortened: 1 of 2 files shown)


climakitae/util/utils.py
```diff
@@ -1183,3 +1185,82 @@ def stack_sims_across_locs(ds):
         for sim_name in ds["simulation"]
     ]
     return ds
+
+
+def clip_to_shapefile(
+    data: xr.Dataset | xr.DataArray,
+    shapefile_path: str,
+    feature: tuple[str, str | int | float | bool | list[str | int | float | bool]] = (),
```
neilSchroeder, line 1193: In this unique circumstance I would list this as `tuple[str, Any]` and `from typing import Any` since it looks like you're not really restricting the second value, but the first value MUST be string

climakitae/util/utils.py
```diff
@@ -1183,3 +1185,82 @@ def stack_sims_across_locs(ds):
...
+    if feature:
+        try:
+            print("Selecting feature", feature)
+            if isinstance(feature[1], list):
+                region = region[region[feature[0]].isin(feature[1])]
+            else:
+                region = region[region[feature[0]] == feature[1]]
+            if len(region) == 0:  # No features found
+                raise Exception
+        except Exception as err:
```
neilSchroeder, line 1242: I would raise an error like `ValueError` here, catch the ValueError, and then catch other exceptions in a more general block. Or customize your exception, `Exception("No features found")` so that you have a clearer idea of where the code breaks if it breaks.

neilSchroeder (commented):
Just a couple of soft requests. Overall this looks great!

acordonez:
@neilSchroeder Super helpful. Thank you!

acordonez:
> Looks good to me!
> 
> Something interesting I noticed: I tested with a shapefile for San Diego county that has 5 features in it. 4 of the 5 features are too small to clip and raise an error; I tried with your `feature` argument and the error was raised. Interestingly, when I feed the entire shapefile to the function, it runs without error. It looks like the `rio.clip` function will only raise a `NoDataInBounds` exception if **all** the features have no data. If even a single feature can be clipped, it doesn't raise an error, even though the others are too small!
> 
> I actually think that is desired behavior for us-- so no changes necessary-- but thought that was kind of interesting and wanted to share. I find it a bit strange that `rio.clip` doesn't raise a warning or anything, but whatever.
> 
> I included a screenshot, some example code, and the shapefile in case you are curious and want to test the behavior!
> shp
> 
> ```python
> shp_path = "san_diego_boundary_datasd.zip"
> 
> # This works! 
> clipped = clip_to_shapefile(tmp,shp_path)
> 
> # This works! 
> clipped = clip_to_shapefile(tmp,shp_path, feature=("objectid",56))
> 
> # This doesn't work 
> clipped = clip_to_shapefile(tmp,shp_path, feature=("objectid",60))
> ```
> 
> [san_diego_boundary_datasd.zip](https://github.com/user-attachments/files/20063556/san_diego_boundary_datasd.zip)

@nicolejkeeney  When you don't set a feature rio uses all the features in the shapefile and clips to their aggregate area. It sounds like the total area of your features was large enough to be clipped in your data.

climakitae/util/utils.py
neilSchroeder: Yeah this looks great. Thank you :slightly_smiling_face:

neilSchroeder (approved):

PR #591, opened 2025-06-25
Remove selections from convert_to_local_time

acordonez:
[What's changed in this PR?]

[Why did you change this and what applicable context is needed to understand why this change is needed?]

## How to test 
[How should reviewers test the changes?] 

## Type of change
- [ ] None of the above  

## To-Do
- [ ] Unit tests
- [ ] Documentation
  - [ ] Complex code commented
- [ ] Black formatting has been utilized

### DIFF
--- a/climakitae/util/utils.py
+++ b/climakitae/util/utils.py
@@ -785,7 +785,6 @@ def convert_to_local_time(data: xr.DataArray) -> xr.DataArray:
 
     Args:
         data (xarray.DataArray): Input data.
-        local_time_slice (tuple[int, int]): Start and end years of returned data.
 
     Returns:
         xarray.DataArray: Data with converted time coordinate.
--- a/climakitae/util/utils.py
+++ b/climakitae/util/utils.py
@@ -779,9 +779,7 @@ def match_attr(data: xr.DataArray, key, value):
     return False
 
 
-def convert_to_local_time(
-    data: xr.DataArray, local_time_slice: tuple[int, int]
-) -> xr.DataArray:
+def convert_to_local_time(data: xr.DataArray) -> xr.DataArray:
     """
     Convert time dimension from UTC to local time for the grid or station.
 
@@ -799,14 +797,6 @@ def convert_to_local_time(
             "You've selected a timescale that doesn't require any timezone shifting, due to its timescale not being granular enough (hourly). Please pass in more granular level data if you want to adjust its local timezone."
         )
         return data
-
-    # 1. Get the time slice from selections
-    start, end = local_time_slice
-
-    if data.time.dt.year[-1].item() == end:
-        print(
-            "Warning: Requested end year identical to last year in data. Final day may be incomplete after conversion to local time."
-        )
 
     # Default lat/lon values in case other methods fail
     lat = None
@@ -861,15 +851,12 @@ def convert_to_local_time(
     )
     data["time"] = new_time
 
-    # Subset the data by the requested time slice
-    sliced_data = data.sel(time=slice(f"{start}-01-01", f"{end}-12-31"))
-
     print(f"Data converted to {local_tz} timezone.")
 
     # Add timezone attribute to data
-    sliced_data = sliced_data.assign_attrs({"timezone": local_tz})
+    data = data.assign_attrs({"timezone": local_tz})
 
-    return sliced_data
+    return data
 
 
 def add_dummy_time_to_wl(wl_da: xr.DataArray) -> xr.DataArray:
--- a/climakitae/util/utils.py
+++ b/climakitae/util/utils.py
@@ -772,55 +772,79 @@ def summary_table(data: xr.Dataset) -> pd.DataFrame:
     return df
 
 
-def match_attr(data: xr.DataArray, key, value):
-    if key in data.attrs:
-        if data.attrs[key] == value:
-            return True
-    return False
-
-
-def convert_to_local_time(data: xr.DataArray) -> xr.DataArray:
+def convert_to_local_time(data: xr.DataArray | xr.Dataset) -> xr.DataArray:
     """
     Convert time dimension from UTC to local time for the grid or station.
 
     Args:
-        data (xarray.DataArray): Input data.
+        data (xarray.DataArray | xr.Dataset): Input data.
 
     Returns:
         xarray.DataArray: Data with converted time coordinate.
     """
 
+    # Only converting hourly data
+    if "frequency" not in data.attrs:
+        # Make a guess at frequency
+        timestep = pd.Timedelta(
+            data.time[1].item() - data.time[0].item()
+        ).total_seconds()
+        match timestep:
+            case 3600:
+                frequency = "hourly"
+            case 86400:
+                frequency = "daily"
+            case _ if timestep > 86400:
+                frequency = "monthly"
+    else:
+        frequency = str(data.attrs["frequency"])
+
     # If timescale is not hourly, no need to convert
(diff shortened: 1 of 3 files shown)


claalmve:
I like these changes! The inclusion of calling this function within new core will definitely reduce code complexity for this function. Match cases look good.

climakitae/util/utils.py
```diff
@@ -772,90 +772,102 @@ def summary_table(data: xr.Dataset) -> pd.DataFrame:
...
+    if frequency != "hourly":
         print(
             "You've selected a timescale that doesn't require any timezone shifting, due to its timescale not being granular enough (hourly). Please pass in more granular level data if you want to adjust its local timezone."
         )
         return data
 
-    # 1. Get the time slice from selections
-    start, end = selections.time_slice
+    # Find out if Stations or Gridded type
+    if "data_type" not in data.attrs:
```
neilSchroeder, line 810:
```suggestion
    if not (data_type := data.attrs.get("data_type", None)):
```
dictionaries support `.get(key, default)` method that makes these kinds of assignments very nice to read and use. And the walrus operator is handy for checking and setting in the same line, so you can remove the try, except block and the `data_type = ...` line below.

climakitae/util/utils.py
```diff
@@ -772,90 +772,102 @@ def summary_table(data: xr.Dataset) -> pd.DataFrame:
...
+            try:
+                # Grab from one of data arrays in dataset
+                variable = list(data.keys())[0]
+                data_type = data[variable].attrs["data_type"]
+            except KeyError:
+                print(
+                    f"Could not find attribute 'data_type' attribute set in {variable} attributes. Please set `data_type` attribute."
+                )
+                return data
+    else:
```
neilSchroeder, line 826: if you use the walrus operator this "else" chain should also be able to be removed

climakitae/util/utils.py
```diff
@@ -772,90 +772,102 @@ def summary_table(data: xr.Dataset) -> pd.DataFrame:
...
-        lat = center_pt.y
-        lon = center_pt.x
+    match data_type:
+        case "Stations":
+            # Read stations database
+            stations_df = read_csv_file(stations_csv_path)
+            stations_df = stations_df.drop(columns=["Unnamed: 0"])
+
+            # Filter by selected station(s) - assume first station if multiple
+            match type(data):
```
neilSchroeder, line 841: typical use of match statements for type checking look like:
```python
match data:
    case xr.DataArray():
        # handle data array case
    case xr.Dataset():
        # handle dataset case
    case _:
        # raise an error because this is the wrong type
```

climakitae/util/utils.py
```diff
@@ -867,60 +879,28 @@ def convert_to_local_time(data: xr.DataArray, selections) -> xr.DataArray:
...
 
     print(f"Data converted to {local_tz} timezone.")
 
     # Add timezone attribute to data
-    sliced_data = sliced_data.assign_attrs({"timezone": local_tz})
-
-    # Reset selections object to what it was originally (if we changed it)
-    if end < 2100:
-        selections.time_slice = (start, end)
+    match type(data):
```
neilSchroeder, line 895: formatting

climakitae/util/utils.py
```diff
@@ -772,90 +772,102 @@ def summary_table(data: xr.Dataset) -> pd.DataFrame:
...
+            case _ if timestep > 86400:
+                frequency = "monthly"
+    else:
+        frequency = str(data.attrs["frequency"])
+
     # If timescale is not hourly, no need to convert
-    if selections.timescale in ["monthly", "daily"]:
+    if frequency != "hourly":
         print(
             "You've selected a timescale that doesn't require any timezone shifting, due to its timescale not being granular enough (hourly). Please pass in more granular level data if you want to adjust its local timezone."
```
neilSchroeder, line 805: I'd like to change this message to something like:
```python
"This dataset's timescale is not granular enough to covert to local time. Local timezone conversion requires hourly data."
```

neilSchroeder (commented):
This looks great! Comments are minor and centered around formatting and structure, and one comment about changing an existing error message.

climakitae/util/utils.py
```diff
@@ -772,90 +772,105 @@ def summary_table(data: xr.Dataset) -> pd.DataFrame:
...
-        data (xarray.DataArray): Input data.
-        selections: DataParameters object containing selection details.
+        data (xarray.DataArray | xr.Dataset): Input data.
 
     Returns:
         xarray.DataArray: Data with converted time coordinate.
     """
 
+    # Only converting hourly data
+    if "frequency" not in data.attrs:
```
neilSchroeder, line 787: This is another good place for a walrus operator:
```suggestion
    if not (frequency := data.attrs.get("frequency", None)):
```
Apologies for not catching this earlier

climakitae/util/utils.py
```diff
@@ -772,90 +772,105 @@ def summary_table(data: xr.Dataset) -> pd.DataFrame:
...
+            data.time[1].item() - data.time[0].item()
+        ).total_seconds()
+        match timestep:
+            case 3600:
+                frequency = "hourly"
+            case 86400:
+                frequency = "daily"
+            case _ if timestep > 86400:
+                frequency = "monthly"
+    else:
```
neilSchroeder, line 799: Which means you can also remove this `else` statement.

climakitae/util/utils.py
neilSchroeder: You can cast this as a string if you need to, but I'd be surprised if its necessary.

neilSchroeder: But be careful about casting as a string inside the `not (...)` statement because strings are "truthy" and the point is to get it to resolve to `None` and not `"None"`. Unfortunately python is weird in that `bool(None)` is `False` but `bool(str(None))` is `True`

neilSchroeder (commented):
Apologies, but I missed another spot for a `.get()` and a walrus operator

acordonez:
Seeing in TMY that this needs some option to pass in the mean lat lon, as area averaged data does not include coordinates

acordonez:
Sorry @neilSchroeder, had to add more code (option to pass lat/lon for area averaged data case)

neilSchroeder (approved):

climakitae/util/utils.py
```diff
@@ -772,90 +772,118 @@ def summary_table(data: xr.Dataset) -> pd.DataFrame:
     return df
 
 
-def convert_to_local_time(data: xr.DataArray, selections) -> xr.DataArray:
+def convert_to_local_time(
+    data: xr.DataArray | xr.Dataset,
+    grid_lon: float | None = None,
```
neilSchroeder, line 777: Please import `UNSET` from `constants` and use it as the default value. `None` should be reserved for situations where code attempted to execute and failed for some reason.

neilSchroeder: You'll also need to change any checks for `is None` to `is UNSET`

climakitae/util/utils.py
```diff
@@ -772,90 +772,118 @@ def summary_table(data: xr.Dataset) -> pd.DataFrame:
...
+                data_type = data[variable].attrs["data_type"]
+            except KeyError:
+                print(
+                    f"Could not find attribute 'data_type' attribute set in {variable} attributes. Please set `data_type` attribute."
+                )
+                return data
 
     # Default lat/lon values in case other methods fail
     lat = None
     lon = None
```
neilSchroeder, line 833: these are already defined, no need to re-define

neilSchroeder (changes requested):
No worries, just a small change for default value handling.

PR #610, opened 2025-07-31
Fix unit conversion bug for derived variables

acordonez:
Fixing a bug where the dew point temperature derived data would not successfully have units converted to selected units. This bug may potentially affected other derived variables. 

I've rearranged L1335-1346 so that the unit conversion happens before copying the data attributes from the selections object. This is because the data array units would be overwritten by the units in the selections object, which affected the subsequent unit conversion.

Need to be able to request dew point temperature in units other than K.

## How to test 
Try loading dew point temperature or other derived variables in different units:
```
from climakitae.core.data_interface import get_data
dewpt_data = get_data(
    variable="Dew point temperature",
    resolution="9 km",
    timescale="hourly",
    data_type="Gridded",
    units="degC",
    latitude=(33.93816 - 0.05, 33.93816 + 0.05),
    longitude=(-118.3866 - 0.06, -118.3866 + 0.06),
    area_average="Yes",
    scenario=["Historical Climate", "SSP 3-7.0"],
    time_slice=(
        1990,
        1991,
    )
)
```

## Type of change
- [ ] None of the above  

## To-Do
- [x] Unit tests
  - [x] Advanced Testing passing (select Advanced Testing label)
- [x] Documentation
  - [x] Complex code commented

### DIFF
--- a/climakitae/core/data_load.py
+++ b/climakitae/core/data_load.py
@@ -1332,13 +1332,13 @@ def read_catalog_from_select(selections: "DataParameters") -> xr.DataArray:
                 )
 
         # ------ Set attributes ------
-        # Reset selections to user's original selections
-        selections.variable_id = [orig_var_id_selection]
-        selections.units = orig_unit_selection
-
         # Convert units before copying data attributes
         da = convert_units(da, selected_units=orig_unit_selection)
         da.name = orig_variable_selection  # Set name of DataArray
+
+        # Reset selections to user's original selections
+        selections.variable_id = [orig_var_id_selection]
+        selections.units = orig_unit_selection
 
         # Some of the derived variables may be constructed from data that comes from the same institution
         # The dev team hasn't looked into this yet -- opportunity for future improvement
--- a/climakitae/core/data_load.py
+++ b/climakitae/core/data_load.py
@@ -1332,18 +1332,18 @@ def read_catalog_from_select(selections: "DataParameters") -> xr.DataArray:
                 )
 
         # ------ Set attributes ------
+        # Reset selections to user's original selections
+        selections.variable_id = [orig_var_id_selection]
+        selections.units = orig_unit_selection
+
+        # Convert units before copying data attributes
+        da = convert_units(da, selected_units=orig_unit_selection)
+        da.name = orig_variable_selection  # Set name of DataArray
+
         # Some of the derived variables may be constructed from data that comes from the same institution
         # The dev team hasn't looked into this yet -- opportunity for future improvement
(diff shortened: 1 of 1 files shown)


acordonez:
> Something is wrong with the unit conversion. Using the code you provided, I get the same value for both degF and degC. Also, the value retrieved from this code:
> 
> ```python
> dewpt_data = get_data(
>     variable="Dew point temperature",
>     resolution="9 km",
>     timescale="hourly",
>     data_type="Gridded",
>     units="degC",
>     latitude=(33.93816 - 0.05, 33.93816 + 0.05),
>     longitude=(-118.3866 - 0.06, -118.3866 + 0.06),
>     area_average="Yes",
>     scenario=["Historical Climate", "SSP 3-7.0"],
>     time_slice=(
>         1990,
>         1991,
>     )
> )
> dewpt_data.isel(time=0).mean().values
> ```
> 
> Is 269.091 (for both degC and degF), which makes me think it is not actually converting from Kelvin.

@nicolejkeeney I haven't been able to reproduce this (getting -4.0589905 C with the above code after pulling these changes into a clean branch), but the issue you're seeing is the one I'm trying to fix with this PR. Is it possible it didn't install correctly?

nicolejkeeney (commented):
See note above regarding my (personal, unrelated to this PR) issues with climakitae + branches.

climakitae/core/data_load.py
```diff
@@ -1332,19 +1332,19 @@ def read_catalog_from_select(selections: "DataParameters") -> xr.DataArray:
...
+        # Convert units before copying data attributes
         da = convert_units(da, selected_units=orig_unit_selection)
         da.name = orig_variable_selection  # Set name of DataArray
 
         # Reset selections to user's original selections
         selections.variable_id = [orig_var_id_selection]
         selections.units = orig_unit_selection
 
+        # Some of the derived variables may be constructed from data that comes from the same institution
+        # The dev team hasn't looked into this yet -- opportunity for future improvement
```
neilSchroeder, line 1344: Is this something we need to look into?

neilSchroeder (approved):
Tested on the hub and it works fine for me.

climakitae/core/data_load.py
acordonez: Probably low priority but it could be worth checking which derived variables are getting what attributes and if there are any that need updating at this point.

neilSchroeder: Seems like a non-issue with `new-core` coming down the pipeline

vicford (approved):
Tested both in `get_data` and in TMY, TY!

PR #647, opened 2025-09-26
Fix/cava data file naming

neilSchroeder:
Allow file naming for cava_data, as well as improve automatic file naming.

request from SCE but also just good practice

## How to test 
Check that cava_data accepts a new file name, or if you pass a single lat/lon it uses 6 significant figures, instead of 3. 

## Type of change
- [ ] New tests (increased test coverage on existing functionality)
- [ ] None of the above  

## To-Do
- [x] Unit tests
- [x] Documentation
  - [ ] Complex code commented

### DIFF
--- a/climakitae/explore/vulnerability.py
+++ b/climakitae/explore/vulnerability.py
@@ -1129,16 +1129,28 @@ def cava_data(
             for _, data_point in enumerate(loaded_data):
                 lat_str = str(round(data_point.lat.item(), 6)).replace(".", "-")
                 lon_str = str(abs(round(data_point.lon.item(), 6))).replace(".", "-")
-                filename_str = f"{clean_params['raw_name']}_" f"{lat_str}N_{lon_str}W"
+                loc_suffix = f"{lat_str}N_{lon_str}W"
+                if file_name is None:
+                    filename_str = f"{clean_params['raw_name']}_{loc_suffix}"
+                else:
+                    # Append location info and _raw suffix to custom filename
+                    filename_str = f"{file_name}_raw_{loc_suffix}"
                 _export_no_e(
                     data_point,
-                    filename=filename_str if file_name is None else file_name,
+                    filename=filename_str,
                     file_format=file_format,
                 )
         else:
+            if file_name is None:
+                raw_filename = "combined_raw_data"
+            else:
+                # Add _raw suffix when export_method="both" to avoid name collision
+                raw_filename = (
+                    f"{file_name}_raw" if export_method == "both" else file_name
+                )
             _export_no_e(
                 loaded_data,
-                filename="combined_raw_data" if file_name is None else file_name,
+                filename=raw_filename,
                 file_format=file_format,
             )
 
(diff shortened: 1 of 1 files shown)


neilSchroeder:
The tests here are broken because the tests in main are broken

climakitae/explore/vulnerability.py
```diff
@@ -492,6 +492,8 @@ class CavaParams(param.Parameterized):
     event_duration : tuple, default=(1, "day")
         Duration and unit for event analysis. Format: (duration, unit).
         Unit options: "hour", "day".
+    file_name: str
+        Base name for output files.
```
nicolejkeeney, line 496: This should specify that it means a filename **without** a file extension, I think this isn’t immediately obvious.

climakitae/explore/vulnerability.py
```diff
@@ -604,6 +606,7 @@ class CavaParams(param.Parameterized):
         doc="Distribution function",
     )
     event_duration = param.Tuple(default=(1, "day"))
+    file_name = param.String(default="cava_output", doc="Base name for output files")
```
nicolejkeeney, line 609: Same here

climakitae/explore/vulnerability.py
```diff
@@ -1181,16 +1187,22 @@ def cava_data(
...
-                    lat_str = str(round(data_point.lat.item(), 3)).replace(".", "")
-                    lon_str = str(round(data_point.lon.item(), 3)).replace(".", "")
+                    lat_str = str(round(data_point.lat.item(), 6)).replace(".", "")
+                    lon_str = str(round(data_point.lon.item(), 6)).replace(".", "")
                     _export_no_e(
                         data_point,
-                        filename=f"{clean_params['calc_name']}_{lat_str}N_{lon_str}W",
+                        filename=(
+                            f"{clean_params['calc_name']}_{lat_str}N_{lon_str}W"
+                            if file_name is None
```
nicolejkeeney, line 1196: Longitude in the US is negative. So this will be “_-117.161087W” which is fine, but just wanted to mention since it's a bit odd with the underscore followed by a hyphen. I don’t have a better suggestion, just pointing this out.

claalmve:
Also seeing this error when I don't pass anything into `file_name`:

`ValueError: String parameter 'CavaParams.file_name' only takes a string value, not value of <class 'NoneType'>.`

climakitae/explore/vulnerability.py
```diff
@@ -1117,18 +1127,32 @@ def cava_data(
...
+            for _, data_point in enumerate(loaded_data):
+                lat_str = str(round(data_point.lat.item(), 6)).replace(".", "-")
+                lon_str = str(abs(round(data_point.lon.item(), 6))).replace(".", "-")
+                loc_suffix = f"{lat_str}N_{lon_str}W"
+                if file_name is None:
+                    filename_str = f"{clean_params['raw_name']}_{loc_suffix}"
+                else:
+                    # Append location info and _raw suffix to custom filename
+                    filename_str = f"{file_name}_raw_{loc_suffix}"
                 _export_no_e(
```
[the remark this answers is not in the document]
claalmve: Oh I see I have been leaving `batch_mode=True`, which makes `separate_files=False`. I should have remembered that since I wrote that logic. But perhaps this logic doesn't make sense anymore? `batch_mode=True` and `separate_files=True` makes sense to have too.

claalmve:
Everything is working properly for me now! I think some more user guidance that tells the user that `batch_mode = True` sets `separate_files = False` would be nice (it's in the table, but a printed statement in the function call would also make things clearer).

PR #710, opened 2026-01-22
Time-based approach option added to Standard Year climate profile generation

vdmachuca:
### **Updated overview for PR re-review**

Users requested the option to generate a time-based Standard Year profile. We are responding to this request in a way that still delivers a warming level profile, in an effort to promote warming levels as the new standard. Here's what this looks like:

1. The user inputs "approach"="Time" and "centered_year"= a year between **1980 and 2099**. They now have the option to input "time_profile_scenario". Valid options are 'SSP 2-4.5', 'SSP 3-7.0', and 'SSP 5-8.5', with 'SSP 2-4.5' and 'SSP 5-8.5' only being valid options for resolutions 9km and 45km. If no "time_profile_scenario" is provided, the default is 'SSP 3-7.0'.

2. retrieve_profile_data() sends input arguments to the helper function _handle_approach_params()

3. _handle_approach_params() validates the inputs and then uses get_gwl_at_year() to get the corresponding GWL for that year and updates the arguments dictionary so "approach" = "Warming Level" and "warming_level" = the new gwl. It catches incorrect "resolution", "centered_year", "approach", and "time_profile_scenario" combinations. 
--> **For 'centered_year' < 1950**:  the corresponding "Historical" warming level is retrieved. 

5. retrieve_profile_data() then retrieves data using the new gwl corresponding to the user's input centered year.

6. retrieve_profile_data() then uses _filter_by_ssp() to select data for the model specified by "time_profile_scenario" or the default "SSP 3-7.0". Note that for SSP 2-4.5 and 5-8.5, ONLY ONE model is thus selected in this step.

8. export_profile_to_csv() outputs a CSV with a filename that includes the approach, centered year, and scenario (approach is only included if it is set to "Time").

Errors are thrown for all time-based input combinations that are not exactly so:
- "approach" = "Time"
- "centered_year" = year between 1980-2099
- no "warming_level" input

ALSO if resolution is 3km for 'time_profile_scenario'='SSP 2-4.5' or 'SSP 5-8.5'

**Ancillary update:**
I also modified _create_simple_dataframe() to output a dataframe with a multi-index and updated the associated unit tests. This came up because 'time_profile_scenario'='SSP 2-4.5' or 'SSP 5-8.5' results in a single gwl, single simulation result, which that helper function handles.

**Unit Tests**
I've done the following:
- moved retrieve_profile_data() unit tests to test_standard_year_data_retrieval.py
- updated retrieve_profile_data() unit tests
- added unit tests for _handle_approach_params()
- updated the test_get_clean_standardyr_filename() unit test
- added unit tests for _filter_by_ssp()
- updated unit tests for _create_simple_dataframe (in class TestCreateSimpleDataframe)
- updated test_construct_profile_dataframe_single_wl_single_sim()

### Old overview, for the record
Users requested the option to generate a time-based Standard Year profile. We are responding to this request in a way that still delivers a warming level profile, in an effort to promote warming levels as the new standard. Here's what this looks like:

1. user inputs "approach"="Time" and "centered_year"= a year between 2015 and 2099 *** changed to 1980-2099

2. retrieve_profile_data() hands this "centered_year" input to the helper function _handle_approach_params()

3. _handle_approach_params() validates the inputs and then uses get_gwl_at_year() to get the corresponding GWL for that year, and updates the arguments dictionary so "approach" = "Warming Level" and "warming_level" = the new gwl

5. retrieve_profile_data() then proceeds as it usually would with the default warming level approach, using the new gwl corresponding to the user's input centered year

6. export_profile_to_csv() outputs a CSV with a filename that includes the approach and centered year (approach is only included if it is set to "Time")

Errors are thrown for all time-based input combinations that are not exactly so:
- "approach" = "Time"
- "centered_year" = year between 2015-2099 *** changed to 1980-2099
- no "warming_level" input

This is meant to provide guardrails around time-based profile generation and promote user understanding of this specific approach.

Users want a time-based approach. Our response gives them the ability to generate a profile for an input year, while still deliveraing a profile that uses the warming levels approach.

## How to test 
Test valid and invalid profile selections inputs using [this notebook](https://github.com/vdmachuca/skillshare_demo/blob/changes/test_new_sy_functionality.ipynb). You should only get a profile if "approach"="Time" and "centered_year"= a year between 1980 and 2099, with no "warming_level" input. Anything outside of this should throw a value error.
You should also get errors if resolution is 3 km for 'time_profile_scenario'='SSP 2-4.5' or 'SSP 5-8.5'.

In standard_year_profile.py, review
- retrieve_profile_data()
- _handle_approach_params()
- _filter_by_ssp()
- export_profile_to_csv()
- _get_clean_standardyr_filename()
- _create_simple_dataframe()

prioritizing the first three functions.

I added a bunch of print statements to make the process clear to the user, so would appreiciate feedback on that also.

## Type of change
- [ ] New tests (increased test coverage on existing functionality)
- [ ] None of the above  

## To-Do
- [x] Unit tests
- [x] Documentation
  - [x] Complex code commented

### DIFF
--- a/climakitae/explore/standard_year_profile.py
+++ b/climakitae/explore/standard_year_profile.py
@@ -703,7 +703,9 @@ def retrieve_profile_data(**kwargs: Any) -> Tuple[xr.DataArray, xr.DataArray]:
         if "latitude" in kwargs or "longitude" in kwargs:
             kwargs.pop("latitude", None)
             kwargs.pop("longitude", None)
-            print("   ⚠️  Note: Using cached_area, ignoring provided latitude/longitude")
+            print(
+                "   ⚠️  Note: Using cached_area, ignoring provided latitude/longitude"
+            )
         if "stations" in kwargs:
             kwargs.pop("stations", None)
             print("   ⚠️  Note: Using cached_area, ignoring provided stations")


climakitae/explore/standard_year_profile.py
```diff
@@ -150,7 +160,13 @@ def _get_clean_standardyr_filename(
     clean_loc_name = location.replace(" ", "_").replace("(", "").replace(")", "")
     clean_q_name = f"{q:.2f}".split(".")[1].lower()
     clean_var_name = var_id.lower()
-    clean_gwl_name = match_str_to_wl(gwl).lower().replace(".", "pt")
+
+    if gwl is None:
+        clean_gwl_name = ""
+    else:
+        clean_gwl_name = match_str_to_wl(gwl).lower().replace(".", "pt")
+        clean_gwl_name = f"_{clean_gwl_name}"
```
neilSchroeder: You can write this a little cleaner as:
```python
clean_gwl_name = ""
if gwl:
    clean_gwl_name = match_str_to_wl(gwl).lower().replace(".", "pt")
    clean_gwl_name = f"_{clean_gwl_name}"
```

Which goes for all of these statements.

climakitae/explore/standard_year_profile.py
```diff
@@ -335,15 +370,33 @@ def export_profile_to_csv(profile, **kwargs):
     # Check profile MultiIndex to pull out data by Global Warming Level
     match profile.keys().nlevels:
         case 2:  # Single WL
-            gwl = global_warming_levels[0]
+            # If time-based approach being used, do not include gwl in the filename
+            if global_warming_levels is None:
+                gwl = global_warming_levels
+            else:
+                gwl = global_warming_levels[0]
```
neilSchroeder: Same type of if/else that can be re-written. The "pythonic" way to do this is to instantiate the default, then add an `if` clause for modifying the behavior. See the comment above.

climakitae/explore/standard_year_profile.py
```diff
@@ -352,7 +405,95 @@ def export_profile_to_csv(profile, **kwargs):
...
+            # get warming level based on year
+            # and set 'warming_level' to this value
+            else:
+                print(
+                    f"You have chosen to produce a time-based Standard Year climate profile centered around {centered_year}. \n"
+                    "Standard year functionality for time-based profiles identifies the closest warming level at that centered year.\n"
+                    f"The corresponding global warming level for input centered year {centered_year} will now \n"
+                    "be determined and used to produce the profile."
+                )
+                gwl_options = get_gwl_at_year(centered_year, "SSP 3-7.0")
```
wkrantz: In my opinion the scenario should be passed into profile_selections as a parameter. It's fine for SSP 3-7.0 to be the default, but it's important conceptually that when asking for a time-based product, the scenario has to be specified.

climakitae/explore/standard_year_profile.py
```diff
@@ -546,8 +692,10 @@ def retrieve_profile_data(**kwargs: any) -> Tuple[xr.Dataset, xr.Dataset]:
             ),
         ),
         "approach": "Warming Level",
-        "warming_level": [1.2],
-        "warming_level_window": kwargs.get("warming_level_window", None),
+        "warming_level": [1.2],  # Historic global warming level
```
wkrantz: Having the default historical level be 1.2 generally makes sense, but using our methods it corresponds to the year 2020. This creates the awkward situation where the user can request "future" data from any year starting in 2015, but the "historical" data is from 2020. 

Options to resolve this:
1) Reduce the allowed range of "future" years to start with 2021 (easiest workaround for now)
2) Allow users to specify their reference period, and also check that the reference period is less than the requested future period. Allowing earlier reference periods means sometimes data from one or more models will not be available, which takes some extra handling.

wkrantz (changes requested):
Related to these suggestions, but not as critical and something we can address at a later date: it would be nice to be able to generate a profile (not a delta) for a historical date/gwl. This would technically work just fine by specifying a historical year, and using the "Historical" scenario in get_gwl_from_year, but it is incompatible with the way input years are currently filtered to within 2015-2099.

climakitae/explore/standard_year_profile.py
wkrantz: This will also need updating if you're allowing center_year inputs before 2015, as "get_gwl_from_year" will only return data from "Historical" before that year, and asking for SSP3-7.0 returns an error.

vdmachuca:
I'm still getting linting errors (after double-checking my save setting on VSCode and re-saving everything) and am not sure how to resolve it!

vicford:
> I feel like calling this new approach “Time” is misleading since it is using a warming levels approach. I don’t have a problem with the method as it’s applied, but the naming is misleading IMO. Especially since we use an approach called “time” throughout `climakitae` that is not this approach.
> 
> In order to reflect what’s happening under the hood better, the approach could be called “Year-Referenced WL” or something like that.
> 
> @vicford any thoughts?

This was an intentional decision and your comment is a fair criticism. I am hesitant to add a new approach (at present) called "Year-Referenced WL" (great clear name, fyi) because that has larger ramifications throughout climakitae. I recommend lightly tagging this idea in the GWL backlog, especially if more "GWL to time and back" functionality is continued to be developed. This is highly relevant to GWL agnostic tools as well.

climakitae/explore/standard_year_profile.py
```diff
@@ -564,10 +814,25 @@ def retrieve_profile_data(**kwargs: any) -> Tuple[xr.Dataset, xr.Dataset]:
     get_data_params.update(kwargs)
     future_data = get_data(**get_data_params)
 
+    # Filter models by input scenario, if time-based approach specified
```
vicford: After talking with @wkrantz This step is actually not needed! We don't need to filter for SSP at steps 3 and 4. This is on me not comprehending some tricky GWL steps. Please remove this block!

vicford (approved):
Minor changes @vdmachuca -- once these are in, no re-review is necessary and you can merge.

PR #754, opened 2026-04-17
Update threshold_tools documentation and fix block_size bug

acordonez:

- Add notes to the return variable functions to be clearer about what timescales the probabilities are for
- Fixing the name of the block_size attribute so it can be correctly used in return variable calculations.
- Replace `.apply()` calls with `.map()` per deprecation message.

## Link to corresponding Jira ticket(s)
Tangentially related to [AE-1400](https://eaglerockanalytics.atlassian.net/browse/AE-1400)

## Testing
- [x] Unit tests written for new/modified code (goal: 80% coverage)
  - All public functions have unit tests
  - Functions that must produce specific values have unit tests
- [x] Verified that notebooks utilizing affected functions still work
- [x] Appropriate manual testing completed
- [ ] Advanced Testing label added to PR if this PR makes major changes to the codebase 

## How to Test
Coverage for threshold_tools.py is 83%
`python -m coverage run -m pytest tests/threshold_tools/`
`python -m coverage html --include=threshold_tools.py`

The threshold_event_types notebook in cae_notebooks can be used to check this code interactively.

## Documentation
- [x] Complex code includes comments explaining logic
- [x] Function documentation created (docstrings)

## Code Quality
- [x] Follows PEP8 naming and style conventions
- [x] Helper functions prefixed with underscore `_`
- [x] Linting completed and all issues resolved
- [x] Does not replicate existing functionality
- [x] Aligns with general coding standards of existing codebase
- [x] Code generalized for multiple uses (unless too complex/time-intensive)

## Review Process
- [x] PR review instructions provided:
  - [x] Type of review requested (scientific/technical/debugging)
  - [ ] Level of review detail needed

## Administrative Reminders
  - Jira ticket moved to "Review" when PR created
  - Jira ticket will be moved to "Done" when complete
  - PR branch will be deleted after merge


acordonez:
@vicford for you review I want to verify that this is keeping the functionality you expect and that the new comments make sense.

climakitae/explore/threshold_tools.py
```diff
@@ -866,13 +875,10 @@ def _get_return_variable(
...
-    # get block_size from the block maxima series attributes, if available. otherwise assume block size=1 year
-    if hasattr(bms, "block size"):
-        block_size = int(
-            bms.attrs["block size"][0:-5]
-        )  # expected string format from get_block_maxima: '2 year'; extract the integer value here
-    else:
-        block_size = 1
+    # Set the block size to 1. The return variable results will be relative to
+    # whatever the block size is in the bms; e.g., if the bms block size is 2,
+    # the 1-in-X return period is for X number of 2-year blocks.
```
vicford: @acordonez This comment to me reads as if it contradicts what we are doing here and the note at lines 607-612. I was assuming that the block size is always 1 no matter what with this change, however this comment reads like block size can be another value, like 2?

climakitae/explore/threshold_tools.py
```diff
@@ -618,6 +624,9 @@ def _calculate_return(
             return_value = fitted_distr.ppf(return_event)
             result = np.round(return_value, 5)
         else:
+            # Use compound probability to get 1 year probability
+            # total probability = 1 - (1 - 1/X)**M
+            # For example see https://journals.ametsoc.org/view/journals/atot/37/11/JTECH-D-20-0070.1.xml
```
vicford: Gorgeous, exactly the kind of notation we want to provide

climakitae/explore/threshold_tools.py
```diff
@@ -604,6 +604,12 @@ def _calculate_return(
     np.ndarray
         Computed extreme value metric.
 
+    Notes
+    -----
+    While this function has the ability to convert probabilities for large block sizes
+    to annual probabilities, the _get_return_variable function passes a block size of 1
+    when calling this no matter the block size used in the bms.
+
```
vicford, line 616: I'm finding it hard to articulate why I'm struggling with this comment. Perhaps it is because the thresholds tools functions are modularized (so it's not obvious the connection between this function and _get_return_variable) and there's not an immediate call to action / next step. 

To me, this reads as "no matter what size block you used in function A, you're going to get your answer from function B in 1-year blocks", which could be confusing as to why the option to do larger block sizes is even a thing. Is this our desired approach?

climakitae/explore/threshold_tools.py
vicford, line 603: This docstring reinforces the "can we do different block sizes?" comment, since it says "by default 1" whereas below we state that it always gets reset to 1.

climakitae/explore/threshold_tools.py
```diff
@@ -1015,6 +1021,12 @@ def get_return_value(
     xarray.Dataset
         Dataset with return values and confidence intervals
 
+    Notes
+    -----
+    The input return period values must be in units of the BMS block size. For example, for
+    annual blocks the units are years. For 2-year blocks the units are 2-years. To get the
+    1-in-X annual return value using 2-year blocks, enter X/2 as the desired return period.
```
vicford: This comment is a bit confusing because of the last sentence: Are we trying to say that if you have a block size that is not 1 year but you want an annual return value, please input as the `return_period` as "annual return level" divided by block size? Aka, if I had a block size of 5, but I wanted a 1-in-15 annual return value, I'd do `return_period = 3`? 

```
The input return period values must be in units of the BMS block size. 
For example, a 1-year block size provides return values at the annual scale. 
A 2-year block size provides return values for each 2-year block. 
If you are trying to calculate a 1-in-X annual return value with a non-default block size 
(i.e., block size != 1), you must pass the annual return level divided by your input block size. 
Example: Input DataArray bms is calculated using a block size of 5-years to be used in a 1-in-15 annual return value calculation. return_period argument must be set to 3 (X / block size)
```

vicford (commented):
General comprehension questions on my end @acordonez.

climakitae/explore/threshold_tools.py
acordonez: @vicford The way this code currently works in practice, as much as I've tested, is that you can use any block size for getting the BMS. However, the code that gets your return variable (return value, return period, or return probability) was always using a block size of 1.

acordonez: For example:
```
ams= get_block_maxima(
    t2max["t2max"].compute(),
    block_size=1 # This could be any positive integer
)

# but when you call the get_return_period function, a block size of 1 is always used due to bug
ref_return_period = get_return_period(ams, return_value=110, multiple_points=False)
```

climakitae/explore/threshold_tools.py
acordonez: It doesn't look like this was originally the desired approach, since the code exists to change the block size. We do have the option of having _get_return_variable pass the block size that was used to generate the BMS (and return annual return values/periods), just need to communicate to end users that there's been a shift in how the code works.

climakitae/explore/threshold_tools.py
vicford: @acordonez This makes sense relative to my comment below on the following code block. As in, if you want an annual return value, you're going to have to divide by your block size for it to appropriately return relative to your block size. If I'm understanding that right, if a user doesn't divide by block size, they'll get an answer but it may likely not be correct. Perhaps in the future we have some sort of test or something that identifies if it is 1-year blocks or not?

> The input return period values must be in units of the BMS block size. For example, for
>     annual blocks the units are years. For 2-year blocks the units are 2-years. To get the
>     1-in-X annual return value using 2-year blocks, enter X/2 as the desired return period.

climakitae/explore/threshold_tools.py
acordonez: Correct, that appears to be the way this currently works.

acordonez: Correct, that appears to be how the return variables have been calculated

climakitae/explore/threshold_tools.py
acordonez: @vicford Lost my original reply, sorry about that. Yes, if the user doesn't divide by block size they'll get an answer that is not incorrect per se but it is not an annual value unless the block size was 1 when generating the BMS. There was code to check the block size attribute of the BMS when getting the return value but it did not function correctly. We could fix that bug and return annual results for all block sizes.

acordonez: In this PR I was just trying to clear up how the code appears to actually work, but am also open to returning annual return variable results using the BMS attributes or including the block size as a variable in get_return_value, get_return_period, etc.

acordonez:
Screenshot 2026-04-21 at 1 38 09 PM
Added a print statement to verify what block size is being used.

acordonez:
@vicford Ready for another look, made changes based on our discussion earlier!

vicford (approved):
Sensational, great work @acordonez

acordonez:
@neilSchroeder Would appreciate a technical code review on this.

neilSchroeder (approved):

PR #811, opened 2026-09-02
fixing warming level file for non-standard warming levels

neilSchroeder:

- `ClimateData` warming level processor uses the wrong file
- `get_data()` does not use the wrong file
- In March of 2026, the processor was briefly updated to use the correct file, but in May of 2026 that change was reverted probably due to a malformed merge/rebase

### What does this affect:
ONLY affects non-standard gwl values
NON-AFFECTED warming levels include:
0.8, 1.0, 1.2, 1.5, 2.0, 2.5, 3.0, 4.0 °C

### GWL Status:
TMY, shock_XMY, and persistence_XMY all use the "buggy" warming level code 
HOWEVER, climate profiles were only produced for 1.2, 1.5, 2.0, and 2.5 so the hosted profiles ARE NOT affected by this bug. 
That said if a user entered a non-standard profile they would have problems. 
Standard Year profiles still use get_data as the data fetch tool so it's not affected at all. 

## Link to corresponding Jira ticket(s)
https://cal-adaptdp.slack.com/archives/C021EAYF0MC/p1788355626531949

## Testing
- [x] Unit tests written for new/modified code (goal: 80% coverage)
  - All public functions have unit tests
  - Functions that must produce specific values have unit tests
- [x] Verified that notebooks utilizing affected functions still work
- [x] Appropriate manual testing completed
- [x] Advanced Testing label added to PR if this PR makes major changes to the codebase 

## How to Test
Confirm that 2.5 and 2.52 produce almost identical results. Confirm 2.5 and 2.6 produce very similar results.

## Documentation
**Which of the following need updating? Check all that apply and complete them:**
- [x] **API reference** (`docs-mkdocs/api/`) — add or update the `.md` file for the affected module (e.g. `processors.md`, `tools.md`, `util.md`)

## Review Process
- [x] PR review instructions provided:
  - [x] Type of review requested (scientific/technical/debugging)
  - [x] Level of review detail needed

### DIFF
--- a/climakitae/new_core/processors/warming_level.py
+++ b/climakitae/new_core/processors/warming_level.py
@@ -14,7 +14,7 @@ import pandas as pd
 import xarray as xr
 
 from climakitae.core.constants import _NEW_ATTRS_KEY, UNSET
-from climakitae.core.paths import GWL_1850_1900_FILE, GWL_1981_2010_TIMEIDX_FILE
+from climakitae.core.paths import GWL_1850_1900_FILE, GWL_1850_1900_TIMEIDX_FILE
 from climakitae.new_core.data_access.data_access import DataCatalog
 from climakitae.new_core.processors.abc_data_processor import (
     DataProcessor,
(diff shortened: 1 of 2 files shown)


acordonez (approved):
Compared time mean WRF tmax values for Alpine County. Warming levels 2.6-2.5 had a max difference over the county of 0.44805908°F. Warming levels 2.53-2.5 had a max difference of 0.16905975°F.

claalmve:
Compared this with original WL code, uptick in warming from 2.5 to 2.53 and 2.6 is much lower now than in the original code.

claalmve (approved):

